PR #58, opened 2023-05-22
Simplify public api

mtmk:
Simplified `INatsCommand` interface to support NATS core protocol (except headers). Code clean-up left for later to keep this change as small as possible.


caleblloyd (approved):
LGTM

PR #61, opened 2023-06-09
Remove internal subscription to same subject

mtmk:
* Removed the reference counting internal subscription feature when subscribing to same subject.
* Also removed now redundant message serialisation mechanisms under `MessagePublisher`.
* Removed `NatsKey` struct since its performance benefit is negligible and it's confusing.
* Removed dead code
* Renamed `subscriptionId` to more concise name `sid` since it's a well known NATS concept


src/NATS.Client.Core/NatsSub.cs
```diff
@@ -45,24 +42,25 @@ public sealed class NatsSub : NatsSubBase
...
 
+    internal NatsSub(NatsConnection connection, SubscriptionManager manager, string subject, string? queueGroup, int sid)
+        : base(connection, manager, subject, queueGroup, sid)
+    {
+    }
+
     public ChannelReader<NatsMsg> Msgs => _msgs.Reader;
 
-    public override async ValueTask DisposeAsync()
+    public override void Dispose()
```
caleblloyd, line 52: Should this remain `DisposeAsync` and wait for the UNSUB to complete?

Even if we don't await the UNSUB, I am a bit wary of using `Dispose` as opposed to `DisposeAsync` here should we ever want to await a task in the future.

caleblloyd (approved):
LGTM

PR #63, opened 2023-06-19
Headers parsing

mtmk:
Implementations of `HPUB` and `HMSG` messages.

Also includes minor code clean-up and some test improvements to reduce false positives.


src/NATS.Client.Core/Commands/ProtocolWriter.cs
```diff
@@ -46,159 +50,63 @@ public void WritePong()
     }
 
     // https://docs.nats.io/reference/reference-protocols/nats-protocol#pub
-    // PUB <subject> [reply-to] <#bytes>\r\n[payload]
-    // To omit the payload, set the payload size to 0, but the second CRLF is still required.
-    public void WritePublish(string subject, string? replyTo, ReadOnlySequence<byte> payload)
+    // PUB <subject> [reply-to] <#bytes>\r\n[payload]\r\n
+    public void WritePublish(string subject, string? replyTo, NatsHeaders? headers, ReadOnlySequence<byte> payload)
     {
-        var offset = 0;
```
mtmk, line 53: Removed precalculated `offset +=...` code since *BufferWriter.Advance()* method (which also marked with *AggressiveInlining*) does just that anyway and makes it much easier comprehend.

tests/NATS.Client.Core.Tests/SubscriptionTest.cs
```diff
@@ -8,94 +6,6 @@ public class SubscriptionTest
 
     public SubscriptionTest(ITestOutputHelper output) => _output = output;
 
-    [Fact]
-    public async Task Subscription_with_same_subject()
-    {
```
mtmk, line 13: This is moved to the ProtocolTest

src/NATS.Client.Core/Internal/BufferExtensions.cs
```diff
@@ -0,0 +1,69 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
caleblloyd, line 2:
```suggestion
// Adapted from https://github.com/dotnet/aspnetcore/blob/v6.0.18/src/Shared/ServerInfrastructure/BufferExtensions.cs
```

src/NATS.Client.Core/Internal/HeaderParser.cs
```diff
@@ -0,0 +1,333 @@
+using System.Buffers;
```
caleblloyd, line 1:
```suggestion
// Adapted from https://github.com/dotnet/aspnetcore/blob/v6.0.18/src/Servers/Kestrel/Core/src/Internal/Http/HttpParser.cs

using System.Buffers;
```

src/NATS.Client.Core/NatsSub.cs
```diff
@@ -9,6 +11,7 @@ internal NatsSubBase(NatsConnection connection, SubscriptionManager manager, str
     {
         Connection = connection;
         Manager = manager;
+        HeaderParser = new HeaderParser(Encoding.UTF8);
```
caleblloyd, line 14: Header Parser is thread safe, right?  Should we can allocate 1 per connection as opposed to 1 per subscription.

Also, the default encoding is technically ASCII since this follows the HTTP RFC

Certain clients enforce ASCOO while others do not though.  For example the Go implementation defaults to UTF-8.

So I think that we should add `HeaderEncoding` to `NatsOptions`, default it to `Encoding.ASCII` there but let someone change it if they want.

mtmk: All makes sense.

caleblloyd (approved):
Implementation LGTM - looks like a test might be flapping though because the CI failed

mtmk:
> Implementation LGTM - looks like a test might be flapping though because the CI failed

CI was failing because of a [change on the server](https://github.com/nats-io/nats-server/pull/4255) reporting WS port instead of standard port which broke cluster reconnect test. There was also an issue reading the server version in tests. All issues in test code and they're fixed now.

PR #65, opened 2023-06-27
Low-level APIs and last clean-up before first release

mtmk:
Implementation of a lower level subscription API to cater for message patterns that might require different approach than current channel based subscription method. Also removed some of the unused / unnecessary code.


caleblloyd (approved):
LGTM

PR #70, opened 2023-06-30
Auto-unsubscribe implementations and fixes

mtmk:
* Auto-unsubscribe on max-msgs, timeout and idle timeout
* Pass NatsMsg by ref to reduce copy overhead
* Hidden internal INatsSub API from public API using explicit interface implementations


caleblloyd (approved):
LGTM

PR #72, opened 2023-07-03
Have DLLs stamped with version too

mtmk:
Assigning `Version` property has the following effect on the DLL files, if the `version.txt` (hence NuGet version) was `2.0.0-alpha.0` for example:
* *File version* and *Assembly version*: `2.0.0.0`
* *Product version*: `2.0.0-alpha.0`


caleblloyd (approved):
LGTM

PR #74, opened 2023-07-03
Initial docfx setup

mtmk:
Initial [DocFX](https://dotnet.github.io/docfx/) setup for API documentation to be published as GitHub Pages. GitHub pages might need to be activated in repo settings to point to the root of `gh-pages` branch. Manually created documentation at the moment is thin and written just to demonstrate docfx features.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,32 @@
+name: Docs
+
+on:
+  push:
+    branches:
+    - main
+
```
caleblloyd, line 7:
```suggestion
# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
permissions:
  contents: read
  pages: write
  id-token: write

# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
concurrency:
  group: "pages"
  cancel-in-progress: false

```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,32 @@
...
+          dotnet-version: '7.x'
+
+      - run: dotnet tool update -g docfx
+      - run: docfx docs/docfx.json
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: docs/_site
```
caleblloyd, line 32:
```suggestion
      - name: Upload artifact
        uses: actions/upload-pages-artifact@v1
        with:
          path: docs/_site
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v2
```

caleblloyd (commented):
Recommended change to use the standardized GH Pages actions

caleblloyd (approved):
LGTM

Issue #75, opened 2023-07-04
API Review - Expand Request Extension Methods


caleblloyd:
After reviewing the Request extension method, I think it might make sense to expand it to more closely match the core API

The idea here is to re-use PubOpts as `requestOpts` and SubOpts as `replyOpts`, which means that any future additions to those options will flow through and be fully supported in Request/Reply.

We are actually very close to being able to implement most of the `RequestMany` options discussed in [ADR-40](https://github.com/nats-io/nats-architecture-and-design/pull/228) as well - so I've included an example of how the `RequestMany` methods could look too

```csharp
    // RequestAsync methods
    // Same as PublishAsync with the following changes
    // - Response is 0 (null) or 1 NatsMsg
    // - PubOpts is called requestOpts
    // - add SubOpts replyOpts
    //   - default replyOpts.MaxMsgs to 1
    //   - if replyOpts.Timeout == null then set to NatsOptions.RequestTimeout
    
    ValueTask<NatsMsg?> RequestAsync(
        string subject,
        ReadOnlySequence<byte> payload = default,
        in NatsPubOpts? requestOpts = default,
        in NatsSubOpts? replyOpts = default,
        CancellationToken cancellationToken = default);

    ValueTask<NatsMsg?> RequestAsync(
        in NatsMsg msg,
        in NatsSubOpts? replyOpts = default,
        CancellationToken cancellationToken = default);
    
    ValueTask<NatsMsg<TReply>?> RequestAsync<TRequest, TReply>(
        string subject,
        TRequest data,
        in NatsPubOpts? requestOpts = default,
        in NatsSubOpts? replyOpts = default,
        CancellationToken cancellationToken = default);
    
    ValueTask<NatsMsg<TReply>?> RequestAsync<TRequest, TReply>(
        in NatsMsg<TRequest> msg,
        in NatsSubOpts? replyOpts = default,
        CancellationToken cancellationToken = default);

    
    // RequestManyAsync methods
    // Same as PublishAsync with the following changes
    // - Response is IAsyncEnumerable<NatsMsg>
    // - PubOpts is called requestOpts
    // - add SubOpts replyOpts
    //   - if replyOpts.Timeout == null then set to NatsOptions.RequestTimeout
    
    ValueTask<IAsyncEnumerable<NatsMsg>> RequestManyAsync(
        string subject,
        ReadOnlySequence<byte> payload = default,
        in NatsPubOpts? requestOpts = default,
[...]

PR #78, opened 2023-07-05
Revised request API

mtmk:
Revised request(many) API based on #75 
Also added start-up timer as per ADR-40. It's the max time to wait for the first message.


mtmk:
Are we happy with the public interface i.e. `NatsSub` and `Request*Async`?

* I had doubts about the options being passed to `Request` but I can see the value of not littering the API with too many types.

* `RequestMany` returning async enumerator forces the developer to use `await foreach`. Returning a type with a channel e.g. `NatsSub` may be a better option.

caleblloyd:
For the `Request` options - we need a way to pass Queue Group, Headers, and a Serializer when using Model types.  So the only thing missing from `PubOpts` is the ReplySubject which is defaulted.

For the `Reply` options of the `RequestAsync` methods we could cut those down, as many of the timeouts do not apply when waiting for a single reply.  Or we could document that they have no effect when waiting for a single reply.

For the `Reply` options of `RequestManyAsync` - this pretty much does need to be the full set of `NatsSubOpts`

I think that returning `NatsSub` is fine for the `RequestManyAsync` methods

src/NATS.Client.Core/NatsRequestManyAsyncExtensions.cs
```diff
@@ -0,0 +1,143 @@
...
+    // Same as PublishAsync with the following changes
+    // - Response is IAsyncEnumerable<NatsMsg>
+    // - PubOpts is called requestOpts
+    // - add SubOpts replyOpts
+    //   - if replyOpts.Timeout == null then set to NatsOptions.RequestTimeout
+    public static async IAsyncEnumerable<NatsMsg> RequestManyAsync(
+        this NatsConnection nats,
+        string subject,
+        ReadOnlySequence<byte> payload = default,
+        NatsPubOpts? requestOpts = default, // Not used
```
caleblloyd, line 19: I think this could be used - Headers and Queue Group could be set here right?

mtmk: Yes, spot on! Thank you. I forgot about headers when publishing. Queue group doesn't apply to pub-opts. But a good question. Does it make sense to have a queue group on an inbox subscription? If it does, how should we go about it when our inbox is mux-ed and shared, as in this case?

mtmk:
> For the `Request` options - we need a way to pass Queue Group and Headers when not using Model types. So the only thing missing from `NatsPubOpts` is the ReplySubject which is defaulted.

see comment above

> For the `Reply` options of the `RequestAsync` methods we could cut those down, as many of the timeouts do not apply when waiting for a single reply. Or we could document that they have no effect when waiting for a single reply.

I'm happy with documenting it. Changed my mind a little. Maybe it's better to keep Type footprint to a minimum. Also, in my opinion at this level of the API it's better to give clues that under the bonnet it is still a pub/sub powering the request reply pattern. We can certainly wrap this later with a much more opinionated/easy/lite API for less demanding applications.

> For the `Reply` options of `RequestManyAsync` - this pretty much does need to be the full set of `NatsSubOpts`

Agreed.

> I think that returning `NatsSub` is fine for the `RequestManyAsync` methods

I was thinking to wrap this PR up as a smaller step and discuss this on a some kind of 'internal redesign' PR. Happy to carry on here though if you like.

src/NATS.Client.Core/NatsRequestManyAsyncExtensions.cs
caleblloyd: Oops - yes - queue group is on subscribe.  Good catch.  No, I don't think an Inbox Subscription really needs to support a Queue Group, but it's also not a big deal if it does

caleblloyd:
> I was thinking to wrap this PR up as a smaller step and discuss this on a some kind of 'internal redesign' PR

That sounds good!

caleblloyd (approved):
LGTM

PR #79, opened 2023-07-05
remove Sid from INatsSub

caleblloyd:
Rough draft - `dotnet format` touched a bunch of files so sorry for the confusing diff

Removes `Sid` from `INatsSub`.  Also introduces `ISubscriptionManager`

`Sid` is an implementation detail of the default subscription manager.  For muxed subscriptions, such as `InboxSubscription`, an `InboxSubscriptionManager` could be used instead.


mtmk (approved):
LGTM

PR #81, opened 2023-07-07
Inbox req-reply redesign

mtmk:
Removed all the timer duplication by removing `MsgWrapper` `IValueTaskSource` implementation. Which means we will incur 3+ additional object allocations (`NatsSub` related) per `RequestAsync()` call, but they should not escape _Gen0_ most of the time, so may be acceptable.

Subject manager is now responsible for inbox creation. Which means any subscription with inbox prefixed subject will be passed on to the inbox subscription, which is created now safely and efficiently.

API changes:
Main API call now is `NatsSub RequestAsync()`. This meant we also had to rename the call to `NatsMsg RequestSingleAsync()`.

- [x] More tests
- [x] Review request / inbox sub cancellation


src/NATS.Client.Core/NatsRequestSingleExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+public static class NatsRequestSingleExtensions
+{
+    // RequestAsync methods
+    // Same as PublishAsync with the following changes
+    // - Response is 0 (null) or 1 NatsMsg
+    // - PubOpts is called requestOpts
+    // - add SubOpts replyOpts
+    //   - default replyOpts.MaxMsgs to 1
+    //   - if replyOpts.Timeout == null then set to NatsOptions.RequestTimeout
+    public static async ValueTask<NatsMsg<TReply?>?> RequestSingleAsync<TRequest, TReply>(
```
caleblloyd, line 15: I think most of the other clients simply call it `Request` - not `RequestSingle` - [Reference](https://natsbyexample.com/examples/messaging/request-reply/go)

caleblloyd (approved):
LGTM

PR #84, opened 2023-07-11
Test matrix for NATS Server versions

mtmk:
Changed GitHub workflow to test against both dev and main versions of `nats-server`.

Sorry for the noise. Main change is in `.github/workflows/test.yml`.


caleblloyd (approved):
LGTM

PR #86, opened 2023-07-11
Initial JetStream support

mtmk:
Started the new JetStream project:

* Hand maintained JSON models
* Using .NET validation
* Naming convention: prefix JetStream classes with `Js` to avoid generic names like `Context` or `Stream` to be distinguished easily and to not be confusing; avoid confusing names like `JetStreamStream` while having consistent naming pattern.
* Currently JSON models exposed and passed through into and out of API. (Should we have an intermediate e.g. `JetStreamConfig` class?)


src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <!-- NuGet Packaging -->
+    <PackageTags>pubsub;messaging</PackageTags>
+    <Description>JetStream support for NATS.Client.</Description>
+    <IsPackable>true</IsPackable>
```
caleblloyd, line 11: May want to flip to `false` with a comment of "change to true once ready to publish"

Otherwise if this gets into `main` before we are ready to release, and a version is incremented, then it will end up going out

tests/NATS.Client.Core.Tests/NATS.Client.Core.Tests.csproj
```diff
@@ -22,6 +22,7 @@
     </ItemGroup>
 
     <ItemGroup>
+        <ProjectReference Include="..\..\src\NATS.Client.JetStream\NATS.Client.JetStream.csproj" />
```
caleblloyd, line 25: Do you think it would be better to have a separate test assembly for JetStream?  That would probably also require adding a 3rd test assembly that has all of the base setup functionality.

src/NATS.Client.JetStream/Models/Config.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System.Text.Json.Serialization;
+
+namespace NATS.Client.JetStream.Models;
+
+public record Config
```
scottf, line 5: Is this StreamConfig? There are multiple configs.

src/NATS.Client.JetStream/Models/State.cs
```diff
@@ -0,0 +1,30 @@
+using System.Text.Json.Serialization;
+
+namespace NATS.Client.JetStream.Models;
+
+public record State
```
scottf, line 5: StreamState?

src/NATS.Client.JetStream/Models/State.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    [JsonPropertyName("messages")]
+    public long Messages { get; set; }
```
scottf, line 8: ulong in this structure
Messages, Bytes, FirstSeq, LastSeq

src/NATS.Client.JetStream/Models/StreamCreateRequest.cs
```diff
@@ -0,0 +1,66 @@
...
+    public string Storage { get; set; } = "file";
+
+    [JsonPropertyName("discard")]
+    public string Discard { get; set; } = "old";
+
+    [JsonPropertyName("num_replicas")]
+    public long NumReplicas { get; set; } = 1;
+
+    [JsonPropertyName("duplicate_window")]
+    public long DuplicateWindow { get; set; } = 120000000000;
```
scottf, line 47: When sending to server, do not set defaults except for very few things. This allows the server to control what defaults are. This includes booleans. Values should not be sent to the server if they are not explicitly set by the user or are specific defaults.

scottf: defaults:
RetentionPolicy = limits
StorageType = file
DiscardPolicy = old

oising:
> * Naming convention: prefix JetStream classes with `Js` to avoid generic names like `Context` or `Stream` to be distinguished easily and to not be confusing; avoid confusing names like `JetStreamStream` while having consistent naming pattern.

According to the official framework design guidelines -- https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions -- two letter prefixes should be capitalized, so consider using JS, not Js. 

> A special case is made for two-letter acronyms in which both letters are capitalized, as shown in the following identifier:
> IOStream

turowicz:
`async Fetch` for the `jetStream.PullSubscribe` please!

src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
mtmk: Very good point!

tests/NATS.Client.Core.Tests/NATS.Client.Core.Tests.csproj
mtmk: Yes. Good idea. Created a `TestUtilities` project for shared functions.

src/NATS.Client.JetStream/Models/Config.cs
mtmk: Yes, I think so, It was part of the stream creation response. Fixed now.

src/NATS.Client.JetStream/Models/State.cs
mtmk: Fixed.

src/NATS.Client.JetStream/Models/State.cs
mtmk: It was string in the generated classes, so I kept that.

mtmk:
> > * Naming convention: prefix JetStream classes with `Js` to avoid generic names like `Context` or `Stream` to be distinguished easily and to not be confusing; avoid confusing names like `JetStreamStream` while having consistent naming pattern.
> 
> According to the official framework design guidelines -- https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions -- two letter prefixes should be capitalized, so consider using JS, not Js.
> 
> > A special case is made for two-letter acronyms in which both letters are capitalized, as shown in the following identifier:
> > IOStream

Good catch, thanks. Fixed.

mtmk:
> @mtmk How do these models support server responses that could either be an error or the actual object. For example if you make a request for consumer info, you could get error json back or consumer info json back. Can models be subclassed?

@scottf good point. I'm not sure, I can give that a try. If there isn't an easy way already, we might have to implement something using `Utf8JsonReader`

mtmk:
> > * Currently JSON models exposed and passed through into and out of API. (Should we have an intermediate e.g. `JetStreamConfig` class?)

>   What would `JetStreamConfig` be composed of / what is it for?

@scottf it's supposed to be a level of abstraction i.e. something other that the JSON representation. I'd be better if we don't need it of course.

src/NATS.Client.JetStream/Models/State.cs
mtmk: I take it back. was getting deserialization errors. Changed them to ulong.

turowicz:
> @turowicz this library is Async I/O only, so it will be Async once implemented!

sorry I don't fully understand

turowicz:
> > @turowicz this library is Async I/O only, so it will be Async once implemented!
> 
> sorry I don't fully understand, other kinds of Subscription support async fetch/pull

@caleblloyd

mtmk:
> > @turowicz this library is Async I/O only, so it will be Async once implemented!
> 
> sorry I don't fully understand, other kinds of Subscription support async fetch/pull

@turowicz, @caleblloyd meant purely C# async/await by Async IO.

caleblloyd (approved):
LGTM

PR #88, opened 2023-07-12
Adding generated records

scottf:
These are probably 90% good, need manual review. Arrays for instance are not addressed, and are written as strings.


mtmk (approved):
LGTM

PR #90, opened 2023-07-13
generate JetStream models with NJsonSchema

caleblloyd:
Uses [NJsonSchema](https://github.com/RicoSuter/NJsonSchema) and [schemas from jsm.go](https://github.com/nats-io/jsm.go/tree/main/schema_source/jetstream/api/v1) to generate JetStream models.

I copied the schema files to this project, as I did end up editing them slightly.  We may just want to treat this as a 1-time import though, where the models can be hand edited afterwards and don't need to stay in sync.  f that is the case, I can remove the schema files and import script from the PR.  Or we can leave them for now, and remove them in the future once they have diverged.


mtmk (approved):
LGTM

PR #93, opened 2023-07-18
Allow Configuration of Subscription Channels

to11mtm:
There is value in being able to configure the `BoundedChannelOptions` for a given subscription (or series of subscriptions.)

For example, in cases where it is preferred to discard older data on a subscription, or cases where _only_ the most recent value is the one that is desired.

This PR also has the benefit, in the default case, we can just re-use our internal `BoundedChannelOptions` rather than passing a new one every time a new subscription is created (i.e. less memory allocation.)

Note that for the sake of reuse, this marked `internal static` on `NatsSub` rather than `private static`, since to have a separate static readonly on `NatsSub<T>` would both be unnecessary and would cause a new static to be allocated for every new `T` on `NatsSub<T>`.

### DIFF
--- a/src/NATS.Client.Core/NatsSub.cs
+++ b/src/NATS.Client.Core/NatsSub.cs
@@ -104,6 +104,8 @@ public abstract class NatsSubBase : INatsSub
     // since INatsSub is marked as internal.
     int? INatsSub.PendingMsgs => _pendingMsgs == -1 ? null : Volatile.Read(ref _pendingMsgs);
 
+    public Exception? Exception => Volatile.Read(ref _exception);
+
     internal NatsSubEndReason EndReason => (NatsSubEndReason)Volatile.Read(ref _endReasonRaw);
 
     protected NatsConnection Connection { get; }
@@ -172,8 +174,6 @@ public abstract class NatsSubBase : INatsSub
 
     protected abstract ValueTask ReceiveInternalAsync(string subject, string? replyTo, ReadOnlySequence<byte>? headersBuffer, ReadOnlySequence<byte> payloadBuffer);
 
-    public Exception? Exception => Volatile.Read(ref _exception);
-
     protected void SetException(Exception exception)
     {
         Interlocked.Exchange(ref _exception, exception);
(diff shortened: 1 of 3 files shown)


src/NATS.Client.Core/NatsSub.cs
```diff
@@ -212,21 +212,47 @@ private void EndSubscription(NatsSubEndReason reason)
...
     public ChannelReader<NatsMsg> Msgs => _msgs.Reader;
 
+    internal static BoundedChannelOptions GetChannelOptions(
+        NatsSubChannelOpts? subChannelOpts)
+    {
+        if (subChannelOpts != null)
+        {
+            var overrideOpts = subChannelOpts.Value;
+            return new BoundedChannelOptions(overrideOpts.Capacity ??
+                                             DefaultChannelOptions.Capacity)
```
to11mtm, line 240: It seemed better from a maintenance perspective to fallback to `DefaultChannelOptions` rather than re-specify defaults for these. If preferred I can just re-specify defaults however.

src/NATS.Client.Core/NatsSub.cs
```diff
@@ -55,7 +55,7 @@ public abstract class NatsSubBase : INatsSub
             _cts.Token.UnsafeRegister(
                 static (self, _) =>
                 {
-                    ((NatsSubBase) self!).EndSubscription(NatsSubEndReason.Cancelled);
+                    ((NatsSubBase)self!).EndSubscription(NatsSubEndReason.Cancelled);
```
to11mtm, line 58: Linter complained about this so I went ahead and did it. If revert is preferred LMK.

src/NATS.Client.Core/NatsSubOpts.cs
```diff
@@ -1,3 +1,5 @@
+using System.Threading.Channels;
```
to11mtm, line 1: I think this `using` came in because of the XmlDoc, if preferred I can just use full type name in xmldoc?

src/NATS.Client.Core/NatsSub.cs
mtmk: Make sure to run `dotnet format`. That should fix these issues as well.

mtmk:
@to11mtm couls you rebase please? There are no conflicts but it'd be easier to see without warnings etc. which was fixed afterwards.

mtmk (approved):
LGTM

PR #96, opened 2023-07-19
JetStream basic usage

mtmk:
Basic use case APIs, publish and consume methods implemented. Consider this as the 'low-level' JetStream API where we can build our more familiar NATS Client Simplified JetStream API.

### DIFF
--- a/src/NATS.Client.JetStream/JSContext.cs
+++ b/src/NATS.Client.JetStream/JSContext.cs
@@ -1,5 +1,11 @@
 ﻿using System.ComponentModel.DataAnnotations;
+using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using NATS.Client.Core;
+using NATS.Client.Core.Internal;
+using NATS.Client.JetStream.Internal;
 using NATS.Client.JetStream.Models;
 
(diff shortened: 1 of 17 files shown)


src/NATS.Client.JetStream/JSContext.cs
```diff
@@ -15,23 +21,158 @@ public JSContext(NatsConnection nats, JSOptions options)
...
+    {
+        await using var sub = await _nats.RequestSubAsync<T, PubAckResponse>(
+                subject: subject,
+                data: data,
+                requestOpts: opts,
+                replyOpts: default,
+                cancellationToken)
+            .ConfigureAwait(false);
+
+        if (await sub.Msgs.WaitToReadAsync(CancellationToken.None).ConfigureAwait(false))
```
caleblloyd, line 62: Saw `CancellationToken.None` in 3 different places - should it be passing `cancellationToken` or is it intentional to use a different cancellation token for these particular calls?

mtmk: It is intentional because the token passed to the subscription should complete the channel to allow graceful exit from the function. That's the idea anyway.

caleblloyd (approved):
The Cancellation changes discussed could be done in a follow-up, this looks good

PR #97, opened 2023-07-21
NatsSub cancellation fix

mtmk:


caleblloyd (approved):
LGTM.  Although looks like there may be a couple flapping tests

mtmk:
> LGTM. Although looks like there may be a couple flapping tests, didn't look related to the change at first glance

Yes, unfortunately we have a few. Reran the failed jobs. Looks like we have a few failures related to server startup in general and there maybe other specific ones failing regularly. Need to investigate a little.

PR #107, opened 2023-08-02
JetStream consumer options

mtmk:
Implement JetStream Consumer API `consume()` options based on [ADR-37](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-37.md).

The design idea is to implement a `NatsJSSubBase` class to handle control messages and timeouts. Then we can implement specific classes for all three calls, `next()`, `fetch()` and `consume()`. In this PR only `consume()` is implemented.


mtmk:
We seem to have an issue with `NATS.Client.Core.Tests.NatsConnectionTestTls.UserCredentialAuthTest` against nats-server dev version. Can't repro on my machine.

src/NATS.Client.JetStream/Internal/NatsNatsJSSubBaseConsume.cs
```diff
@@ -0,0 +1,136 @@
...
+        // User messages are buffered here separately to allow smoother flow while control loop
+        // pulls more data in the background. This also allows control messages to be dealt with
+        // in the same loop as the control messages to keep state updates consistent. This is as
+        // opposed to having a control and a message channel at the point of serializing the messages
+        // in NatsJSSub class.
+        _userMessageChannel = Channel.CreateBounded<NatsJSMsg<T?>>(new BoundedChannelOptions(1_000)
+        {
+            FullMode = BoundedChannelFullMode.Wait,
+            AllowSynchronousContinuations = false,
+        });
```
caleblloyd, line 82: should this take the options from `NatsSubOpts.ChannelOptions`?

src/NATS.Client.JetStream/Internal/NatsNatsJSSubBaseConsume.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        // We drop the old message if notification handler isn't able to keep up.
+        // This is to avoid blocking the control loop and making sure we deliver all the messages.
+        // Assuming newer messages would be more relevant and worth keeping than older ones.
+        _notificationChannel = Channel.CreateBounded<NatsJSNotification>(new BoundedChannelOptions(1_000)
+        {
+            FullMode = BoundedChannelFullMode.DropOldest,
+            AllowSynchronousContinuations = false,
+        });
```
caleblloyd, line 91: 1k seems excessive but I don't really know what a good number would be - 10?  100? :laughing: 

Also doesn't really seem worth making configurable.  I guess they are small so 1k might be fine.

src/NATS.Client.JetStream/Internal/NatsNatsJSSubBaseConsume.cs
mtmk: True 💯  I didn't think of that.

caleblloyd (approved):
LGTM

PR #108, opened 2023-08-07
NatsSub simplification and JetStream support

mtmk:
* Remove INatsSubBuilder interfaces: This is to reduce unnecessary complexity since NatsSubBase can be used as the contract in SubscriptionManager

* Create a separation of public interfaces and internal handling for subscriptions by using INatsSub* interfaces for public methods and using NatsSubBase as the internal handling interface with internals exposed such as Pending messages.

* Push responsibility of reconnections to SubscriptionManager so that not only we can renew subscriptions but in the case of JetStream consumers recovery, we can issue pull requests for example.


caleblloyd:
Looks good, I like this approach!  Want to clean up the commented sections so we can merge it?

mtmk:
> Looks good, I like this approach! Want to clean up the commented sections so we can merge it?

Brilliant! I shall clean it up.

caleblloyd (approved):
LGTM

PR #109, opened 2023-08-09
Add INatsConnectionPool interface 

jasper-d:
https://github.com/nats-io/nats.net.v2/issues/105

### DIFF
--- a/src/NATS.Client.Core/INatsConnection.cs
+++ b/src/NATS.Client.Core/INatsConnection.cs
@@ -1,4 +1,5 @@
 using System.Buffers;
+using System.Runtime.CompilerServices;
 
 namespace NATS.Client.Core;
 
@@ -69,4 +70,92 @@ public interface INatsConnection
     /// <typeparam name="T">Specifies the type of data that may be received from the NATS Server.</typeparam>
     /// <returns>A <see cref="ValueTask{TResult}"/> that represents the asynchronous send operation.</returns>
     ValueTask<INatsSub<T>> SubscribeAsync<T>(string subject, NatsSubOpts? opts = default, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Request and receive a single reply from a responder.
+    /// </summary>
+    /// <param name="subject">Subject of the responder</param>
+    /// <param name="data">Data to send to responder</param>
+    /// <param name="requestOpts">Request publish options</param>
+    /// <param name="replyOpts">Reply handler subscription options</param>
+    /// <param name="cancellationToken">Cancel this request</param>
+    /// <typeparam name="TRequest">Request type</typeparam>
+    /// <typeparam name="TReply">Reply type</typeparam>
+    /// <returns>Returns the <see cref="NatsMsg{T}"/> received from the responder as reply.</returns>
+    /// <exception cref="OperationCanceledException">Raised when cancellation token is used</exception>
+    /// <remarks>
+    /// Response can be (null) or one <see cref="NatsMsg"/>.
+    /// Reply option's max messages will be set to 1.
+    /// if reply option's timeout is not defined then it will be set to NatsOptions.RequestTimeout.
+    /// </remarks>
+    ValueTask<NatsMsg<TReply?>?> RequestAsync<TRequest, TReply>(
+        string subject,
+        TRequest? data,
+        NatsPubOpts? requestOpts = default,
+        NatsSubOpts? replyOpts = default,
+        CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Request and receive a single reply from a responder.
+    /// </summary>
+    /// <param name="subject">Subject of the responder</param>
+    /// <param name="payload">Payload to send to responder</param>
+    /// <param name="requestOpts">Request publish options</param>
+    /// <param name="replyOpts">Reply handler subscription options</param>
+    /// <param name="cancellationToken">Cancel this request</param>
+    /// <returns>Returns the <see cref="NatsMsg"/> received from the responder as reply.</returns>
+    /// <exception cref="OperationCanceledException">Raised when cancellation token is used</exception>
+    /// <remarks>
+    /// Response can be (null) or one <see cref="NatsMsg"/>.
+    /// Reply option's max messages will be set to 1 (one).
+    /// if reply option's timeout is not defined then it will be set to NatsOptions.RequestTimeout.
+    /// </remarks>
+    ValueTask<NatsMsg?> RequestAsync(
+        string subject,
+        ReadOnlySequence<byte> payload = default,
+        NatsPubOpts? requestOpts = default,
+        NatsSubOpts? replyOpts = default,
+        CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Request and receive zero or more replies from a responder.
+    /// </summary>
+    /// <param name="subject">Subject of the responder</param>
+    /// <param name="data">Data to send to responder</param>
+    /// <param name="requestOpts">Request publish options</param>
+    /// <param name="replyOpts">Reply handler subscription options</param>
+    /// <param name="cancellationToken">Cancel this request</param>
+    /// <typeparam name="TRequest">Request type</typeparam>
+    /// <typeparam name="TReply">Reply type</typeparam>
+    /// <returns>An asynchronous enumerable of <see cref="NatsMsg"/> objects</returns>
+    /// <exception cref="OperationCanceledException">Raised when cancellation token is used</exception>
+    /// <remarks>
+    /// if reply option's timeout is not defined then it will be set to NatsOptions.RequestTimeout.
+    /// </remarks>
+    IAsyncEnumerable<NatsMsg<TReply?>> RequestManyAsync<TRequest, TReply>(
+        string subject,
+        TRequest? data,
+        NatsPubOpts? requestOpts = default,
+        NatsSubOpts? replyOpts = default,
+        CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Request and receive zero or more replies from a responder.
+    /// </summary>
+    /// <param name="subject">Subject of the responder</param>
+    /// <param name="payload">Payload to send to responder</param>
+    /// <param name="requestOpts">Request publish options</param>
+    /// <param name="replyOpts">Reply handler subscription options</param>
+    /// <param name="cancellationToken">Cancel this request</param>
+    /// <returns>An asynchronous enumerable of <see cref="NatsMsg"/> objects</returns>
+    /// <exception cref="OperationCanceledException">Raised when cancellation token is used</exception>
+    /// <remarks>
+    /// if reply option's timeout is not defined then it will be set to NatsOptions.RequestTimeout.
+    /// </remarks>
+    IAsyncEnumerable<NatsMsg> RequestManyAsync(
+        string subject,
+        ReadOnlySequence<byte> payload = default,
+        NatsPubOpts? requestOpts = default,
+        NatsSubOpts? replyOpts = default,
+        CancellationToken cancellationToken = default);
 }
--- a/src/NATS.Client.Core/INatsConnection.cs
+++ b/src/NATS.Client.Core/INatsConnection.cs
@@ -72,6 +72,12 @@ public interface INatsConnection
     ValueTask<INatsSub<T>> SubscribeAsync<T>(string subject, NatsSubOpts? opts = default, CancellationToken cancellationToken = default);
 
     /// <summary>
+    /// Create a new inbox subject with the form {Inbox Prefix}.{Unique Connection ID}.{Unique Inbox ID}
+    /// </summary>
+    /// <returns>A <see cref="string"/> containing a unique inbox subject.</returns>
+    string NewInbox();
+
+    /// <summary>
     /// Request and receive a single reply from a responder.
     /// </summary>
     /// <param name="subject">Subject of the responder</param>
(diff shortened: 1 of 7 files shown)


src/NATS.Client.Core/NatsConnectionPool.cs
```diff
@@ -52,11 +56,6 @@ public IEnumerable<NatsConnection> GetConnections()
         }
     }
 
-    public INatsConnection GetCommand()
```
jasper-d, line 55: I removed this, because it's no identical to `INatsConnectionPool.GetConnection()`.

src/NATS.Client.Core/NatsConnectionPool.cs
```diff
@@ -38,12 +38,16 @@ public NatsConnectionPool(int poolSize, NatsOptions options, Action<NatsConnecti
         }
     }
 
+    INatsConnection INatsConnectionPool.GetConnection() => GetConnection();
```
jasper-d, line 41: Opted for an explicit interface implementation so clients could still call the non-virtual `GetConnection` if they are concerned about perf. That being said, I am almost certain that at least the virtual call to `GetConnection` could be inlined with GDV in .NET 8.

src/NATS.Client.Hosting/NatsHostingExtensions.cs
```diff
@@ -28,6 +28,11 @@ public static IServiceCollection AddNats(this IServiceCollection services, int p
                 return new NatsConnectionPool(poolSize, options, configureConnection ?? (_ => { }));
```
jasper-d, line 28: Somewhat unrelated: I was about to fix the XML docs for the method and while the comment got longer and longer, I realized that it would probably better to just have two methods.

Something like:

```diff 
-- public static IServiceCollection AddNats(this IServiceCollection services, int poolSize = 1, Func<NatsOptions, NatsOptions>? configureOptions = null, Action<NatsConnection>? configureConnection = null)
++ public static IServiceCollection AddNatsConnectionPool(this IServiceCollection services, int poolSize, Func<NatsOptions, NatsOptions>? configureOptions = null, Action<NatsConnection>? configureConnection = null)
++ public static IServiceCollection AddNatsConnection(this IServiceCollection services, Func<NatsOptions, NatsOptions>? configureOptions = null, Action<NatsConnection>? configureConnection = null)
```

src/NATS.Client.Core/NatsConnection.RequestSub.cs
```diff
@@ -1,31 +1,167 @@
...
+
+        if (sub is { EndReason: NatsSubEndReason.Exception, Exception: not null })
+        {
+            throw sub.Exception;
+        }
+
+        return null;
     }
 
+    public ValueTask<NatsMsg?> RequestAsync(
```
jasper-d, line 75: This and a few other request methods that only delegate and do not access non-public members could be kept as extension methods to keep interface size in check. Only though about this now :/

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -92,6 +92,8 @@ public NatsConnection(NatsOptions options)
 
     internal string InboxPrefix { get; }
 
+    public string NewInbox() => $"{InboxPrefix}{Guid.NewGuid():n}";
```
jasper-d, line 95: This used to be an extension method too. I'm not sure if it's necessary to keep this public. If so, it should be part of the interface.

src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -69,4 +69,165 @@ public interface INatsConnection
     /// <typeparam name="T">Specifies the type of data that may be received from the NATS Server.</typeparam>
     /// <returns>A <see cref="ValueTask{TResult}"/> that represents the asynchronous send operation.</returns>
     ValueTask<NatsSub<T>> SubscribeAsync<T>(string subject, in NatsSubOpts? opts = default, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
jasper-d: Just saw that's probably already solved with #108 👍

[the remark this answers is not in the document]
mtmk: Yes, sorry I should've ping the issue. We've started putting more interfaces including INatsSub. #108 was just merged today.

mtmk:
btw @jasper-d please sign your commits.

src/NATS.Client.Hosting/NatsHostingExtensions.cs
mtmk: I agree, it is getting long. But I kind of like a single entry point `AddNats(opts = default)` and an options lambda of sorts. That way it creates a natural progression for the API user. One can start with just `AddNats()` without being overwhelmed by overloads, then add options as and when needed. @caleblloyd might have a much better idea about this.

src/NATS.Client.Core/NatsConnection.cs
mtmk: You're right. It should be internal or not even be exposed here. Please feel free to leave this one out. I don't think It'd be necessary for unit testing.

src/NATS.Client.Hosting/NatsHostingExtensions.cs
caleblloyd: I like the idea of doing 2 extension methods - `AddNats` / `AddNatsPool` - or `AddNatsConnection` / `AddNatsConnectionPool` (the latter breaking BC but we are still in Alpha)

The `NatsPool` could have more settings than just a pool size in `NatsConnectionPoolOpts`.  It'd probably even be a good idea to default that to `Environment.ProcessorCount`

src/NATS.Client.Core/NatsConnection.RequestSub.cs
```diff
@@ -1,31 +1,167 @@
 using System.Buffers;
+using System.Runtime.CompilerServices;
 using NATS.Client.Core.Internal;
 
 namespace NATS.Client.Core;
 
 public partial class NatsConnection
 {
-    internal async ValueTask<NatsSub> RequestSubAsync(
+    public async ValueTask<NatsMsg<TReply?>?> RequestAsync<TRequest, TReply>(
```
caleblloyd, line 9: Are we definitely set on making moving these to `NatsConnection`?  Or should we try to convert them to extension methods that extend `INatsConnection` instead?

The extension methods could be nice for someone else who implements `INatsConnection` but I'm not sure if that will be common

src/NATS.Client.Core/NatsConnectionPool.cs
jasper-d: Not for the DI integration in NATS.Client.Hosting. It's always resolving `NatsConnectionPool` and than calling the non-virtual method.

src/NATS.Client.Hosting/NatsHostingExtensions.cs
jasper-d: I'll leave that for a follow up PR. :)

src/NATS.Client.Core/NatsConnection.cs
jasper-d: Done in 37a4b00fcf772e0666eb78d6e387b3c5160c3ffa

src/NATS.Client.Core/NatsConnection.RequestSub.cs
jasper-d: I've made changes to keep the four extension methods that don't need to be instance members as extension methods and only moved those four to the interface that access internal methods (i.e. `RequestSubAsync`).

I think outside of tests it's highly unlikely that anyone want's to implement INatsConnection.

src/NATS.Client.Core/NatsConnectionPool.cs
mtmk: Looks like still using callvirt:

https://sharplab.io/#v2:C4LglgNgPgAgzAAjAO2AUwE4DMCGBjNBASQQG8EYBGABgQFkAKASgG4EBfAWACh4KAmBAGEEIYmR4IpFGsQB0jJggC8APnrMWk6Xyq1FK9VTkBRZADcwGAPbIAtmlRyA4mmBnLN+4+AA1HBhgOABGEGgMAEQAGhFMAIRa3Fy8iDCCAAoS3NIUqQAs9AD6lAwieEpqCHgKmtpSugV0hfwMJGAV6mA1rDzJQA=

jasper-d: It's callvirt in MSIL, but since JIT sees the exact type in M_2, it uses a uses a non-virtual call. I find it easier to see in Compiler Explorer: https://godbolt.org/z/fYEnxEqo9

Here, `C.M()` got inlined which only happens because the call is non-virtual.

jasper-d: But for me it's fine either way, only having those methods in the interface or having the explicit interface implementation in addition to the methods on the class.

mtmk: > It's callvirt in MSIL, but since JIT sees the exact type in M_2, it uses a uses a non-virtual call. I find it easier to see in Compiler Explorer: https://godbolt.org/z/fYEnxEqo9

I see. However commenting out explicit implementation isn't making any difference https://godbolt.org/z/6Y8K9E3vf

mtmk: > But for me it's fine either way, only having those methods in the interface or having the explicit interface implementation in addition to the methods on the class.

We do care a lot about performance and this won't make any functional difference here, so if you think there may still be benefit, please add a few lines of comment explaining why we used an explicit interface here.

jasper-d: > However commenting out explicit implementation isn't making any difference

The "problem" is, that w/o the explicit implementations, clients can't obtain a `NatsConnection` instance but only an `INatsConnection` and thus all calls on that connection instance will be virtual.

I removed it anyway, because the longer I look at it, the more it feels like premature optimization. It can still be added anytime.

Some figures:

|      Method |           Job |       Runtime |      Mean |     Error |    StdDev |    Median | Ratio | RatioSD |
|------------ |-------------- |-------------- |----------:|----------:|----------:|----------:|------:|--------:|
|  CallDirect |      .NET 6.0 |      .NET 6.0 | 1.0188 ns | 0.0365 ns | 0.0342 ns | 1.0136 ns |  1.00 |    0.00 |
| CallVirtual |      .NET 6.0 |      .NET 6.0 | 2.6844 ns | 0.0399 ns | 0.0373 ns | 2.6744 ns |  2.64 |    0.10 |
|             |               |               |           |           |           |           |       |         |
|  CallDirect |      .NET 7.0 |      .NET 7.0 | 0.5650 ns | 0.0365 ns | 0.0324 ns | 0.5700 ns |  1.00 |    0.00 |
| CallVirtual |      .NET 7.0 |      .NET 7.0 | 2.1274 ns | 0.0560 ns | 0.0437 ns | 2.1451 ns |  3.75 |    0.23 |
|             |               |               |           |           |           |           |       |         |
|  CallDirect |      .NET 8.0 |      .NET 8.0 | 0.2981 ns | 0.0077 ns | 0.0068 ns | 0.2977 ns |  1.00 |    0.00 |
| CallVirtual |      .NET 8.0 |      .NET 8.0 | 0.3829 ns | 0.0284 ns | 0.0251 ns | 0.3783 ns |  1.28 |    0.08 |
|             |               |               |           |           |           |           |       |         |
|  CallDirect | NativeAOT 7.0 | NativeAOT 7.0 | 1.5757 ns | 0.0573 ns | 0.0926 ns | 1.6269 ns |  1.00 |    0.00 |
| CallVirtual | NativeAOT 7.0 | NativeAOT 7.0 | 1.1989 ns | 0.0529 ns | 0.0650 ns | 1.2314 ns |  0.78 |    0.06 |
|             |               |               |           |           |           |           |       |         |
|  CallDirect | NativeAOT 8.0 | NativeAOT 8.0 | 0.3246 ns | 0.0153 ns | 0.0143 ns | 0.3159 ns |  1.00 |    0.00 |
| CallVirtual | NativeAOT 8.0 | NativeAOT 8.0 | 0.5110 ns | 0.0382 ns | 0.0886 ns | 0.5193 ns |  1.42 |    0.39 |

<details>
<summary>Benchmark code</summary>

```csharp

[SimpleJob(RuntimeMoniker.Net60)]
[SimpleJob(RuntimeMoniker.Net70)]
[SimpleJob(RuntimeMoniker.Net80)]
[SimpleJob(RuntimeMoniker.NativeAot70)]
[SimpleJob(RuntimeMoniker.NativeAot80)]
public class Calls
{
    public interface IFoo
    {
        int Do();
    }

    public sealed class Foo : IFoo
    {
        public int Do() => 42;
    }
    
    public static IFoo _virtualFoo = new Foo();
    public static Foo _directFoo = new Foo();

    [Benchmark(Baseline = true)]
    public int CallDirect()
    {
        return Do(_directFoo);

        static int Do(Foo foo)
        {
            return foo.Do();
        }
    }
    
    [Benchmark]
    public int CallVirtual()
    {
        return Do(_virtualFoo);

        static int Do(IFoo foo)
        {
            return foo.Do();
        }
    }
}

```

</details>

mtmk: Quite a difference in benchmark results, double, triple in some cases. Definitely we must use it on hot paths. But I agree with you maybe not necessary for this scenario. Thanks for your hard work @jasper-d we really do appreciate your contribution.

mtmk (approved):
LGTM @caleblloyd can you have a look as well please?

caleblloyd (approved):

caleblloyd:
Looks good, I think that the Request-Reply methods should probably either All be in the interface, or All be in extensions of INatsConneciton.  But we can make a decision on that and move those around in a follow-up

PR #110, opened 2023-08-10
Connect auth test subscriber sync fix

mtmk:
Added subscription synchronization to avoid intermittent test failures. Actual failure was at the end of the test method where it was testing the re-connect, but to be consistent put same method of synchronization to the previous pub/sub assert as well.

Recent test reconnect failures due to subscription synchronization were exacerbated because of the change to the way subscriptions are restored on reconnect.


caleblloyd (approved):
LGTM

PR #111, opened 2023-08-11
Enable async for test wait signals

mtmk:
Make TaskCompletionSource used in wait signals run completions asynchronously to avoid potential deadlocks during test runs.

Note that this PR doesn't have any effect on the tests. It's not solving any current issues. It's just a precautionary fix for potential issues in general.


caleblloyd (approved):
LGTM

PR #112, opened 2023-08-14
Reconnect NatsSub commands

mtmk:
Collect all subscription and additional commands to be prioritized when reconnecting. Additional commands will be important for JetStream consumers to issue pull requests in order to reestablish their state on the server.


caleblloyd (approved):
LGTM

PR #113, opened 2023-08-14
Auth test output and proxy tidy-up

mtmk:
Fixed auth test output to return just the authorization method name rather than a dump of all variables.

Moved test proxy to its own file and suppressed unnecessary socket exception during connect and disconnect.


caleblloyd (approved):
LGTM

PR #114, opened 2023-08-15
JetStream consumer reconnect

mtmk:
When NATS client reconnects to the server JetStream pull requests must be restored along with subscriptions.


caleblloyd (approved):
LGTM

PR #121, opened 2023-08-31
consistent Nats class prefix

caleblloyd:
- Consistently use `Nats` prefix on public classes
- Change `Nkey` -> `NKey`


mtmk (approved):
LGTM

Issue #53, opened 2023-02-27
Support distributed tracing / OpenTelemetry


Kaarel:
Support distributed tracing with out of the box automatic instrumentation similar to ASP server, HttpClient etc https://opentelemetry.io/docs/instrumentation/net/libraries/

OpenTelemetry spec for messaging systems:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md

PR #124, opened 2023-09-01
Add OpenTelemetry Instrumentation

Cooksauce:
Implements: #53 
Background for OpenTelemetry in a messaging context: https://youtu.be/WsTYClGrOVI

As of now, this PR contains a rough implementation approach which is working in the sandbox. See below for how to run this. 

I have this set as a draft at the moment b/c I would like a bit of feedback on the general approach before moving forward and cleaning everything up well. Specifically, I'd like to know if the maintainers here approve of adding an `Activity` object to all of the commands. This was the best approach I came up with due to the pipelining nature of the actual msg publishing.
_Note that the objects should be null if no tracing listeners have been wired up._

#### Todo:
- [x] Propagation of trace context across publish --> subscribe
- [x] Create publisher span (i.e. `Activity`) on pub
- [x] Create consumer span (i.e. `Activity`) on sub
- [x] Complete implementation on all commands
- [x] Add jetstream receive activities
- ~~Surface "metrics"... counters, etc~~  (_tabled for follow up work_)

### Open Questions

1. Should jetstream control messages be surfaced or hidden? (maybe behind a different activity source?)
2. Should higher level messages (e.g. `kv put`, etc) be surfaced as such or as their underlying pub / sub commands? 

---
#### How to run in sandbox:
1. `docker-compose up` from `./sandbox` directory
2. run `Example.Core.SubscribeRaw` project
3. run `Example.Core.PublishModel` project
4. view traces in Jaeger UI at `http://localhost:16686`

#### Example trace at current state:
image


src/NATS.Client.Core/Telemetry.cs
```diff
@@ -0,0 +1,206 @@
...
+
+    public static ActivitySource NatsActivities { get; } = new(ActivitySourceName);
+
+    public static TracerProviderBuilder AddNatsInstrumentation(this TracerProviderBuilder builder)
+    {
+        builder.AddSource(ActivitySourceName);
+        return builder;
+    }
+
+    internal class Constants
```
Cooksauce, line 24: Need to agree on constant values. These probably should not change after release, b/c they could break monitoring rules & alerts users have set up.

sandbox/Example.Core.SubscribeRaw/Program.cs
```diff
@@ -1,25 +1,82 @@
...
+async Task ReadMessages(ChannelReader<NatsMsg> msgs, NatsConnection conn, CancellationToken ct)
 {
-    var data = Encoding.UTF8.GetString(msg.Data.ToArray());
-    Print($"[RCV] {msg.Subject}: {data}\n");
+    while (true)
+    {
+        try
+        {
+            var msg = await msgs.ReadAsync(ct);
+            using var _ = Activity.Current = msg.Activity;
```
Cooksauce, line 46: This isn't great, but since we have no way to know when `msg` goes out of scope, we have dispose/stop the activity via a `using` statement. Further, setting `Activity.Current` ensures context propagation on line 52.

Open to exposing this in a different way... but remember, `ChannelReader<NatsMsg>` can be consumed in parallel.

mtmk:
@Cooksauce thank you for the contribution!

src/NATS.Client.Core/NATS.Client.Core.csproj
```diff
@@ -19,6 +19,8 @@
           <PrivateAssets>all</PrivateAssets>
           <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
         </PackageReference>
+        <PackageReference Include="OpenTelemetry.Api" Version="1.4.0" />
+        <PackageReference Include="OpenTelemetry.SemanticConventions" Version="1.0.0-rc9.9" />
```
mtmk, line 23: Not knowing that much about OTEL I have a quick question here. I thought `Activity` (AFAIK, which is part of .NET now) was all we needed to enable OTEL. Why do we need these `OpenTelemetry` dependencies?

Cooksauce: I would agree these aren't needed. We should use the `string` values that are in the `SemanticConventions`, but these can easily be hardcoded to avoid taking that dependency.

The `OpenTelemetry.Api` dependency is only needed for the `.AddNatsInstrumentation()` extension method. This is not strictly necessary. I think the ideal place for this extension method would actually be in the contrib repo so that a nuget package can be created under the `OpenTelemetry...` top level namespace.
See: [github.com/open-telemetry/opentelemetry-dotnet-contrib/.../TracerProviderBuilderExtensions.cs](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.GrpcCore/TracerProviderBuilderExtensions.cs)

larsfjerm: Here's an example for how Npgsql does it (separate project/package): 
https://github.com/npgsql/npgsql/tree/main/src/Npgsql.OpenTelemetry

sandbox/Example.Core.SubscribeRaw/Program.cs
Cooksauce: I currently see two alternative options to this api here. @mtmk What are your thoughts on these two options?

## 1. Stop (i.e. "complete") the `Activity` when it is returned from `msgs.ReadAsync()`.
It is then on the user to instrument their application w/ some kind of `Processing` `Activity` (if they choose). I think we'd want to have a convenience method to propagate the context here. So user code might look something like:
```csharp
var msg = await msgs.ReadAsync(ct); // receive activity completes here

// ---- optional user-managed "process" activity  ----
// helper option A
using var processActivity = Activity.Current = msg.StartChildActivity("my_processing_activity");

// helper option B - user brings their own ActivitySource
using var processActivity = Activity.Current = MyActivitySource.StartActivity(
    "my_processing_activity",
    ActivityKind.Internal,
    parentContext: msg.GetActivityContext());
// ---------------------------------------------------


// ...do stuff with msg
``` 
#### Span Timing:
```markdown
| 'receive_message' span |
                          | 'my_processing_activity' span |
```


## 2. Wrap the message in a `scope` type which would be disposed by user code at end of processing.
This can be a `struct` to avoid an allocation. It might look something like this:
```csharp
using var msgScope = await msgs.ReadAsync(ct);

// ---- optional user-managed "process" activity  ----
// helper option A
using var processActivity = Activity.Current = msgScope.StartChildActivity("my_processing_activity");

// helper option B - user brings their own ActivitySource
using var processActivity = Activity.Current = MyActivitySource.StartActivity(
    "my_processing_activity",
    ActivityKind.Internal,
    parentContext: msgScope.GetActivityContext());
// -------------------------------------------------

var msg = msgScope.Msg;
// ...do stuff with msg
```
#### Span Timing:
```markdown
| ---- 'receive_message' span ------- |
     | 'my_processing_activity' span |
```

---
#### My Opinion
Personally I like the first option better because, in the case where the user doesn't want to deal w/ tracing, it doesn't get in their way. Further, this seems to better align w/ the OpenTelemetry [documentation and definitions](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/messaging/#consumer). It aligns with the example given [here](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/messaging/#apache-kafka-with-quarkus-or-spring-boot-example).

mtmk: I agree. First option seems better to me too.

mtmk:
@Cooksauce btw if you're not already on https://slack.nats.io `#dotnet` channel we'd love to have you there as well 💯

mtmk:
A few updates:
* should we have the activity attached on the Header? would that help implementing something like HTTP trace headers as well e.g. in request-reply?
* There is also some work to create an Aspire hosting add-on. Tracing and metrics are desirable there as well.
* what about metrics?

Cooksauce:
Rebased this on top of the `pipelines-send` branch from PR #303 since there is quite a bit of overlapping changes. So there's not actually 80 files changed here. _I'm assuming that branch is gonna merge at some point_ (need to rebase again then).

---
> - should we have the activity attached on the Header? would that help implementing something like HTTP trace headers as well e.g. in request-reply?

I may be missing understanding you here, but this is how the trace is currently propagated in the implementation. Specifically, [here](https://github.com/Cooksauce/nats.net.v2/blob/9623470525ab73bd4400d9194b9636938b77abb2/src/NATS.Client.Core/Telemetry.cs#L83) after rebase and changes. Are there other headers you're thinking to add?

---
> - There is also some work to create an Aspire hosting add-on. Tracing and metrics are desirable there as well.

Aspire OTEL is a direct implementation of the standard, so this plugs right with the standard OpenTelemetry exporter. Aspire just sets the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. 
_(screenshot of Aspire dashboard with latest changes from here below)_

---
> - what about metrics?

I think OpenTelemetry metrics export is a good idea. That said, the implementation for logging, tracing, and metrics is all pretty orthogonal in dotnet, so my opinion would be that be a separate, follow-up PR.


### Aspire Trace View
image

mtmk:
> I may be missing understanding you here, but this is how the trace is currently propagated in the implementation. Specifically, [here](https://github.com/Cooksauce/nats.net.v2/blob/9623470525ab73bd4400d9194b9636938b77abb2/src/NATS.Client.Core/Telemetry.cs#L83) after rebase and changes. Are there other headers you're thinking to add?

I was referring to passing activity down in headers instead of `NatsMsg` i.e. https://github.com/nats-io/nats.net.v2/pull/124/files#diff-82509a9a86e1e07fe3c25e4c89639eac80bb91a7dab2d1fafb46f786ae10232dR124

We also have an effort to minimize `NatsMsg` footprint since it's a struct in #282 

> I think OpenTelemetry metrics export is a good idea. That said, the implementation for logging, tracing, and metrics is all pretty orthogonal in dotnet, so my opinion would be that be a separate, follow-up PR.

yep, that makes sense.

> Aspire OTEL is a direct implementation of the standard, so this plugs right with the standard OpenTelemetry exporter

excellent! good to know. just wanted to bring that to your attention. the PR is here https://github.com/dotnet/aspire/pull/1175

mtmk:
@Cooksauce I also noticed @stebet is working on rabbitmq implementation https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261

martinjt:
Just been pointed here, love the work...

Noticed a point on dependencies. We're not planning on taking anything in Contrib for .NET unless there's a big reason to in terms of extra functionality needed. We advise that you create a .OpenTelemetry package in your own namespace for that.

Be careful taking a dependency on the semantic conventions package, we're introducing some change there are some point. Not to say you shouldn't, it just might be better not to take that dependency if you don't need to.

I'm more than happy to help with any questions you have around the spec and how to implement in the best way like I did with a few other OSS projects like Rabbit.

martinjt:
Had a quick look... couple of small tweaks you can make... 

check ActivitySource.HasListeners instead of getting the Activity and checking for null.

Also, the comment around propagation is yes, it should propagate the Activity.Current properties, regardless of whether your instrumentation is active.

Finally, when creating/starting activities, add as many tags as you can inside the create method, it better perf (upto 8 and it uses a different faster approach) and it also provides people with the ability to Sample better as the parameters in the create/start are all you get access to.

I do open office hours, feel free to book some time if I can help move it along!

mtmk:
Thank you @martinjt we're eager to learn from you and we'd love to take you up on your kind offer. I'm still learning, @Cooksauce is our expert here. btw I noticed we're a bit behind. I'll get the branch up to date as soon as I can.

mtmk:
> Noticed a point on dependencies. We're not planning on taking anything in Contrib for .NET unless there's a big reason to in terms of extra functionality needed. We advise that you create a .OpenTelemetry package in your own namespace for that.

Just created `NATS.Net.OpenTelemetry` package (empty shell at the moment). Let me know if you all happy with the package naming or prefer something else @Cooksauce @stebet @martinjt @caleblloyd

Cooksauce:
Reworked this quite a bit to address the discussion here, and to make it play more nicely with jetstream & the higher level features built on top of that. I think it's ready for a real review now with deeper scrutiny. 
- [x] activity context should propagate even when Nats telemetry is off
- [x] activity is moved into headers to reduce msg size (note that this has the effect of allocating `NatsHeaders` if they were otherwise not present)

---

> check ActivitySource.HasListeners instead of getting the Activity and checking for null.

@martinjt I'm curious on this one, is there a reason you mention this other than the allocation of parameters such as `tags` 
in `.StartActivity(...)`?

> Finally, when creating/starting activities, add as many tags as you can inside the create method, it better perf (upto 8 and it uses a different faster approach) and it also provides people with the ability to Sample better as the parameters in the create/start are all you get access to.

@martinjt Thanks for this tip! The source definitely looks more optimized.

--- 

@mtmk Decisions to make before moving forward:
1. Are the Nats folks open to tweaking `NatsMsg<T>` to expose `NatsConnection` so that these couple things can be addressed? https://github.com/Cooksauce/nats.net.v2/blob/611093cf416d6fde1360d94191e806b2360a0843/src/NATS.Client.Core/NatsMsg.cs#L174
2. Are we okay leaving out OTEL Baggage support similar to RabbitMQ? - [github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261#issuecomment-1875667625](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261#issuecomment-1875667625)

mtmk:
> @mtmk Decisions to make before moving forward:
> 
> 1. Are the Nats folks open to tweaking `NatsMsg<T>` to expose `NatsConnection` so that these couple things can be addressed? https://github.com/Cooksauce/nats.net.v2/blob/611093cf416d6fde1360d94191e806b2360a0843/src/NATS.Client.Core/NatsMsg.cs#L174

I think we can change the NatsMsg.ctor to take in a concrete NatsConnection since we do that pretty much with everything else (e.g. contexts, NatsJSMsg). For unit testing purposes we have NatsMsg-INatsMsg cast as an option. I don't think anyone will be passing interfaces to NatsMsg for unit testing if so I think it'd be reasonable to ask them to convert to accepting INatsMsg. So I'd say yes we can change NatsMsg to take NatsConnection (rather than INatsConnection) unless someone hollers otherwise cc @caleblloyd @sspates @to11mtm @paulomf

> 2. Are we okay leaving out OTEL Baggage support similar to RabbitMQ? - [github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261#issuecomment-1875667625](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261#issuecomment-1875667625)

No idea TBH. If it's good enough for Rabbit good for us too I'm guessing 😅
@stebet do you have an objection?

oising:
> > 2. Are we okay leaving out OTEL Baggage support similar to RabbitMQ? - [github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261#issuecomment-1875667625](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261#issuecomment-1875667625)
> 
> No idea TBH. If it's good enough for Rabbit good for us too I'm guessing 😅 @stebet do you have an objection?

Just my 2c, I think it's fine to leave it for another PR. The cooperative story behind OTEL's Baggage.* and Activity.Baggage.* APIs is vague at best, and frankly I could never get them to work reliably in tandem. Microsoft's legacy uses Trace-Context as a carrier, and OTEL use "baggage." Hopefully Microsoft will soon clear up the confusion caused by hiding their OTEL support inside of legacy APIs.

to11mtm:
> > @mtmk Decisions to make before moving forward:
> > 
> > 1. Are the Nats folks open to tweaking `NatsMsg<T>` to expose `NatsConnection` so that these couple things can be addressed? https://github.com/Cooksauce/nats.net.v2/blob/611093cf416d6fde1360d94191e806b2360a0843/src/NATS.Client.Core/NatsMsg.cs#L174
> 
> I think we can change the NatsMsg.ctor to take in a concrete NatsConnection since we do that pretty much with everything else (e.g. contexts, NatsJSMsg). For unit testing purposes we have NatsMsg-INatsMsg cast as an option. I don't think anyone will be passing interfaces to NatsMsg for unit testing if so I think it'd be reasonable to ask them to convert to accepting INatsMsg. So I'd say yes we can change NatsMsg to take NatsConnection (rather than INatsConnection) unless someone hollers otherwise cc @caleblloyd @sspates @to11mtm @paulomf
> 
> > 2. Are we okay leaving out OTEL Baggage support similar to RabbitMQ? - [github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261#issuecomment-1875667625](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261#issuecomment-1875667625)
> 
> No idea TBH. If it's good enough for Rabbit good for us too I'm guessing 😅 @stebet do you have an objection?

Thanks for the tag here. Thanks to a tangentially related conversation, I actually have two important-ish points/questions.

1: Can we get benchmarks for this before/after? To be clear we don't have to have telemetry actually 'wired up', moreso want to make sure perf impact is minimized for minimal cases.

2: Agree OTEL could be a separate PR if folks really need it (AFAIR it's the 'old thing' anyway) but no strong feelings either way.

martinjt:
I'm not sure what you mean by OTEL being the old way and leaving that to
another PR?

stebet:
Also the plan is not to leave baggage propagation out for the RabbitMQ client. It'll be added very soon, we just wanted to get the core ActivitySource implementation PR done first.

mtmk:
> Thanks for the tag here.

Thank you! I appreciate everyone's help here since my OTEL knowledge is quite limited until I can find some time to get up to speed with it all.

> Thanks to a tangentially related conversation, I actually have two important-ish points/questions.
> 
> 1: Can we get benchmarks for this before/after? To be clear we don't have to have telemetry actually 'wired up', moreso want to make sure perf impact is minimized for minimal cases.

My quick publish only benchmarks didn't show anything alarming but I plan to run more comprehensive benchmarks especially to make sure we're not affecting applications not using OTEL. Could you expand a little on what you meant by "we don't have to have it wired up" please?

> 
> 2: Agree OTEL could be a separate PR if folks really need it (AFAIR it's the 'old thing' anyway) but no strong feelings either way.

👍

oising:
> I'm not sure what you mean by OTEL being the old way and leaving that to another PR?

I think they mean "OTEL baggage," though I don't think that changes your sentiment. Baggage isn't an old thing, and is a first-class and orthogonal feature to Tracing.

src/NATS.Net.OpenTelemetry/OpenTelemetryTracingExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+public static class OpenTelemetryTracingExtensions
+{
+    /// <summary>
+    /// Enables trace collection on activity sources from the NATS.Client nuget package.
+    /// </summary>
+    /// <param name="builder"><see cref="TracerProviderBuilder"/> being configured.</param>
+    /// <param name="includeInternal">Include traces from internal messaging used for control flow and lower level usage during high level abstractions (e.g. JetStream, KvStore, etc)</param>
+    /// <returns>The same instance of <see cref="TracerProviderBuilder"/> to chain the calls.</returns>
+    public static TracerProviderBuilder AddNatsInstrumentation(this TracerProviderBuilder builder, bool includeInternal = false)
```
davidfowl, line 14: Is this library needed? It's wrapper around calling a method with these 2 strings.

to11mtm:
Unless Nats.Net.Core is somehow pulling in the `OpenTelemetry` package by
some other virtue it makes sense to me despite the low line count, as it
avoids unnecessary library churn for those who don't need it and/or those
who are hoping for constrained/IOT deploy scenarios.

Alas I'm semi out of pocket so can't easily check if there's a better spot
in existing project structure, if there is that may be a better option.

src/NATS.Net.OpenTelemetry/OpenTelemetryTracingExtensions.cs
martinjt: This is pretty standard. Providing a second package that provides a strongly typed extension to add the instrumentation.

What we've found is that people don't want the magic strings for `AddSource()` and they want the semantic meaning of "Adding instrumentation". It's also providing the semantics and documentation around what internal vs standard telemetry is.

Adding a second package means that you don't need pollute the main packages with OpenTelemetry. That said, that will be required for baggage to work properly at some point soon though.

davidfowl: I get why but IMO, I wouldn't use this package because it does so little.

martinjt:
The audience for the package is not really you or I.

There is a big audience for strongly typed abstractions like that, even if
it was for a single method.

From talking to people about this kind of thing, they feel adding the
string is a hack, as it could change etc.

It's lightweight, and shouldn't really change, so the maintainence overhead
being minimal means the advantage around adoption outweighs it in our
opinion on the Otel teams.

davidfowl:
<rant>

Well .NET did all of this work to make sure that library authors could use core primitives to emit traces, metrics and logs without taking on extra dependencies. Except these core libraries still need an extra companion package to call an extension method to turn on the telemetry. The story is incomplete, and we'll get to a point where that isn't needed. I hope these glue packages go away.

</rant>

martinjt:
There's lots of ways .NET could solve this, I'm surprised that the research
already done by the .NET hasn't already flagged this gap.

Until there is a better way, this is rhe recommended approach from
OpenTelemetry, if the .NET team has other recommendations then it would be
great to hear them as part of the OpenTelemetry SIGs which are the ones
guiding, you're welcome to join sometime and discuss it.

The only solution I've heard so far is to move the magic strings to app
config (like the idea for meters), which doesn't solve the problem that
developers want strongly typed extensions.

I suppose its upto the team here whether they want to follow the
conventions, or stick with the advice of using the strings.

iby-dev:
Are there any updates in this space? The target branch contains  a few nice-to-haves? I don't think these changes have made into the main branch.

mtmk:
> Are there any updates in this space? The target branch contains a few nice-to-haves? I don't think these changes have made into the main branch.

@iby-dev thanks for your interest. please feel free to open a new issue or PR with features you want to see implemented.

PR #126, opened 2023-09-06
Jetstream preview release 1

mtmk:
* Consume and fetch error handling removes
* Consume and fetch now throws exception on JS terminal errors
* Removed ErrorHandler option
* Async enumerable initial List implementation
* Domain option added to JSOpts
* Logging improvements for testing
* Option validations now throw exception


src/NATS.Client.JetStream/Internal/NatsJSOpsDefaults.cs
```diff
@@ -10,13 +10,11 @@ internal static class NatsJSOpsDefaults
     private static readonly TimeSpan HeartbeatMin = TimeSpan.FromSeconds(.5);
 
     internal static (long MaxMsgs, long MaxBytes, long ThresholdMsgs, long ThresholdBytes) SetMax(
```
caleblloyd, line 12: I'm a little late in catching this, but I think this Class Name is supposed to be `NatsJSOptsDefaults`? (missing t in Opts)

Not a huge deal since it's internal

src/NATS.Client.JetStream/NatsJSOpts.cs
```diff
@@ -4,19 +4,29 @@ namespace NATS.Client.JetStream;
 
 public record NatsJSOpts
 {
-    public NatsJSOpts(NatsOpts opts, string? apiPrefix = default, int? maxMsgs = default, AckOpts? ackOpts = default, string? inboxPrefix = default)
+    public NatsJSOpts(NatsOpts opts, string? apiPrefix = default, string? domain = default, AckOpts? ackOpts = default, string? inboxPrefix = default)
     {
         ApiPrefix = apiPrefix ?? "$JS.API";
-        MaxMsgs = maxMsgs ?? 1000;
         AckOpts = ackOpts ?? new AckOpts(opts.WaitUntilSent);
         InboxPrefix = inboxPrefix ?? opts.InboxPrefix;
```
caleblloyd, line 11: Should we get rid of `InboxPrefix` and always use the one from `NatsOpts`?  I can't see a reason why we'd need a different one for the JS Context

caleblloyd (commented):
Looks good, a couple small comments

caleblloyd (approved):
LGTM

src/NATS.Client.JetStream/NatsJSOpts.cs
mtmk: Agreed.

PR #127, opened 2023-09-07
Add connection state to nats connection interface

vahpetr:
**Summary**
This Pull Request aims to extend the `INatsConnection` interface by adding a new property `NatsConnectionState ConnectionState { get; }`. This new property will provide real-time information about the state of the NATS connection.

**Rationale**
1. Increased Transparency
Knowing the state of the NATS connection at any point in time is crucial for efficiently managing resources and ensuring that messages are successfully published or consumed. By exposing this state through `INatsConnection`, we allow all client code to make informed decisions based on the state.

2. Enhanced Control Flow
In multi-threaded and distributed applications like ours, it's vital to have more control over asynchronous operations. The ConnectionState property can serve as a gatekeeper before performing certain actions, like sending messages, thus minimizing failed attempts and errors.

For example, in the given Producer class:

```csharp
while (!cancellationToken.IsCancellationRequested)
{
    if (_nats.ConnectionState != NatsConnectionState.Open)
    {
        await Task.Delay(1000, cancellationToken);
        continue;
    }
    // ... (publish message)
}
```
Here, we can efficiently pause and resume the message-publishing loop based on the NATS connection state.

3. Simplified Debugging and Monitoring
By exposing the NATS connection state, we make it easier to debug issues related to connectivity. It becomes straightforward to write monitoring tools that can alert developers or take corrective action when the NATS connection transitions to an undesirable state.

**Conclusion**
Adding a `NatsConnectionState ConnectionState { get; }` property to the `INatsConnection` interface is a minimal yet powerful extension that can benefit virtually every NATS client application. It simplifies control flow, makes debugging easier, and increases transparency in system operations.

### DIFF
--- a/src/NATS.Client.Core/INatsConnection.cs
+++ b/src/NATS.Client.Core/INatsConnection.cs
@@ -6,6 +6,11 @@ namespace NATS.Client.Core;
 public interface INatsConnection
 {
     /// <summary>
+    /// Gets the current state of the NATS connection.
+    /// </summary>
+    NatsConnectionState ConnectionState { get; }
+
+    /// <summary>
     /// Send PING command and await PONG. Return value is similar as Round Trip Time (RTT).
     /// </summary>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the command.</param>


mtmk:
@vahpetr thank you for your contribution and detailed explanation. `ConnectionState` is already a public property of `NatsConnection` so I don't see an issue. @caleblloyd what do you think?

Just to note, connection is initiated when the first call is made. I haven't tried it but in your while loop example, it might not publish at all if no call (e.g. ping, pub, sub) was made using the same `NatsConnection` object before. A first connection must be in place for retry mechanism to kick-in when disconnected.

mtmk:
Should we expose `ServerInfo` as well?

vahpetr:
> Should we expose `ServerInfo` as well?

Thanks for the quick feedback. You're right about the connection initiation. However, the `BackgroundService` in my example operates continuously, and pausing it isn't an option. That's why exposing `ConnectionState` in the interface would help manage the flow better. And yes, adding ServerInfo sounds like a great idea. Should I add PR to `ServerInfo`?

caleblloyd:
Don't think we should put them on the Interface.  An example of why that might not be a good idea is the `AlterNats` sharding Nats Connection, or a Nats Connection pool.  There is no single connection state, the backing implementation has a pool of connections.

`ServerInfo` is Internal right now - I guess we could make it public

mtmk:
@vahpetr I think changing the public interface isn't going to work because of the points raised by @caleblloyd. In your scenario you might want to use `Ping` to achieve a similar function.

I also think you have a good point about connection management. Currently it's an internal function and not extensible, of course feel free to look at that area as well. BTW we'd love to have you on https://slack.nats.io `#dotnet` channel if you aren't already there. So please feel free to start an issue, discussion (here on GitHub or Slack) to explore this topic further.

PR #128, opened 2023-09-07
Options tidy-up

mtmk:
* Removed variable PubOpts Headers and ReplyTo
* Added headers and replyTo params to Publish methods
* Make all Opts classes (not structs)
* Added default options for JS context

### DIFF
--- a/src/NATS.Client.Core/INatsConnection.cs
+++ b/src/NATS.Client.Core/INatsConnection.cs
@@ -16,12 +16,12 @@ public interface INatsConnection
     /// </summary>
     /// <param name="subject">The destination subject to publish to.</param>
     /// <param name="payload">The message payload data.</param>
-    /// <param name="replyTo">Optional reply-to subject.</param>
     /// <param name="headers">Optional message headers.</param>
+    /// <param name="replyTo">Optional reply-to subject.</param>
     /// <param name="opts">A <see cref="NatsPubOpts"/> for publishing options.</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the command.</param>
     /// <returns>A <see cref="ValueTask"/> that represents the asynchronous send operation.</returns>
-    ValueTask PublishAsync(string subject, ReadOnlySequence<byte> payload = default, string? replyTo = default, NatsHeaders? headers = default, NatsPubOpts? opts = default, CancellationToken cancellationToken = default);
+    ValueTask PublishAsync(string subject, ReadOnlySequence<byte> payload = default, NatsHeaders? headers = default, string? replyTo = default, NatsPubOpts? opts = default, CancellationToken cancellationToken = default);
 
     /// <summary>
     /// Publishes the message payload to the given subject name, optionally supplying a reply subject.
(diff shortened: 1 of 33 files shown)


src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -17,10 +17,12 @@ public interface INatsConnection
     /// </summary>
     /// <param name="subject">The destination subject to publish to.</param>
     /// <param name="payload">The message payload data.</param>
+    /// <param name="replyTo">Optional reply-to subject.</param>
+    /// <param name="headers">Optional message headers.</param>
     /// <param name="opts">A <see cref="NatsPubOpts"/> for publishing options.</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the command.</param>
     /// <returns>A <see cref="ValueTask"/> that represents the asynchronous send operation.</returns>
-    ValueTask PublishAsync(string subject, ReadOnlySequence<byte> payload = default, NatsPubOpts? opts = default, CancellationToken cancellationToken = default);
+    ValueTask PublishAsync(string subject, ReadOnlySequence<byte> payload = default, string? replyTo = default, NatsHeaders? headers = default, NatsPubOpts? opts = default, CancellationToken cancellationToken = default);
```
caleblloyd, line 25: do you think there is a benefit to having `headers` appear before `replyTo`, as it is likely a more common option than `replyTo` - for those who call with a positional syntax

I can also see the argument that `replyTo` should come first, because it was defined in the protocol before Headers were

I don't have a strong preference but thought it was worth bringing up

src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -97,6 +115,8 @@ public interface INatsConnection
     ValueTask<NatsMsg<TReply?>?> RequestAsync<TRequest, TReply>(
         string subject,
         TRequest? data,
+        string? queueGroup = default,
```
caleblloyd, line 118: I don't think `queueGroup` would ever get used - since `Request` makes a new inbox and doesn't expose it, the subscribe that it creates is likely the only subscriber that will ever listen, so there is not a need for a work queue

mtmk: Good catch. Removed.

src/NATS.Client.Core/INatsConnection.cs
mtmk: This is a good point. Like you said, I was thinking of the protocol order but that shouldn't be a concern here. Usability is more important and I agree headers are more likely to be used as metadata than reply-to.

mtmk: Fixed. Also swapped them for `NatsMsg.ReplyAsync()` methods.

src/NATS.Client.Core/Internal/SubscriptionManager.cs
```diff
@@ -47,15 +47,20 @@ public SubscriptionManager(NatsConnection connection, string inboxPrefix)
...
     internal InboxSubBuilder InboxSubBuilder { get; }
 
-    public async ValueTask SubscribeAsync(string subject, NatsSubOpts? opts, NatsSubBase sub, CancellationToken cancellationToken)
+    public async ValueTask SubscribeAsync(string subject, string? queueGroup, NatsSubOpts? opts, NatsSubBase sub, CancellationToken cancellationToken)
     {
         if (IsInboxSubject(subject))
         {
+            if (queueGroup != null)
+            {
+                throw new NatsException("Inbox subscriptions don't support queue groups");
```
caleblloyd, line 56: I don't believe this is a technical limitation, is it? It does seem uncommon but I don't think it's worth throwing an exception on.

I was just thinking the Request abstraction would never hit this case because the Inbox Subject is never exposed. But someone could come up with their own pattern that does use an inbox subject with a queue group

mtmk: Fixed.

caleblloyd (approved):
LGTM

PR #132, opened 2023-09-21
Initial Key Value Store implementation

mtmk:
Initial KV implementation and internal tidy-up and fixes.

Kv implementation:
* Store create
* put/get
* watch

Missing KV features for another PR:
* Mirror support
* Get keys and potentially other methods

Internal tidy-up and fixes:
* Internal subscription signature simplification
* Mux-inbox reconnect issue fix
* Raw Serializer `IMemoryOwner<byte>` support
* Removed consumer creation restrictions to support KV push consumers
* Prevent JetStream API calls from hanging forever by adding cancellation timeout


src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
```diff
@@ -0,0 +1,401 @@
...
+            Name = Consumer,
+            DeliverPolicy = ConsumerConfigurationDeliverPolicy.all,
+            AckPolicy = ConsumerConfigurationAckPolicy.none,
+            DeliverSubject = _sub.Subject,
+            FilterSubject = _filter,
+            FlowControl = true,
+            IdleHeartbeat = _idleHbNanos,
+            AckWait = _ackWaitNanos,
+            MaxDeliver = 1,
+            MemStorage = true,
```
caleblloyd, line 345: `MemStorage` may be an issue - some server configurations don't have Memory Store enabled, which will make this fail.  Would be best to leave this out and default to the Stream's backing store

mtmk: AFAIK this is explicitly set by [Go](https://github.com/nats-io/nats.go/blob/main/js.go#L1567) (also Rust I think) clients and in [Ordered Consumer ADR](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-17.md#consumer-configuration-checks) I believe. I don't know the reason. @piotrpio?

caleblloyd: Maybe `MemStorage` still works for consumers, even if `max_memory_store` is set to 0 then.  Otherwise this would always fail on NGS, and I am assuming if Go is doing it then someone is successfully using this with NGS :smile:

caleblloyd (approved):
LGTM

PR #135, opened 2023-09-29
Use supported runtime/language pair

jasper-d:
Using `latest` with .NET SDK 7 installed uses C# 11 which isn't supported on .NET 6.

> Choosing a language version newer than the default can cause hard to diagnose compile-time and runtime errors.
>
> [C# 11](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11) is supported only on .NET 7 and newer versions. [C# 10](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-10) is supported only on .NET 6 and newer versions.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version


mtmk (approved):
LGTM

mtmk:
Thank you @jasper-d 👍

PR #136, opened 2023-09-29
Serialization tests

jasper-d:


mtmk:
Once #143 is merged these tests should pass.

mtmk:
Also #144 is merged. Should get better exceptions now.

PR #139, opened 2023-09-29
Add JetStream metadata to msg parsed from reply string

simonhoss:
Hey

This PR adds the sequence and DateTime info to the msg. 

Any feedback is appreciated.


mtmk:
Thank you @simonhoss great to see this because it's implementing a missing feature in JetStream.

My initial thoughts are:

* We don't want to penalize applications which are not using JetStream but using the reply subject for other patterns of messaging,
* The feature really belongs to JetStream project,
* `NatsMsg` is a struct so we need to be careful about enlarging it,
* Have you considered calculating it on demand? (I'll also check with other client maintainers see how they're doing it)

simonhoss:
Thanks for the feedback. When I started the implementation I didn't realize that there are two msg's in (`NatsMsg` and `NatsJSMsg`). With this new knowledge and your feedback, I was thinking about to implement it only in the NatsJSMsg with a lazy implementation. That means that the Seq and DateTime gets only calculated if someone wants to know it.

What do you think?

mtmk:
Brilliant! Is there a way we can retrieve the new properties on NatsJSMsg without storing them on NatsMsg? Because NatsMsg is a struct it means every message needs to be copied so we want to keep it as small as possible 😉 

BTW @simonhoss please sign your commits

simonhoss:
I added a new approach, which doesn't touch the `NatsMsg` and only calculates the `Sequence` and `DateTime`when someone really consumes it.

src/NATS.Client.JetStream/Internal/ReplyToDateTimeAndSeq.cs
```diff
@@ -0,0 +1,28 @@
...
+        if (reply == null)
+        {
+            return (null, null);
+        }
+
+        var ackSeperated = reply.Split(".");
+
+        // It must be seperated by 9 dots as defined in
+        // https://docs.nats.io/reference/reference-protocols/nats_api_reference#acknowledging-messages
+        if (ackSeperated.Length != 9)
```
mtmk, line 16: This has couple of variations actually. See the design doc:
https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-15.md#jsack

you can also check the Go client implementation if you like:
https://github.com/nats-io/nats.go/blob/main/internal/parser/parse.go#L57

src/NATS.Client.JetStream/Internal/ReplyToDateTimeAndSeq.cs
```diff
@@ -0,0 +1,28 @@
+namespace NATS.Client.JetStream.Internal;
+
+internal static class ReplyToDateTimeAndSeq
+{
+    internal static (DateTimeOffset? DateTime, long? Seq) Parse(string? reply)
```
mtmk, line 5: This should return a new `MsgMetadata` class with all the fields populated. For example:

```csharp
class NatsJSMsgMetadata
{
  SequencePair Sequence;
  ulong NumDelivered;
  ulong NumPending;
  DateTimeOffset Timestamp;
  string Stream;
  string Consumer;
  string Domain;
}

class NatsJSSequencePair
{
  ulong Stream;
  ulong Consumer;
}
```

See also Go client implementation https://github.com/nats-io/nats.go/blob/main/jetstream/message.go#L64

mtmk (changes requested):
Looks good 👍 just needs a couple of changes to complete the implementation.

src/NATS.Client.JetStream/Internal/ReplyToDateTimeAndSeq.cs
simonhoss: I was only focusing on Timestamp and Stream Sequence, because this is what I need in my current project.
But you're right better make it right ;)

src/NATS.Client.JetStream/Internal/ReplyToDateTimeAndSeq.cs
simonhoss: Oh I didn't know that there is also a v2. Funny that the official docs says nothing about it.

simonhoss:
New version pushed. I think it should now parse both versions correct.

Also, Im not quite sure if struct makes sense for `NatsJSMsgMetadata`, but `NatsJSMsg` is also a struct so I took the same.

mtmk (approved):
LGTM

Thank you for the contribution @simonhoss 🥇

mtmk:
Also @simonhoss please feel free to join https://slack.nats.io `#dotnet` channel if you haven't already done so. I'd love to see you there too 💯

simonhoss:
Thanks a lot!

PR #143, opened 2023-10-05
Support all binary payload APIs with only generics

mtmk:
With this proposal we reduce the API footprint and make the default behaviour work as expected when sending byte types and strings:

```csharp
await using var sub = await SubscribeAsync<T>(string subject);

await PublishAsync<T>(string subject, T data);

await PublishAsync(string subject); // send empty message / sentinel
```

The default serializer is smart enough to do the *right thing* handling known types avoiding surprises, especially when publishing.

```csharp
// Send as binary data
await PublishAsync(string subject, new byte[] {1,2,3});
await PublishAsync(string subject, new Memory(bytes));
await PublishAsync(string subject, myMemory); // IMemoryOwner

// JSON serialized
await PublishAsync(string subject, complexObject);
```

The idea is to reduce the code duplication between generic calls and non-generic calls using binary data as payload and make the defaults work for the vast majority of the cases while having the flexibility of defining a custom serializer.

Related to #137


mtmk:
Performance considerations for publish:

Without this change:
```
| Method             | Iter | Mean        | Error         | StdDev      | Gen0   | Allocated |
|------------------- |----- |------------:|--------------:|------------:|-------:|----------:|
| WaitUntilSentTrue  | 64   |  2,995.8 us |   1,481.06 us |    81.18 us |      - |    6987 B |
| WaitUntilSentFalse | 64   |    156.2 us |      21.80 us |     1.19 us |      - |     591 B |
| WaitUntilSentTrue  | 512  | 26,541.7 us | 115,608.79 us | 6,336.91 us |      - |   53614 B |
| WaitUntilSentFalse | 512  |    439.3 us |      63.39 us |     3.47 us | 0.4883 |    9575 B |
| WaitUntilSentTrue  | 1024 | 45,792.0 us |  29,602.46 us | 1,622.61 us |      - |  106941 B |
| WaitUntilSentFalse | 1024 |    711.4 us |      64.15 us |     3.52 us | 2.9297 |   46649 B |
```

With this change:
```
| Method             | Iter | Mean        | Error        | StdDev      | Gen0   | Allocated |
|------------------- |----- |------------:|-------------:|------------:|-------:|----------:|
| WaitUntilSentTrue  | 64   |  2,894.7 us |  1,268.27 us |    69.52 us |      - |    6990 B |
| WaitUntilSentFalse | 64   |    161.6 us |     23.09 us |     1.27 us |      - |     520 B |
| WaitUntilSentTrue  | 512  | 21,691.3 us | 11,121.19 us |   609.59 us |      - |   53606 B |
| WaitUntilSentFalse | 512  |    438.5 us |     70.65 us |     3.87 us | 0.4883 |   10168 B |
| WaitUntilSentTrue  | 1024 | 44,436.2 us | 37,207.27 us | 2,039.46 us |      - |  106921 B |
| WaitUntilSentFalse | 1024 |    738.8 us |    111.61 us |     6.12 us | 2.9297 |   44885 B |
```

src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -45,6 +24,8 @@ public interface INatsConnection
     /// <returns>A <see cref="ValueTask"/> that represents the asynchronous send operation.</returns>
     ValueTask PublishAsync<T>(string subject, T data, NatsHeaders? headers = default, string? replyTo = default, NatsPubOpts? opts = default, CancellationToken cancellationToken = default);
 
+    ValueTask PublishAsync(string subject, NatsHeaders? headers = default, string? replyTo = default, NatsPubOpts? opts = default, CancellationToken cancellationToken = default);
```
caleblloyd, line 27: We should probably make this `PublishVoidAsync`, `PublishEmptyAsync`, `PublishSentinelAsync`, or something along those lines?  Any of those names resonate with you?

mtmk: What's your preference?

caleblloyd: On the flip side I had never heard of `Sentinel` before getting involved in the NATS ecosystem :slightly_smiling_face:

mtmk: > On the flip side I had never heard of `Sentinel` before getting involved in the NATS ecosystem 🙂

Same here. I should've mentioned where I heard the term :-P

caleblloyd: I guess `Empty` is my preference.  The generic equivalent would probably be an `Array.Empty<byte>()`.

But I am now second-guessing providing only a `Publish` method that deals with `Empty`.  Since we don't have a Subscribe Empty, Request Empty, etc.  Is it more confusing to have just a Publish Empty on the interface?

sandbox/Example.Core.SubscribeRaw/Program.cs
```diff
@@ -11,11 +11,11 @@
 
 Print($"[SUB] Subscribing to subject '{subject}'...\n");
 
-var sub = await connection.SubscribeAsync(subject);
+var sub = await connection.SubscribeAsync<byte[]>(subject);
 
 await foreach (var msg in sub.Msgs.ReadAllAsync())
 {
-    var data = Encoding.UTF8.GetString(msg.Data.ToArray());
+    var data = Encoding.UTF8.GetString(msg.Data!);
```
caleblloyd, line 18: Why is the [null-forgiving operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving) needed here?  Is `SubscribeAsync<T>` eventually returning `T?` instead?  Would be nice if we could avoid that

caleblloyd (approved):
Excellent!  I really like how it simplifies the codebase.

Left one comment, but don't let that hold up merging this PR

sandbox/Example.Core.SubscribeRaw/Program.cs
mtmk: Default values in general yield empty/sentinel payloads that's why there are `T?`s in a lot of places. Do you think we shouldn't allow nulls? How would we handle sending and receiving empty payloads?

caleblloyd: True, I guess there is no way to avoid it for reference types when `default` may be retruned.  Seeing as [System.Text.JSON even does it](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-7.0#system-text-json-jsonserializer-deserialize-1(system-readonlyspan((system-byte))-system-text-json-serialization-metadata-jsontypeinfo((-0))))

PR #145, opened 2023-10-05
Micro benchmarks tidy up

mtmk:
* Added switcher so we can add more benchmarks
* Added a benchmark comparing pub wait-until-sent option


caleblloyd (approved):
LGTM

PR #146, opened 2023-10-05
More publish overloads for binary payloads and SubscribeAllAsync method

mtmk:
This fixes the issue when `byte[]`, `Memory<byte>` and `ReadOnlyMemory<byte>` payloads are passed to publish methods.

Also adds the `SubscribeAllAsync()` method for convenient enumeration for subscribers.


mtmk:
Just complicating the API too much for what it offers. They can be implemented as extension methods later by the library or end devs.

PR #147, opened 2023-10-06
Implement NUID

### LINKED ISSUE #34: NUID Port

jasper-d:
Fixes #34 

```
| Method               | Job           | Runtime       | Mean     | Error    | StdDev   | Median   | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|--------------------- |-------------- |-------------- |---------:|---------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| TryWriteNuid         | .NET 6.0      | .NET 6.0      | 31.00 ns | 0.365 ns | 0.305 ns | 31.06 ns |  0.99 |    0.03 |      - |         - |          NA |
| NewInbox_ShortPrefix | .NET 6.0      | .NET 6.0      | 51.74 ns | 1.078 ns | 1.476 ns | 51.90 ns |  1.64 |    0.07 | 0.0612 |     128 B |          NA |
| NewInbox_LongPrefix  | .NET 6.0      | .NET 6.0      | 76.69 ns | 1.568 ns | 2.662 ns | 76.52 ns |  2.44 |    0.12 | 0.1988 |     416 B |          NA |
| TryWriteNuid         | .NET 7.0      | .NET 7.0      | 31.61 ns | 0.661 ns | 0.926 ns | 31.21 ns |  1.00 |    0.00 |      - |         - |          NA |
| NewInbox_ShortPrefix | .NET 7.0      | .NET 7.0      | 54.38 ns | 1.116 ns | 1.044 ns | 54.72 ns |  1.74 |    0.06 | 0.0612 |     128 B |          NA |
| NewInbox_LongPrefix  | .NET 7.0      | .NET 7.0      | 77.58 ns | 0.909 ns | 0.850 ns | 77.68 ns |  2.48 |    0.08 | 0.1988 |     416 B |          NA |
| TryWriteNuid         | .NET 8.0      | .NET 8.0      | 25.50 ns | 0.499 ns | 0.390 ns | 25.58 ns |  0.82 |    0.03 |      - |         - |          NA |
| NewInbox_ShortPrefix | .NET 8.0      | .NET 8.0      | 46.84 ns | 0.986 ns | 1.317 ns | 46.94 ns |  1.48 |    0.06 | 0.0612 |     128 B |          NA |
| NewInbox_LongPrefix  | .NET 8.0      | .NET 8.0      | 66.03 ns | 0.838 ns | 0.784 ns | 66.22 ns |  2.11 |    0.07 | 0.1988 |     416 B |          NA |
| TryWriteNuid         | NativeAOT 8.0 | NativeAOT 8.0 | 25.15 ns | 0.495 ns | 0.463 ns | 25.33 ns |  0.80 |    0.03 |      - |         - |          NA |
| NewInbox_ShortPrefix | NativeAOT 8.0 | NativeAOT 8.0 | 45.00 ns | 0.945 ns | 1.843 ns | 44.25 ns |  1.44 |    0.06 | 0.0612 |     128 B |          NA |
| NewInbox_LongPrefix  | NativeAOT 8.0 | NativeAOT 8.0 | 56.34 ns | 1.139 ns | 1.066 ns | 55.62 ns |  1.80 |    0.07 | 0.1989 |     416 B |          NA |
```

Opted to go with base-62 here to follow nuid.go. base-64 (as implemented and reviewed in https://github.com/nats-io/nats.net/pull/360) would avoid some bias and be about twice as fast.

### DIFF
--- a/src/NATS.Client.Core/Internal/NuidWriter.cs
+++ b/src/NATS.Client.Core/Internal/NuidWriter.cs
@@ -1,0 +1,140 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics;
+using System.Security.Cryptography;
+
+namespace NATS.Client.Core.Internal;
+
+[SkipLocalsInit]
+internal sealed class NuidWriter
+{
+    private const nuint BASE = 62;
+    private const ulong MAX_SEQUENTIAL = 839299365868340224; // 62^10   // 0x1000_0000_0000_0000; // 64 ^10
+    private const uint PREFIX_LENGTH = 12;
+    private const nuint SEQUENTIAL_LENGTH = 10;
+    private const int MIN_INCREMENT = 33;
+    private const int MAX_INCREMENT = 333;
+    internal const nuint NUID_LENGTH = PREFIX_LENGTH + SEQUENTIAL_LENGTH;
+
+    [ThreadStatic]
+    private static NuidWriter? t_writer;
+
+    // TODO: Use UTF8 string literal when upgrading to .NET 7+
+    private static ReadOnlySpan<char> Digits => "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+
+    private char[] _prefix;
+    private ulong _increment;
+    private ulong _sequential;
+
+    internal static int PrefixLength => (int)PREFIX_LENGTH;
+
+    private NuidWriter()
+    {
+        Refresh(out _);
+    }
+
+    public static bool TryWriteNuid(Span<char> nuidBuffer)
+    {
+        if(t_writer is not null)
+        {
+            return t_writer.TryWriteNuidCore(nuidBuffer);
+        }
+
+        return InitAndWrite(nuidBuffer);
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static bool InitAndWrite(Span<char> span)
+    {
+        t_writer = new NuidWriter();
+        return t_writer.TryWriteNuidCore(span);
+    }
+
+    private bool TryWriteNuidCore(Span<char> nuidBuffer)
+    {
+        ulong sequential = _sequential += _increment;
+
+        if(sequential < MAX_SEQUENTIAL)
+        {
+            return TryWriteNuidCore(nuidBuffer, _prefix, sequential);
+        }
+
+        return RefreshAndWrite(nuidBuffer);
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        bool RefreshAndWrite(Span<char> buffer)
+        {
+            char[] prefix = Refresh(out sequential);
+            return TryWriteNuidCore(buffer, prefix, sequential);
+        }
+    }
+
+    private static bool TryWriteNuidCore(Span<char> buffer, Span<char> prefix, ulong sequential)
+    {
+        if ((uint)buffer.Length < NUID_LENGTH || prefix.Length != PREFIX_LENGTH || (uint)prefix.Length > (uint)buffer.Length)
+        {
+            return false;
+        }
+
+        Unsafe.CopyBlockUnaligned(ref Unsafe.As<char, byte>(ref buffer[0]), ref Unsafe.As<char, byte>(ref prefix[0]), PREFIX_LENGTH * sizeof(char));
+
+        // NOTE: We must never write to digitsPtr!
+        ref char digitsPtr = ref MemoryMarshal.GetReference(Digits);
+
+        for(nuint i = PREFIX_LENGTH; i < NUID_LENGTH; i++)
+        {
+            nuint digitIndex = (nuint)(sequential % BASE);
+            Unsafe.Add(ref buffer[0], i) = Unsafe.Add(ref digitsPtr, digitIndex);
+            sequential /= BASE;
+        }
+
+        return true;
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    [MemberNotNull(nameof(_prefix))]
+    private char[] Refresh(out ulong sequential)
+    {
+        char[] prefix = _prefix = GetPrefix();
+        _increment = GetIncrement();
+        sequential = _sequential = GetSequential();
+        return prefix;
+    }
+
+    private static uint GetIncrement()
+    {
+        return (uint)Random.Shared.Next(MIN_INCREMENT, MAX_INCREMENT + 1);
+    }
+
+    private static ulong GetSequential()
+    {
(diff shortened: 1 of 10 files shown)


src/NATS.Client.Core/NatsConnection.RequestReply.cs
```diff
@@ -61,6 +64,42 @@ public partial class NatsConnection
         }
     }
 
+    [SkipLocalsInit]
+    private static string NewInbox(ReadOnlySpan<char> prefix)
```
jasper-d, line 68: This method is a bit of a mess. It is rather complicated and still slow and allocates quite heavily. Allocations could be reduced by renting buffers instead of new'ing them, but that reduces throughput.

Maybe it would be possible to not operate with strings (apart from public interfaces) for inbox prefix/subjects and instead use only `byte[]`. But that would be a more invasive change.

src/NATS.Client.Core/Internal/NuidWriter.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        Unsafe.CopyBlockUnaligned(ref Unsafe.As<char, byte>(ref buffer[0]), ref Unsafe.As<char, byte>(ref prefix[0]), PrefixLength * sizeof(char));
+
+        // NOTE: We must never write to digitsPtr!
+        ref var digitsPtr = ref MemoryMarshal.GetReference(Digits);
+
+        for (nuint i = PrefixLength; i < NuidLength; i++)
+        {
+            var digitIndex = (nuint)(sequential % Base);
+            Unsafe.Add(ref buffer[0], i) = Unsafe.Add(ref digitsPtr, digitIndex);
```
jasper-d, line 58: The usage of `Unsafe` is necessary here to avoid range checks, both when accessing buffer and Digits.

src/NATS.Client.Core/NatsConnection.RequestReply.cs
```diff
@@ -61,6 +64,42 @@ public partial class NatsConnection
...
+    {
+        Span<char> buffer = stackalloc char[64];
+        var separatorLength = prefix.Length > 0 ? 1u : 0u;
```
jasper-d, line 71: I'm not sure if support for empty prefixes is really needed, but right now nothing is preventing users from doing it.

src/NATS.Client.Core/NatsConnection.RequestReply.cs
jasper-d: E.g. this alternative implementation which looks much more reasonable
```csharp
    [SkipLocalsInit]
    private static string NewInbox(ReadOnlySpan<char> prefix)
    {
        Span<char> buffer = stackalloc char[22];
        NuidWriter.TryWriteNuid(buffer);

        if (prefix.IsEmpty)
        {
            return new string(buffer);
        }
        else
        {
            return $"{prefix}.{buffer}";
        }
    }
```

is slower but also reduces the allocations for long prefixes:
```
| Method               | Job           | Runtime       | Mean     | Error    | StdDev   | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|--------------------- |-------------- |-------------- |---------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| NewInbox_ShortPrefix | .NET 6.0      | .NET 6.0      | 79.85 ns | 1.617 ns | 2.370 ns |  0.91 |    0.05 | 0.0612 |     128 B |        1.00 |
| NewInbox_ShortPrefix | .NET 7.0      | .NET 7.0      | 87.93 ns | 1.789 ns | 2.732 ns |  1.00 |    0.00 | 0.0612 |     128 B |        1.00 |
| NewInbox_ShortPrefix | .NET 8.0      | .NET 8.0      | 72.40 ns | 1.353 ns | 1.266 ns |  0.82 |    0.03 | 0.0612 |     128 B |        1.00 |
| NewInbox_ShortPrefix | NativeAOT 8.0 | NativeAOT 8.0 | 66.20 ns | 1.358 ns | 1.270 ns |  0.75 |    0.03 | 0.0612 |     128 B |        1.00 |
|                      |               |               |          |          |          |       |         |        |           |             |
| NewInbox_LongPrefix  | .NET 6.0      | .NET 6.0      | 82.68 ns | 0.958 ns | 1.025 ns |  0.97 |    0.02 | 0.0994 |     208 B |        1.00 |
| NewInbox_LongPrefix  | .NET 7.0      | .NET 7.0      | 85.93 ns | 0.881 ns | 0.688 ns |  1.00 |    0.00 | 0.0994 |     208 B |        1.00 |
| NewInbox_LongPrefix  | .NET 8.0      | .NET 8.0      | 78.37 ns | 1.593 ns | 1.705 ns |  0.91 |    0.02 | 0.0994 |     208 B |        1.00 |
| NewInbox_LongPrefix  | NativeAOT 8.0 | NativeAOT 8.0 | 72.60 ns | 1.492 ns | 1.532 ns |  0.85 |    0.01 | 0.0994 |     208 B |        1.00 |
```

src/NATS.Client.Core/NatsConnection.RequestReply.cs
mtmk: AFAIK, practice is to have a prefix so the accounts can be permissined accordingly.

mtmk: fwiw, Actually just realized an empty prefix might create interesting behaviours in subscription manager i.e. `SubscriptionManager.IsInboxSubject()` would always return `true` causing all subscriptions to go through an internal muxed 'inbox' subscription.

mtmk:
Thanks @jasper-d looks awesome! Sorry for the delay. Started reviewing today and trying to get my head around it.

jasper-d:
> Sorry for the delay.

No worries, this isn't exactly urgent.

src/NATS.Client.Core/Internal/NuidWriter.cs
```diff
@@ -0,0 +1,135 @@
...
+        if (rng == null)
+        {
+            RandomNumberGenerator.Fill(randomBytes);
+        }
+        else
+        {
+            rng.GetBytes(randomBytes);
+        }
+
+        var newPrefix = new char[PrefixLength];
```
mtmk, line 89: Just curious. Is there a technical reason why _prefix shouldn't be preallocated since it'd be on TLS? I assume it'd be over optimization since this isn't on hot path. Only a question really.

mtmk (approved):
LGTM

Nicely follows [the Go implementation](https://github.com/nats-io/nuid/blob/master/nuid.go).

Thanks @jasper-d. I have a question below out of curiosity, but please feel free to ignore it if you don't have time. It's not a merge-stopper.

Also a quick note on using subject as string: performance benefit seems negligible to me (e.g. about 1-2% overall allocations) at some point we should revisit though.

src/NATS.Client.Core/Internal/NuidWriter.cs
jasper-d: No, there is no reason. I implemented NUID a long time ago using locks and never changed this.

PR #149, opened 2023-10-12
KV docs and release prep 2.0.0-alpha.4

mtmk:
* Initial Key Value Store implementation
* Implement NUID (jasper-d)
* Support all binary payload APIs with only generics
* Handle publish serialization exceptions
* Add JetStream metadata to msg parsed from reply string (simonhoss)
* Added subject transforms to stream source config (server 2.10 support)


caleblloyd (approved):
LGTM

PR #156, opened 2023-10-16
TLS first connection

mtmk:
Currently only supported in server PR https://github.com/nats-io/nats-server/pull/4642
- [x] Automated test once PR merged

### DIFF
--- a/sandbox/Example.TlsFirst/Example.TlsFirst.csproj
+++ b/sandbox/Example.TlsFirst/Example.TlsFirst.csproj
@@ -5,6 +5,7 @@
         <TargetFramework>net6.0</TargetFramework>
         <ImplicitUsings>enable</ImplicitUsings>
         <Nullable>enable</Nullable>
+        <IsPackable>false</IsPackable>
     </PropertyGroup>
 
     <ItemGroup>
--- a/sandbox/Example.TlsFirst/Example.TlsFirst.csproj
+++ b/sandbox/Example.TlsFirst/Example.TlsFirst.csproj
@@ -1,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net6.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\..\src\NATS.Client.Core\NATS.Client.Core.csproj" />
+    </ItemGroup>
+
+</Project>
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -265,7 +265,7 @@ public partial class NatsConnection : IAsyncDisposable, INatsConnection
                     {
                         // upgrade TcpConnection to SslConnection
                         var sslConnection = conn.UpgradeToSslStreamConnection(Opts.TlsOpts, _tlsCerts);
-                        await sslConnection.AuthenticateAsClientAsync(FixTlsHost(uri)).ConfigureAwait(false);
+                        await sslConnection.AuthenticateAsClientAsync(uri).ConfigureAwait(false);
                         _socket = sslConnection;
                     }
                 }
@@ -536,13 +536,18 @@ public partial class NatsConnection : IAsyncDisposable, INatsConnection
 
     private NatsUri FixTlsHost(NatsUri uri)
     {
+        var lastSeedHost = _lastSeedConnectUri?.Host;
+
+        if (string.IsNullOrEmpty(lastSeedHost))
+            return uri;
+
         // if the current URI is not a seed URI and is not a DNS hostname, check the server cert against the
         // last seed hostname if it was a DNS hostname
         if (!uri.IsSeed
             && Uri.CheckHostName(uri.Host) != UriHostNameType.Dns
-            && Uri.CheckHostName(_lastSeedConnectUri!.Host) == UriHostNameType.Dns)
+            && Uri.CheckHostName(lastSeedHost) == UriHostNameType.Dns)
         {
-            return uri.CloneWith(_lastSeedConnectUri.Host);
+            return uri.CloneWith(lastSeedHost);
         }
 
         return uri;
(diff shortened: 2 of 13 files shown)


sandbox/Example.TlsFirst/Example.TlsFirst.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net6.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
```
caleblloyd, line 7:
```suggestion
        <Nullable>enable</Nullable>
        <IsPackable>false</IsPackable>
```

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -23,5 +23,8 @@ public sealed record NatsTlsOpts
     /// <summary>When true, skip remote certificate verification and accept any server certificate</summary>
     public bool InsecureSkipVerify { get; init; }
 
+    /// <summary>When true, client initiates TLS connection immediately when connecting</summary>
+    public bool TlsFirst { get; init; }
```
caleblloyd, line 27: I think the technical term for this is "Implicit TLS", if we are looking to standardize around an industry-specific term.

In the future, if NATS introduces an Implict TLS port (like 4223 or whatever) this may define the behavior.  So we might want something like this:

```csharp
enum TlsMode {
  Auto,
  Disabled,
  Explicit,
  Implicit
}
```

That would allow us to drop the `Disabled` boolean and roll it into the enum as well

sandbox/Example.TlsFirst/Example.TlsFirst.csproj
mtmk: Doh! thank you 💯

src/NATS.Client.Core/NatsTlsOpts.cs
mtmk: Sounds like a sensible thing to me.

We should probably update the spec as well @Jarema?

caleblloyd: Added a comment to the spec - https://github.com/nats-io/nats-architecture-and-design/pull/247#discussion_r1360759283

caleblloyd: We may even want to expand out `Explicit` even further- within `Explicit` there is:

- `Prefer` - default when the `nats://` scheme is used - if the Server supports TLS, then use it, otherwise use plain-text
- `Require` - default when the `tls://` scheme is used- if the Server does not support TLS, then fail

mtmk: What's the difference between `prefered` and `auto`?

caleblloyd: `Auto` means use some criteria to decide.  The client logic as of today for `Auto` would be to use `Prefer` if a `nats://` scheme was used and `Require` if the `tls://` scheme was used.

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -222,9 +222,11 @@ private async ValueTask InitialConnectAsync()
         Debug.Assert(ConnectionState == NatsConnectionState.Connecting, "Connection state");
 
         var uris = Opts.GetSeedUris();
-        if (Opts.TlsOpts.Disabled && uris.Any(u => u.IsTls))
-            throw new NatsException($"URI {uris.First(u => u.IsTls)} requires TLS but NatsTlsOpts.Disabled is set to true");
-        if (Opts.TlsOpts.Required)
+        if (Opts.TlsOpts.EffectiveMode == TlsMode.Require && uris.Any(u => !u.IsTls))
+            throw new NatsException($"URI {uris.First(u => !u.IsTls)} doesn't support TLS but TlsMode is set to Explicit");
```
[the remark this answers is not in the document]
mtmk: Removed

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -1,5 +1,36 @@
...
+
+    /// <summary>
+    /// Upgrades the connection to TLS as soon as the connection is established.
+    /// </summary>
+    Implicit,
+
+    /// <summary>
+    /// Disabled mode will not attempt to upgrade the connection to TLS.
+    /// </summary>
+    Disabled,
```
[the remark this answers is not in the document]
mtmk: fixed

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -17,14 +48,17 @@ public sealed record NatsTlsOpts
...
-    public bool TlsFirst { get; init; }
+    /// <summary>TLS mode to use during connection</summary>
+    public TlsMode Mode { get; init; }
+
+    internal bool HasTlsFile => CertFile != default || KeyFile != default || CaFile != default;
 
-    internal bool Required => CertFile != default || KeyFile != default || CaFile != default;
+    internal TlsMode EffectiveMode => Mode switch
+    {
+        TlsMode.Auto => TlsMode.Prefer,
```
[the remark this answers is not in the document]
mtmk: fixed

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -265,7 +265,7 @@ private async ValueTask InitialConnectAsync()
                     {
                         // upgrade TcpConnection to SslConnection
                         var sslConnection = conn.UpgradeToSslStreamConnection(Opts.TlsOpts, _tlsCerts);
-                        await sslConnection.AuthenticateAsClientAsync(uri).ConfigureAwait(false);
+                        await sslConnection.AuthenticateAsClientAsync(FixTlsHost(uri)).ConfigureAwait(false);
```
caleblloyd, line 268: `FixTlsHost` uses `_lastSeedConnectUri!.Host` which is not set until `SetupReaderWriterAsync` is called on line 298.  Fixing the host up is only needed for reconnects, where IP Addresses have been gossiped but the original hostname of the Seed Server needs to be tracked

mtmk: Good catch! Fixed.

caleblloyd (approved):
LGTM

PR #161, opened 2023-10-19
Consume clean exit fixes

mtmk:
* Enlarge subscription channel to avoid blocking other socket reads on the same connection.

* Pass cancellation token to Consume() so it can be gracefully stopped.

* 'Consume' can also be gracefully stopped when Stop() is called.

* There is no graceful stop for 'Fetch' since we have no option but wait for the single pull request to complete.

* Disposing the 'Consume' or 'Fetch' would exit loops a.s.a.p.


sandbox/Example.JetStream.PullConsumer/Program.cs
```diff
@@ -50,32 +50,42 @@ void Report(int i, Stopwatch sw, string data)
...
+    if (cmd == "fetch")
     {
         while (!cts.Token.IsCancellationRequested)
         {
             try
             {
                 Console.WriteLine($"___\nFETCH {maxMsgs}");
                 await consumer.RefreshAsync(cts.Token);
-                await using var sub = await consumer.FetchAsync<RawData>(fetchOpts, cts.Token);
+                await using var sub = await consumer.FetchAsync<IMemoryOwner<byte>>(fetchOpts, cts.Token);
```
caleblloyd, line 77: If you are a fan of #162 you may want to convert these examples!

caleblloyd (approved):
LGTM

PR #162, opened 2023-10-20
add NatsMemoryOwner

caleblloyd:
Adapts `MemoryOwner<T>` from `CommunityToolkit.HighPerformance` ([docs](https://learn.microsoft.com/en-us/dotnet/communitytoolkit/high-performance/memoryowner), [source](https://github.com/CommunityToolkit/dotnet/blob/v8.2.1/src/CommunityToolkit.HighPerformance/Buffers/MemoryOwner%7BT%7D.cs))

Changes it to a struct for non-nullable de-serialization - this was the main motivation


mtmk (approved):
LGTM

PR #165, opened 2023-10-23
Initial Service API implementation

mtmk:
Implementation of [Service API ADR-32](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-32.md)


src/NATS.Client.Services/NatsSvcService.cs
```diff
@@ -0,0 +1,329 @@
...
+using NATS.Client.Core.Internal;
+using NATS.Client.Services.Internal;
+using NATS.Client.Services.Models;
+
+namespace NATS.Client.Services;
+
+/// <summary>
+/// NATS service.
+/// </summary>
+public class NatsSvcService : IAsyncDisposable
```
caleblloyd, line 15: Maybe call it `NatsSvcServer` ?  Seems to be serving a service, and `SvcService` is a little confusing

src/NATS.Client.Services/NatsSvcService.cs
```diff
@@ -0,0 +1,329 @@
...
+    /// <param name="name">Optional endpoint name.</param>
+    /// <param name="subject">Optional endpoint subject.</param>
+    /// <param name="metadata">Optional endpoint metadata.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to stop the endpoint.</param>
+    /// <typeparam name="T">Serialization type for messages received.</typeparam>
+    /// <returns>A <seealso cref="ValueTask"/> representing the asynchronous operation.</returns>
+    /// <remarks>
+    /// One of name or subject must be specified.
+    /// </remarks>
+    public ValueTask AddEndPointAsync<T>(Func<NatsSvcMsg<T>, ValueTask> handler, string? name = default, string? subject = default, IDictionary<string, string>? metadata = default, CancellationToken cancellationToken = default) =>
```
caleblloyd, line 88: Nit - I am a fan of "endpoint" as one word so the capitalization should probably be `Endpoint` wherever this appears

Yes... `System.Net.EndPoint` does "end point"...  But `Microsoft.AspNetCore.Http.Endpoint` does "endpoint" :smile:

mtmk: Hehe yes I thought about that, then checked IPEndPoint 😆 I agree. 'endpoint' is a word so..

caleblloyd (approved):
LGTM.  Looks like tests failing on latest commits

mtmk:
> LGTM. Looks like tests failing on latest commits

there was some issue pulling nats-server. resolved now.

PR #166, opened 2023-10-23
Add StreamInfo request to GetStream

simonhoss:
This PR adds the StreamInfoRequest to GetStream

Note: This option is not present in the nats go client. To get the required StreamInfo in the nats go client they have the `JetStreamManagement`. I added it to this method, because nats.net.v2 doesn't have the `JetStreamManagement` at the moment.

We need it to figure out the subjects inside a stream. If there is a better way, I'm happy to hear about this.


mtmk (approved):
LGTM

Thank you @simonhoss 🥇

PR #167, opened 2023-10-24
Inbox subscription options fix

mtmk:
Internal inbox subscription options was being used to create the real mux-inbox subscription (on first subscription attempt), which I noticed when trying to use timeouts with request-reply calls. Timeout was ending the real mux-inbox which was causing the rest of the request-reply calls to silently fail.


caleblloyd (approved):
LGTM

PR #170, opened 2023-10-26
Mark StreamState Subjects as nullable

simonhoss:
This PR marks the Subjects property as nullable. Subjects gets only set when StreamInfo has a subjects filter set and nats is initialized.


mtmk:
None of the other IDictionary fields in the models is nullable. Is this actually causing an error or is it about applications accessing this field e.g. `if (state.Subjects == null)` as opposed to `if (state.Subjects.Count == 0)`?

simonhoss:
`if (state.Subjects.Count == 0)` would not work because state.Subjects is in some circumstances empty.

mtmk (approved):
LGTM

I see it now. If the property isn't defined in JSON string, it's serialized as `null` even though the C# property is not nullable.

Thank you @simonhoss 💯

PR #176, opened 2023-11-01
Key/Value Store fixes

mtmk:
Key/Value Store fixes and missing features.

* Enlarged channel sizes to avoid blocking socket
* Implemented missing features: GetKeys, Update, Create, Purge, History, etc.
* Simplified the watcher to async enumeration only (await foreach)


caleblloyd (approved):
LGTM

PR #182, opened 2023-11-07
Subscribe API async enumerable change

mtmk:
This changes the core subscription to be in parity with JetStream consume APIs using async enumerable:

```csharp
await foreach (var msg in _nats.SubscribeAsync<int>("data", cancellationToken: cancellationToken))
{
    Console.WriteLine($"{msg.Subject}: {msg.Data}");
    await msg.ReplyAsync(42, cancellationToken: cancellationToken);
}
```

The idea is to keep the public interface as minimal as possible while being consistent across the board.


caleblloyd (approved):
LGTM

PR #185, opened 2023-11-07
JetStream consume idle heartbeat fix

mtmk:
Keep the idle heartbeat timer ticking so if all else fails, it will try to send pull requests as long as we have an open connection. We also made a preventive change to connection state management making it thread safe since we suspect there may be race conditions on ARM processors for example.

(Issue was found during failground testing: consumer stopped pulling after about 20 hours)


caleblloyd (approved):
LGTM

Issue #105, opened 2023-07-31
Facilitate unit testing of client code of nats.net.v2


jasper-d:
## Feature Request
Make testing easier by using interfaces instead of sealed classes.

#### Use Case:
Writing tests for code that uses nats.net.v2 is challenging, because the types nats.net.v2 exposes are sealed classes which cannot be replaced with test doubles.

#### Proposed Change:
Define interfaces for at least `NatsConnectionPool` and `NatsConnection` to facilitate the use of test doubles so that users of nats.net.v2 can easily replace those and test their code without requiring nats-server.

```diff
+ public interface INatsConnectionPool : IAsyncDisposable {
+  INatsConnection GetConnection();
+  IEnumerable<INatsConnection> GetConnections();
+  INatsConnection GetCommand(); // I'm wondering why this is named GetCommand but that's a separate issue
+}

- public sealed class NatsConnectionPool : IAsyncDisposable
+ public sealed class NatsConnectionPool : INatsConnectionPool

// Similar changes for NatsConnection
```

#### Who Benefits From The Change(s)?
Every user of nats.net.v2 who wants to test their code.

#### Alternative Approaches
Have nats-server configured and running when executing tests.

mtmk:
I don't see why we shouldn't create an `INatsConnectionPool` interface.

We already have an `INatsConnection` interface, but extension methods (I think just request-reply methods) works with `NatsConnection` due  to internal calls. We'd need to revisit that.

As for running nats-server for tests, we already have a test utility for that. Do you think it would be handy to publish the test utility on NuGet?

mtmk:
> > I don't see why we shouldn't create an INatsConnectionPool interface.
> 
> Great, I can spend some time on it if you don't mind and hopefully come up with a proper proposal.

That'd be brilliant! Could be a quick PR if you're happy with that.

> > We already have an INatsConnection interface, but extension methods (I think just request-reply methods) works with NatsConnection due to internal calls. We'd need to revisit that.
> 
> I found `NatsRequestExtensions` and `NatsRequestManyExtensions`. Technically those methods could be moved to `INatsConnection` and become ordinary instance methods I suppose. But there is probably a reason for them being extension methods.
> 
> Alternatively, the few internal calls used by the extension methods could possibly be made public and become part of the interface. Mocking would then work, although it can be a bit cumbersome to setup.

The idea is to keep public interfaces as succinct as possible. @caleblloyd's input would be invaluable here.

[...]

PR #194, opened 2023-11-08
Key Value store interfaces

mtmk:
Defined Key Value Store public interfaces to indicate what is supported and help with dependency injection and unit testing scenarios.

contributes to #105


caleblloyd (approved):
LGTM

PR #195, opened 2023-11-08
Object store interfaces and missing features

mtmk:
Defined Object Store public interfaces to indicate what is supported and help with dependency injection and unit testing scenarios.

Also implemented missing features:
* Add bucket link
* Seal
* List
* Get status


caleblloyd (approved):
LGTM

PR #196, opened 2023-11-08
Services interfaces

mtmk:
Defined Services public interfaces to indicate what is supported and help with dependency injection and unit testing scenarios.

contributes to #105


caleblloyd (approved):
LGTM

PR #197, opened 2023-11-09
TLS validation error logging on connection

mtmk:
When connection fails because of SSL validation checks the exception doesn't tell you that (just says could not connect). If the user enables logging then they'd at least be able to see if the connection failed because of TLS validation errors and why.


caleblloyd (approved):
LGTM

PR #198, opened 2023-11-09
Ordered consumer sequence and clean-up fix

mtmk:
* Fixed sequence number bug when recreating consumers.
* Also made sure the ephemeral consumer was deleted on errors.
* Also added a memory test to make sure consumer subscription isn't being GCed prematurely.


caleblloyd (approved):
LGTM

PR #199, opened 2023-11-09
Test JS cluster and reconnect utils

mtmk:
* Run the server strictly on single IP (127.0.0.1). This gives us predictable client URLs from the server, as opposed to all the IP of external interfaces, which might vary.
* Had to regenerate certs to have 127.0.0.1 in SAN fro the server.
* Made the cluster setup JetStream capable.
* Server can be restarted.


caleblloyd (approved):
LGTM

PR #204, opened 2023-11-13
Client TLS tests

mtmk:
Tests to make sure TLS client certificates works.

* Also allow TLS 1.3
* Tweak to how client certificates taken in for Windows
* Format fixes

### DIFF
--- a/src/NATS.Client.Core/Internal/TlsCerts.cs
+++ b/src/NATS.Client.Core/Internal/TlsCerts.cs
@@ -1,3 +1,4 @@
+using System.Runtime.InteropServices;
 using System.Security.Cryptography.X509Certificates;
 
 namespace NATS.Client.Core.Internal;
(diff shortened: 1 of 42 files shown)


mtmk:
Checking why tests are hanging.

src/NATS.Client.Core/Internal/TlsCerts.cs
```diff
@@ -25,7 +26,19 @@ public TlsCerts(NatsTlsOpts tlsOpts)
 
         if (tlsOpts.CertFile != default && tlsOpts.KeyFile != default)
         {
-            ClientCerts = new X509Certificate2Collection(X509Certificate2.CreateFromPemFile(tlsOpts.CertFile, tlsOpts.KeyFile));
+            var clientCert = X509Certificate2.CreateFromPemFile(tlsOpts.CertFile, tlsOpts.KeyFile);
+
+            // On Windows, ephemeral keys/certificates do not work with schannel. e.g. unless stored in certificate store.
+            // https://github.com/dotnet/runtime/issues/66283#issuecomment-1061014225
+            // https://github.com/dotnet/runtime/blob/380a4723ea98067c28d54f30e1a652483a6a257a/src/libraries/System.Net.Security/tests/FunctionalTests/TestHelper.cs#L192-L197
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
caleblloyd, line 34: Is there a better way to detect SChannel than OS Platform?  I feel like there are probably .NET derivatives out there running on Windows linked against OpenSSL.  Although maybe it doesn't matter because this probably produces a valid result with OpenSSL?

mtmk: That is a good point. I could not find a way to check if underlying implementation is using SChannel or OpenSSL, but it must be possible. In the meantime I tried it without the platform check on Linux and it worked fine.

caleblloyd (approved):
LGTM

PR #205, opened 2023-11-13
Subscribe core method

mtmk:
Expose slightly lower level API for subscriptions to make it easier to build custom messaging patterns.


caleblloyd (approved):
Welcome back, INatsSub! LGTM

PR #210, opened 2023-11-14
JetStream double ACK and NAK delay

mtmk:
* Messages received by the consumer can be acknowledged optionally asking the server acknowledging the message acknowledgment.
* Messages can be 'rejected' (-NAK) with a delay.


caleblloyd (approved):
LGTM

PR #211, opened 2023-11-15
Reconnect backoff

mtmk:
To help minimize the impact on the servers during failures.

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -35,6 +35,8 @@ public partial class NatsConnection : IAsyncDisposable, INatsConnection
     private readonly CancellationTokenSource _disposedCancellationTokenSource;
     private readonly string _name;
     private readonly TimeSpan _socketComponentDisposeTimeout = TimeSpan.FromSeconds(5);
+    private readonly int _maxReconnectRetry;
+    private readonly int[] _backOffPolicy;
 
     private int _pongCount;
     private bool _isDisposed;
@@ -51,6 +53,7 @@ public partial class NatsConnection : IAsyncDisposable, INatsConnection
     private TlsCerts? _tlsCerts;
     private ClientOpts _clientOpts;
     private UserCredentials? _userCredentials;
+    private int _connectRetry;
 
     public NatsConnection()
         : this(NatsOpts.Default)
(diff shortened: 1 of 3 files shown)


src/NATS.Client.Core/NatsOpts.cs
```diff
@@ -8,88 +8,85 @@ namespace NATS.Client.Core;
...
+
+    /// <summary>
+    /// Backoff policy for reconnect attempts. (default: 0, 250, 250, 500, 500, 3000, 5000)
+    /// </summary>
+    /// <remarks>
+    /// This is a list of milliseconds to wait between reconnect attempts where the next element is used with every attempt.
+    /// When the number of retry attempts reaches the end of the list, it will keep using the last element
+    /// which is 5 seconds by default.
+    /// </remarks>
+    public int[] ReconnectBackoffPolicyMilliseconds { get; init; } = { 0, 250, 250, 500, 500, 3000, 5000 };
```
caleblloyd, line 89: Could we use a simpler exponential backoff mechanism, such as:

`TimeSpan? ReconnectWaitMax`

When `ReconnectWaitMax` is set, `ReconnectWait` will be doubled on each reconnect attempt until it reaches `ReconnectWaitMax`, at which point it will stay at ReconnectWaitMax until MaxReconnectRetry has been reached

caleblloyd: Or if you want to default it could just be `TimeSpan ReconnectWaitMax = TimeSpan.FromSeconds(5)` and not even make it nullable.  If it is <= ReconnectWait just ignore it

mtmk: I like that. Should we also consider a 'Policy' exponential being the default? (really a question, it might make it unnecessarily complex)

caleblloyd: I thought about that, but at that point it might be easier to supply a callback

caleblloyd (approved):
LGTM

caleblloyd:
Only other comments is that with `ReconnectWaitMax` set by default, it might make more sense to rename `ReconnectWait` to `ReconnectWaitMin`

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -561,8 +566,40 @@ private NatsUri FixTlsHost(NatsUri uri)
...
+            else if (_backoff == TimeSpan.Zero)
+            {
+                _backoff = Opts.ReconnectWait;
+            }
+            else if (_backoff == Opts.ReconnectWaitMax)
+            {
+            }
+            else
+            {
+                _backoff *= 2;
```
caleblloyd, line 588: Should probably make this something like:

```
                Math.Max(_backoff *= 2, TimeSpan.FromSeconds(1));
```

Because if _backoff is 0 or negative it'll never grow towards ReconnectWaitMax

caleblloyd: Or maybe just check if _backoff <=0 and if so set it to 1 second... The Math.Max might be too aggressive for smaller values

mtmk: done

PR #214, opened 2023-11-15
Key/Value and Object Store compression

mtmk:
contributes to nats-io/nats-architecture-and-design#253

### DIFF
--- a/src/NATS.Client.KeyValueStore/NatsKVConfig.cs
+++ b/src/NATS.Client.KeyValueStore/NatsKVConfig.cs
@@ -56,6 +56,11 @@ public record NatsKVConfig
     /// </summary>
     public NatsKVRepublish? Republish { get; init; }
 
+    /// <summary>
+    /// Use compressed storage.
+    /// </summary>
+    public bool IsCompressed { get; init; }
+
     // TODO: Bucket mirror configuration.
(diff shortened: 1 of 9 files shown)


caleblloyd:
Looks like Deno uses `Compression` with the actual Compression options in the config.  Which makes more sense if there will ultimately be multiple Compression options to choose from

caleblloyd:
https://github.com/nats-io/nats.deno/pull/625/files

mtmk:
> Looks like Deno uses `Compression` with the actual Compression options in the config. Which makes more sense if there will ultimately be multiple Compression options to choose from

I think there was a discussion about it and ADR says it should be bool also Go and Java using bool. Not sure why deno is different also rust looks the same as well. huh?!

mtmk:
Turns out it should be bool.

mtmk:
> What do the other Clients call the Config param? Or is there a matching Stream Config param name that we should align?

Go is using 'Compression' and 'IsCompressed'. [ADR is the same](https://github.com/nats-io/nats-architecture-and-design/pull/252/files#diff-c195b0afc458299fabf26f0d2fb05a9b6bc0db0475dad8885d020b199046c355R288).

caleblloyd (approved):
LGTM

PR #220, opened 2023-11-17
JetStream consume error notifications

mtmk:
Notify users of events like timeouts while consuming messages.

/cc @Jarema

### DIFF
--- a/src/NATS.Client.JetStream/INatsJSNotification.cs
+++ b/src/NATS.Client.JetStream/INatsJSNotification.cs
@@ -1,0 +1,11 @@
+﻿namespace NATS.Client.JetStream;
+
+public interface INatsJSNotification
+{
+    string Name { get; }
+}
+
+public class NatsJSTimeoutNotification : INatsJSNotification
+{
+    public string Name => "Timeout";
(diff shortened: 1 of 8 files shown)


caleblloyd:
Since the API is all `IAsyncEnumerable` driven right now, something doesn't feel quite right about using callbacks for notifications.  It may ultimately be the right solution but I am trying to think about a practical example.

For example, what if I am running a `Consume` loop.  Now I receive a notification that I want to respond to, and break out of my loop.  It would be useful if the Notification callback could return a decision, such as whether to continue iterating or not.

Also should the handlers have cancellation support?  They are async handlers.

mtmk:
> For example, what if I am running a `Consume` loop. Now I receive a notification that I want to respond to, and break out of my loop. It would be useful if the Notification callback could return a decision, such as whether to continue iterating or not.

Good point. An exception can be thrown to signal failure which would be propagated to the async iteration to be caught by the user code.

> Also should the handlers have cancellation support? They are async handlers.

Yes! Added.

caleblloyd (approved):
LGTM

PR #223, opened 2023-11-20
Project targets updated to .NET 8.0

mtmk:
* Updated targets to net6 + net8
* Fixed AOT warnings for JSON enums
* Fixed CancellationTokenSource.CancelAsync warnings for .net8
* Removed unused failground project
* Test should run both for ne6 & 8 now


src/NATS.Client.Core/NATS.Client.Core.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
```
caleblloyd, line 4: Is there any point in targeting both right now?  Since we want to support the .NET LTS Release, and `net6.0` is supported until November 12, 2024, should we just stay on `net6.0` now.  Then on November 12, 2024 move the target to `net8.0`.

If there was a compelling enough use case for a `net8.0`-only API we could target both at that time, and use preprocessor directives to support both `net6.0` and `net8.0` until then.

mtmk: See the changes in JetStream models.

caleblloyd: Ah.  But shouldn't only that one assembly need the `net.8.0` target instead of all of them?

mtmk: I think it's good to have it on across the board since it would give us warnings if there is an obsolete apis or suggestions. There is one about the cancellation token source new `CancelAsync()` method for example.

mtmk: Another example is it makes it easier to spot trimming warnings.

caleblloyd: OK, let's document a policy to tag every assembly with the supported .NET LTS Releases then, and move forward with this

caleblloyd (approved):
LGTM

PR #246, opened 2023-11-27
Increase consume and fetch channel size

mtmk:
Increase consume and fetch channel size to avoid blocking the socket reads for other operations. I think we have done this change before but somehow reverted back during merges perhaps.


caleblloyd (approved):
LGTM

PR #248, opened 2023-11-27
Removed redundant StreamUpdateRequest class

mtmk:
StreamUpdateRequest was a placeholder class inheriting StreamConfig and adding no value to it. Using StreamUpdateRequest in update stream methods and StreamConfig in creation method was inconsistent and it was causing unnecessary object creation in some cases forcing the end developer to map everything from config object.


caleblloyd (approved):
LGTM

PR #250, opened 2023-11-27
Use TimeSpan for duration instead of nanoseconds

mtmk:


caleblloyd (approved):
Nice, much easier to deal with.  LGTM

PR #251, opened 2023-11-27
Stream CreateConsumer renamed CreateOrUpdateConsumer

mtmk:


caleblloyd:
Do Streams exhibit the same behavior?

mtmk:
> Do Streams exhibit the same behavior?

Streams has a create() and an update() method.

Just tested create() make sure: Using create() is idempotent. It won't update if configuration is different with an error `stream name already in use with a different configuration` but it doesn't complain if it has the exact same configuration.

caleblloyd (approved):
LGTM

caleblloyd:
> Just tested create() make sure: Using create() is idempotent. It won't update if configuration is different with an error `stream name already in use with a different configuration` but it doesn't complain if it has the exact same configuration.

I wonder if we should add a stream CreateOrUpdateAsync that does 2 calls - a Get, and conditionally a Create or Update after that.  Wonder if that is standard in other clients or not?

PR #255, opened 2023-11-28
Fixed JetStream ACK custom serialization

mtmk:
When a custom serializer used JetStream msg ACKs are still being serialized using the custom serializer which is resulting in +ACK being sent to the server with a distorted format. Now we just used the NATS default serializer on message ACKs to format the +ACK/NAK/etc. payload correctly.

ConsumerFetchTest.cs was also passing for the wrong reasons, which is also fixed. Plus minor format issue fixed.


caleblloyd (approved):
LGTM

mtmk:
usual test flappers left.

PR #256, opened 2023-11-28
Releasing 2.0.1

mtmk:
Blocking bugs and breaking changes:
* JS ack won't work with custom serializer
* Fetch required maxMsgs parameter
* JS config durations are now TimeStamps
* JS CreateConsumer() renamed to CreateOrUpdateConsumer()


caleblloyd (approved):
LGTM

Issue #258, opened 2023-11-28
Add support for the no_responders feature


stebet:
### Observed behaviour

The old client enabled the [no_responders](https://docs.nats.io/nats-concepts/core-nats/reqreply#no-responders) client feature by default, but this one seems to have it disabled and no way to enable it currently. 

This feature has been immensely helpful for request-reply patterns so I'd love to see it added back in. I can take a look at getting an initial PR started for it :)

### Expected behaviour

Clients should receive a NatsNoRespondersException (503 status code) when sending a subject message that has a reply_to header but no subscriber is registered to receive that subject. 

### Server and client version

.NET client v2.0.0
Latest NATS server

### Host environment

Latest Linux ARM docker image

### Steps to reproduce

Publish a message/request with a reply_to header to a subject that has no subscribers. The reply_to address should receive an empty message with an error code 503 in the header indicating that no eligible subscriber for that subject is registered.

stebet:
Guess this can be closed now @mtmk? :)

PR #259, opened 2023-11-28
Adding support for the no_responders feature.

stebet:
This solves #258

* Feature can be enabled via. `NatsOpts.NoResponders` (defaults to `false` like before the PR)
* Throws a new `NatsNoRespondersException` if triggered cia. `RequestAsync`, but not if triggered via `PublishAsync`.
* A normal subscriber that receives a message due to `no_responders`, in case the client is using his own `request-reply` mechanism through PubSub and `reply_to`, does NOT throw an exception but the subscriber receives an empty message with `msg.Headers.Code` equal to `503`.
  * We could also have this throw an exception but I felt it didn't make sense outside of `RequestAsync` but that's open for discussion. It could even be another configuration value for NatsOpts (`bool ThrowOnNoResponders`) so let's discuss that.
* Added tests for both `RequestAsync` and custom `PubSub` to trigger the behavior.
* Also ran `dotnet format` which formatted some unrelated files.

### DIFF
--- a/src/NATS.Client.Core/NatsMsg.cs
+++ b/src/NATS.Client.Core/NatsMsg.cs
@@ -60,9 +60,9 @@ public readonly record struct NatsMsg<T>(
         }
 
         var size = subject.Length
-            + (replyTo?.Length ?? 0)
-            + (headersBuffer?.Length ?? 0)
-            + payloadBuffer.Length;
+                   + (replyTo?.Length ?? 0)
+                   + (headersBuffer?.Length ?? 0)
+                   + payloadBuffer.Length;
 
         return new NatsMsg<T>(subject, replyTo, (int)size, headers, data, connection);
     }
@@ -141,4 +141,6 @@ public readonly record struct NatsMsg<T>(
             throw new NatsException("unable to send reply; ReplyTo is empty");
         }
     }
+
+    public bool IsNoRespondersError => Headers?.Code == 503;
 }
--- a/src/NATS.Client.Core/NatsMsg.cs
+++ b/src/NATS.Client.Core/NatsMsg.cs
@@ -30,6 +30,8 @@ public readonly record struct NatsMsg<T>(
     T? Data,
     INatsConnection? Connection)
 {
+    public bool IsNoRespondersError => Headers?.Code == 503;
+
     internal static NatsMsg<T> Build(
         string subject,
         string? replyTo,
@@ -141,6 +143,4 @@ public readonly record struct NatsMsg<T>(
             throw new NatsException("unable to send reply; ReplyTo is empty");
         }
     }
-
-    public bool IsNoRespondersError => Headers?.Code == 503;
 }
--- a/src/NATS.Client.Core/Internal/ClientOpts.cs
+++ b/src/NATS.Client.Core/Internal/ClientOpts.cs
@@ -21,7 +21,6 @@ internal sealed class ClientOpts
         Password = opts.AuthOpts.Password;
         AuthToken = opts.AuthOpts.Token;
         JWT = opts.AuthOpts.Jwt;
-        NoResponders = opts.NoResponders;
     }
 
     /// <summary>Optional boolean. If set to true, the server (version 1.2.0+) will not send originating messages from this connection to its own subscriptions. Clients should set this to true only for server supporting this feature, which is when proto in the INFO protocol is set to at least 1.</summary>
@@ -89,7 +88,7 @@ internal sealed class ClientOpts
     public bool Headers { get; init; } = false;
 
     [JsonPropertyName("no_responders")]
-    public bool NoResponders { get; init; } = false;
+    public bool NoResponders { get; init; } = true;
 
     public static ClientOpts Create(NatsOpts opts)
     {
--- a/src/NATS.Client.Core/Internal/ClientOpts.cs
+++ b/src/NATS.Client.Core/Internal/ClientOpts.cs
@@ -21,6 +21,7 @@ internal sealed class ClientOpts
         Password = opts.AuthOpts.Password;
         AuthToken = opts.AuthOpts.Token;
         JWT = opts.AuthOpts.Jwt;
+        NoResponders = opts.NoResponders;
     }
 
     /// <summary>Optional boolean. If set to true, the server (version 1.2.0+) will not send originating messages from this connection to its own subscriptions. Clients should set this to true only for server supporting this feature, which is when proto in the INFO protocol is set to at least 1.</summary>
--- a/src/NATS.Client.Core/NatsConnection.RequestReply.cs
+++ b/src/NATS.Client.Core/NatsConnection.RequestReply.cs
@@ -32,7 +32,7 @@ public partial class NatsConnection
         {
             if (sub.Msgs.TryRead(out var msg))
             {
-                if (msg.Headers?.Code == 503)
+                if (msg.IsNoRespondersError)
                 {
                     throw new NatsNoRespondersException();
                 }
--- a/src/NATS.Client.Core/NatsConnection.RequestReply.cs
+++ b/src/NATS.Client.Core/NatsConnection.RequestReply.cs
@@ -32,6 +32,11 @@ public partial class NatsConnection
         {
             if (sub.Msgs.TryRead(out var msg))
             {
+                if (msg.Headers?.Code == 503)
+                {
+                    throw new NatsNoRespondersException();
+                }
+
                 return msg;
             }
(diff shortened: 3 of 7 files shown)


mtmk:
> A normal subscriber that receives a message due to `no_responders`, in case the client is using his own `request-reply` mechanism through PubSub and `reply_to`, does NOT throw an exception but the subscriber receives an empty message with `msg.Headers.Code` equal to `503`.
> * We could also have this throw an exception but I felt it didn't make sense outside of `RequestAsync` but that's open for discussion. It could even be another configuration value for NatsOpts (`bool ThrowOnNoResponders`) so let's discuss that.

Maybe we should make it a property on the message itself? ([Rust client seems to be doing that](https://github.com/nats-io/nats.rs/blob/aebbe14e65324150fd28c19fd256e08915020057/nats/src/message.rs#L104))

Also looking at [Go](https://github.com/nats-io/nats.go/blob/8712190da1d17ab0c4719bffa7c0174214c56e6c/nats.go#L2502) and [Rust](https://github.com/nats-io/nats.rs/blob/25c47a6faf5f82737f60b49757ce96b831890649/async-nats/src/connector.rs#L175) clients there isn't an option to set `no_responders`. If the client is supporting headers `no_responders` is implied.

stebet:
> Maybe we should make it a property on the message itself? ([Rust client seems to be doing that](https://github.com/nats-io/nats.rs/blob/aebbe14e65324150fd28c19fd256e08915020057/nats/src/message.rs#L104))

That's possible. I still think an exception is more intuitive though from a C# convention perspective (compared to the NatsNoReplyException for timeouts), unless explicitly documented, since another side effect that the Msg.Data property will be a default value (null for objects, 0 for ints etc.)

> Also looking at [Go](https://github.com/nats-io/nats.go/blob/8712190da1d17ab0c4719bffa7c0174214c56e6c/nats.go#L2502) and [Rust](https://github.com/nats-io/nats.rs/blob/25c47a6faf5f82737f60b49757ce96b831890649/async-nats/src/connector.rs#L175) clients there isn't an option to set `no_responders`. If the client is supporting headers `no_responders` is implied.

In that case we should default the `NatsOpts.NoResponders` property to true, so it can rather be explicitly disabled for clients if they want, since this client does support headers. Or should it just not be configurable at all?

mtmk:
> > Maybe we should make it a property on the message itself? ([Rust client seems to be doing that](https://github.com/nats-io/nats.rs/blob/aebbe14e65324150fd28c19fd256e08915020057/nats/src/message.rs#L104))
> 
> That's possible. I still think an exception is more intuitive though from a C# convention perspective (compared to the NatsNoReplyException for timeouts), unless explicitly documented, since another side effect is that the Msg.Data property will be a default value (null for objects, 0 for ints etc.)
> 

right, I think exception is the right thing to do. I was suggesting the way we detect no responders can be a calculated property on the message object.

> > Also looking at [Go](https://github.com/nats-io/nats.go/blob/8712190da1d17ab0c4719bffa7c0174214c56e6c/nats.go#L2502) and [Rust](https://github.com/nats-io/nats.rs/blob/25c47a6faf5f82737f60b49757ce96b831890649/async-nats/src/connector.rs#L175) clients there isn't an option to set `no_responders`. If the client is supporting headers `no_responders` is implied.
> 
> In that case we should default the `NatsOpts.NoResponders` property to true, so it can rather be explicitly disabled for clients if they want, since this client does support headers. Or should it just not be configurable at all?

I'm thinking less options the better but if you say there may be a use case for it we could leave it in the options but yes I think it should default to true if we left it as an option.

stebet:
@mtmk 
* Removed the NoResponders option (don't really see a use-case for disabling it).
* Added a `bool IsNoRespondersError` property to NatsMsg for easy checking.

mtmk:
Excellent. thank you @stebet! One last thing, should we have the check in `RequestMany` as well for completeness? Is that something you've seen being used?

mtmk:
@stebet don't worry about tests. there may be quite a few failing I can fix them later.

stebet:
> @stebet don't worry about tests. there may be quite a few failing I can fix them later.

Fixed the ones that my changes broke :) which were counting on timeouts when they were received NoResponders instead. We could also just drop the `NatsNoRespondersException` and just use the existing `NatsNoReplyException` with a clear "No responders" message since it's implied that you will never get a reply if there is no responder?

mtmk:
> > @stebet don't worry about tests. there may be quite a few failing I can fix them later.
> 
> Fixed the ones that my changes broke :) which were counting on timeouts when they were received NoResponders instead. We could also just drop the `NatsNoRespondersException` and just use the existing `NatsNoReplyException` with a clear "No responders" message since it's implied that you will never get a reply if there is no responder?

Thank you :-) yes good point. would there be a need to distinguish the two? e.g. trying to pick up the fastest responder in a set of responders?

stebet:
> Thank you :-) yes good point. would there be a need to distinguish the two? e.g. trying to pick up the fastest responder in a set of responders?

Not apart from the exception message I think. They're sort of equivalent, one will just be returned pretty much instantly (no responders) since there is no responder that will ever answer, or they simply took too long and the request timed out (exceeded the NatsOpts.RequestTimeout threshold)

We can also just keep them separate for clarity to avoid confusion.

src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -141,4 +141,6 @@ private void CheckReplyPreconditions()
             throw new NatsException("unable to send reply; ReplyTo is empty");
         }
     }
+
+    public bool IsNoRespondersError => Headers?.Code == 503;
```
caleblloyd, line 145: I don't know if it makes sense to offer just 1 public `IsNoRespondersError` here.  If we do one, we should probably enumerate every single error code.  I don't know if using Go or Rust as examples is great.  Those languages don't have exceptions.

caleblloyd:
I am pretty sure that Go will return a No Responders error in any subscription if it receives one, not just in the Request-Reply APIs:

https://github.com/nats-io/nats.go/blob/48e070d968fc25ce61883c47bf4cba8a677ba0ff/nats.go#L4682-L4684

Also, I do think they should be separate Exceptions.  `NatsNoRespondersException` is immediate and indicates that no service is listneing.  Wheras `NatsNoReplyException` indicates that there was interest in the subject at the time it was sent, but a client-side timeout ocurred.  So I like the separate Exceptions here.

src/NATS.Client.Core/NatsMsg.cs
mtmk: Not a standard error very specific [message from the server](https://github.com/nats-io/nats-server/blob/d96b489963cca10981b45342f9fdb008c2f7bf88/server/client.go#L3922). Other errors should still be check from the headers. we can't rely throw exceptions at this level without destroying the subscription.

⚠️ we should also check that payload is empty as other clients are doing. We don't have a reliable way of doing that (so ignore it for this issue). I'll create a separate issue for that.

mtmk:
> I am pretty sure that Go will return a No Responders error in any subscription if it receives one, not just in the Request-Reply APIs:

We can't throw exception per message. We'd need to end the subscription.

mtmk:
see also #262

mtmk:
I somewhat disagree with handling no-responders at subscription level having said that I also can't see a nice way of doing that. But maybe I'm missing something.

stebet:
No responders at a subscription level is an edge case imo so I don't think it warrants an exception. It should never happen unless the client is explicitly doing his own response-request mechanism, since it's only returned to the reply_to subject if the connection is the same as the one making the publish. It's not sent if the connection listening to the reply_to subject is different from the one making the publish.

caleblloyd:
> > I am pretty sure that Go will return a No Responders error in any subscription if it receives one, not just in the Request-Reply APIs:
> 
> We can't throw exception per message. We'd need to end the subscription.

Ah good point.  Then it's probably fine just to throw in the Request-Reply patterns

> a class with constant values for possible status codes might be a good idea for a later PR to make it clearer in the code what each status code means (similar to the [StatusCodes](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.statuscodes?view=aspnetcore-7.0) constants in the ASP.NET Core libraries).

That sounds like a good idea!

mtmk:
@stebet please resolve conflict then we can merge. thanks!

stebet:
> @stebet please resolve conflict then we can merge. thanks!

Rebased on latest main

mtmk (approved):
LGTM thanks @stebet!

PR #266, opened 2023-11-30
Create an INatsJSMsg<T> interface

mtmk:
An INatsJSMsg interface can help with unit testing in common .net (see also #263)

* There maybe performance implications due to struct to class change assuming not noticeable. Struct to class change should not introduce functional change sine it is immutable.

* This would be a breaking change but the impact would be relatively small in most code bases.


caleblloyd:
I thought we were originally avoiding this to prevent boxing `NatsMsg` and `NatsJSMsg` because they are structs.

Also, only offering `INatsJSMsg` and not `INatsMsg` feels inconsistent.

Is there another way to introduce easier testability without sacrificing performance?

mtmk:
> I thought we were originally avoiding this to prevent boxing `NatsMsg` and `NatsJSMsg` because they are structs.

That's true. My thinking was `NatsJSMsg` being on the JetStream side would not be the performance bottleneck since JS is relatively slower because server has to deal with persistence.

I should run a few benchmarks as well.

> 
> Also, only offering `INatsJSMsg` and not `INatsMsg` feels inconsistent.

This is a very good point.

> 
> Is there another way to introduce easier testability without sacrificing performance?

Another way I think we can achieve this is to have the structs implement both `INatsJSMsg` and `INatsMsg` but not use them. That way nothing would change. Then on the consumer side optionally message can be cast / or implicitly converted to the interface if application code wants to use the interface e.g. `Process(INatsJSMsg)`. At that point boxing which would be a choice.

mtmk:
Please have a look the last commit for another way of providing testability. Only when you pass the message to a method accepting the message interface you'd pay the performance penalty, which is a boxing allocation. This also isn't a breaking change.

src/NATS.Client.JetStream/NatsJSMsg.cs
```diff
@@ -6,17 +6,113 @@
 
 namespace NATS.Client.JetStream;
 
+public interface INatsJSMsg<out T>
```
to11mtm, line 9: Same as `INatsMsg<T>`, we should annotate this type for awareness.

to11mtm (commented):
As odd as it looks to me I get it; We should however have XMLDocs to avoid footguns, given the boxing cost, iirc for INatsJSMsg it's 36-68 bytes on top of the object header and however large `T` is (sizeof pointer or Struct `T`)

src/NATS.Client.JetStream/NatsJSMsg.cs
mtmk: added a few lines. please feel free to suggest edit or rewrite

caleblloyd:
I don't know if I am sold on doing this in the Core Assembly.  `INatsConnection` already does not return the interface, so anyone wanting to add a mock for `INatsConnection` is going to need to add a wrapping type.

Seems like it might be better to provide another assembly for testing, and provide an `IBoxedNatsConnection` or something, with all of the types wrapped

src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -3,6 +3,77 @@
 
 namespace NATS.Client.Core;
 
+public interface INatsMsg<T>
```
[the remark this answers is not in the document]
to11mtm: Looks great, really appreciate the exposition here!

mtmk:
> I don't know if I am sold on doing this in the Core Assembly.

We have to for `INatsMsg`s you can't make a type implement a class you don't own. It's no Go you know 😛 


> `INatsConnection` already does not return the interface, so anyone wanting to add a mock for `INatsConnection` is going to need to add a wrapping type.
> 

true. There is also `INatsSub`. But i don't think we need to worry about transient dependencies that much. as long as we make it easy to construct `NatsMsg` types it maybe acceptable.

> Seems like it might be better to provide another assembly for testing, and provide an `IBoxedNatsConnection` or something, with all of the types wrapped

Wouldn't that be a little confusing? Could we give an example of it in docs rather? shouldn't be hard to implement.

caleblloyd:
I guess I am not understanding how it would be practically used in a test without a supporting wrapper.  Say my code uses an `INatsConnection` from DI.  It subscribes to messages using that, and gets `NatsMsg<T>` out of that.

How do I now create a test that starts using `INatsMsg<T>`?

Wouldn't I need to sub out my `INatsConnection` in DI for some wrapping interface that returns `INatsMsg<T>`, and use that throughout my code?

mtmk:
> I guess I am not understanding how it would be practically used in a test without a supporting wrapper. Say my code uses an `INatsConnection` from DI. It subscribes to messages using that, and gets `NatsMsg<T>` out of that.
> 
> How do I now create a test that starts using `INatsMsg<T>`?
> 
> Wouldn't I need to sub out my `INatsConnection` in DI for some wrapping interface that returns `INatsMsg<T>`, and use that throughout my code?

Yes that's right except you'd use e.g. `INatsConnectionWrapper`. My point was if `NATS.Client.Core` doesn't have a definition for `INatsMsg<T>` we won't be able to create a wrapper because the real methods will be returning the concrete type `NatsMsg<T>` which must implement `INatsMsg<T>`.

```csharp
interface INatsConnectionWrapper
{
  ValueType<INatsMsg<T>> PublishAsync<T>(...);
}

class NatsConnectionWrapper : INatsConnectionWrapper
{
  public async ValueType<INatsMsg<T>> PublishAsync<T>(...)
  {
     NatsMsg<T> msg = await _connection.PublishAsync<T>(...);
     return msg; // NatsMsg<T> must implement INatsMsg<T> in NATS.Client.Core assembly
  }
}
```

Second point is if using a wrapper isn't prefered then it should be easy to construct a `NatsMsg<T>` or `NatsJSMsg<T>` so the mocked `INatsConnection` methods can return an instance with test values.

mtmk:
> I was advocating for developing the wrapper in a separate assembly, and maybe trying to figure out a way to shoehorn `INatsMsg<T>` into that assembly as well.

Oh I see, like a wrapper for `NatsMsg<T>` as well?

> Is it worthwhile to still create the wrapper though? Maybe in a separate assembly, or even in the core assembly? Or should we leave that up to the end-user?

We could do an assembly without touching core if we wrapped `NatsMsg<T>` but would you want to use it knowing it's going to affect application performance quite a bit?

Or implement `INatsMsg` in Core and have an `INatsConnectionWrapper` (or same name but in a different namespace); would this new package be confusing?

...or provide an example in docs and leave it to the application to copy & paste it? I think I'd be in favor of this approach since it'd be fairly easy to develop a custom wrapper and we avoid confusion in nuget packages.

Unfortunately anyway we do it it's not a great user experience when writing tests but I think this is the best we can do balancing performance concerns and providing easier to test APIs.

caleblloyd:
> ...or provide an example in docs and leave it to the application to copy & paste it? I think I'd be in favor of this approach since it'd be fairly easy to develop a custom wrapper and we avoid confusion in nuget packages.

That sounds good

caleblloyd (approved):
LGTM

PR #267, opened 2023-11-30
Fixing test flappers

mtmk:


src/NATS.Client.JetStream/Internal/NatsJSNotificationChannel.cs
```diff
@@ -25,7 +25,7 @@ internal class NatsJSNotificationChannel : IAsyncDisposable
             SingleWriter = false,
             FullMode = BoundedChannelFullMode.DropOldest,
         });
-        _loop = Task.Run(NotificationLoop, _cancellationToken);
```
mtmk, line 28: We should not cancel the task but let the notifications queue drain.

tests/NATS.Client.Perf/Program.cs
```diff
@@ -11,7 +11,7 @@
     Subject = "test",
     PubTasks = 10,
     MaxNatsBenchRatio = 0.20,
-    MaxMemoryMb = 500,
+    MaxMemoryMb = 600,
```
mtmk, line 14: Not sure why it's this high on github instances. On my local Linux VM it's about 150M.

caleblloyd: Hmm, mine is around 325M on a pretty fast Linux Desktop.  Probably something about # of Cores (is your VM 1-core?) plus the slow GitHub instance CPUs causing buffers to build up

PR #273, opened 2023-12-06
Watcher fix and API naming fix

mtmk:
* Fixed watcher not to exit after receiving initial set of objects. Introduced a new option to accomodate for that behaviour rather than returning null for example.
* Also fixed the method naming to be affixed 'Async'.

**Breaking change:**
* Renamed `INatsObjContext` `CreateObjectStore()` method to `CreateObjectStoreAsync()`


caleblloyd (approved):
LGTM

PR #274, opened 2023-12-06
Added info and stats to services

mtmk:
Added info and stats accessors to service object as they are returned from the service query APIs $SRV.INFO and STATS

### DIFF
--- a/src/NATS.Client.Services/INatsSvcServer.cs
+++ b/src/NATS.Client.Services/INatsSvcServer.cs
@@ -1,4 +1,5 @@
 using NATS.Client.Core;
+using NATS.Client.Services.Models;
 
 namespace NATS.Client.Services;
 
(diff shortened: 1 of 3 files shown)


src/NATS.Client.Services/INatsSvcServer.cs
```diff
@@ -39,4 +40,16 @@ public interface INatsSvcServer : IAsyncDisposable
...
+    /// Get current stats for the service.
+    /// </summary>
+    /// <returns>Stats response object</returns>
+    StatsResponse GetStats();
+
+    /// <summary>
+    /// Get current info for the service.
+    /// </summary>
+    /// <returns>Info response object</returns>
+    InfoResponse GetInfo();
```
caleblloyd, line 54: Only comment is naming - I always second-guess myself when adding `Get` as a prefix of a method name.

Would `Stats()` and `Info()` be better?  They also match the ADR.

oising: > Only comment is naming - I always second-guess myself when adding `Get` as a prefix of a method name.
> 
> Would `Stats()` and `Info()` be better? They also match the ADR.

using a Get prefix is much more aligned with expectations of dotnet programmers.

mtmk: Ah I knew why i named them like that. I am one 🤣 

renmed already to without 'get'. what do you all think? revert?

caleblloyd: Aren't paramaterless Get methods typically discouraged, in favor of `{get;}` properties?  So `Stats {get;}` and `Info {get;}`?

I guess the caveat for that is when it's a computationally expensive method, sometimes I see it as `GetFoo()` in order to make 

These didn't strike me as computationally expensive though?

I think I just try to stay away from the paramaterless GetFoo() method because of the deep scars that Java left

caleblloyd: I am fine with whatever the consensus is - `Stats {get;}` or `GetStats()`.  I wasn't aware of BCL Guidelines of methods starting with Verbs but it makes sense, thanks for sharing.

mtmk: reverted. let me know if you prefer properties although I think they are a bit too busy for that especially stats, I can make them getter only properties.

caleblloyd (approved):
LGTM

PR #277, opened 2023-12-08
handle no responders in NatsSubBase

caleblloyd:
Handles No Responders in `NatsSubBase`, pre-de-serialization

Also handles `StopOnEmptyMsg`

This prevents the library from ever needing to check `msg.Data` for an Empty value

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.RequestReply.cs
+++ b/src/NATS.Client.Core/NatsConnection.RequestReply.cs
@@ -10,6 +10,18 @@ public partial class NatsConnection
     /// <inheritdoc />
     public string NewInbox() => NewInbox(InboxPrefix);
 
+    private static readonly NatsSubOpts ReplyOptsDefault = new NatsSubOpts
+    {
+        MaxMsgs = 1,
+        ThrowIfNoResponders = true,
+    };
+
+    private static readonly NatsSubOpts ReplyManyOptsDefault = new NatsSubOpts
+    {
+        StopOnEmptyMsg = true,
+        ThrowIfNoResponders = true,
+    };
+
     /// <inheritdoc />
     public async ValueTask<NatsMsg<TReply>> RequestAsync<TRequest, TReply>(
         string subject,
@@ -98,14 +110,18 @@ public partial class NatsConnection
 
     private NatsSubOpts SetReplyOptsDefaults(NatsSubOpts? replyOpts)
     {
-        var opts = replyOpts ?? new NatsSubOpts { MaxMsgs = 1 };
+        var opts = replyOpts ?? ReplyOptsDefault;
+        if (!opts.MaxMsgs.HasValue)
+        {
+            opts = opts with { MaxMsgs = 1 };
+        }
+
         return SetBaseReplyOptsDefaults(opts);
     }
 
     private NatsSubOpts SetReplyManyOptsDefaults(NatsSubOpts? replyOpts)
     {
-        var opts = replyOpts ?? new NatsSubOpts();
-
+        var opts = replyOpts ?? ReplyManyOptsDefault;
         if (!opts.StopOnEmptyMsg.HasValue)
         {
             opts = opts with { StopOnEmptyMsg = true };
--- a/src/NATS.Client.Core/NatsConnection.RequestReply.cs
+++ b/src/NATS.Client.Core/NatsConnection.RequestReply.cs
@@ -7,8 +7,6 @@ namespace NATS.Client.Core;
 
 public partial class NatsConnection
 {
-    private static readonly NatsSubOpts DefaultReplyOpts = new() { MaxMsgs = 1 };
-
     /// <inheritdoc />
     public string NewInbox() => NewInbox(InboxPrefix);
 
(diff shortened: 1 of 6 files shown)


src/NATS.Client.Core/NatsSubBase.cs
```diff
@@ -193,6 +195,20 @@ public virtual async ValueTask ReceiveAsync(string subject, string? replyTo, Rea
     {
         ResetIdleTimeout();
 
+        // check for empty payload conditions
+        if (payloadBuffer.Length == 0)
+        {
+            switch (Opts)
+            {
+            case { ThrowIfNoResponders: true } when headersBuffer is { Length: >= 12 } && headersBuffer.Value.Slice(8, 4).ToArray().SequenceEqual(NoRespondersHeaderSequence):
```
caleblloyd, line 203: The header code checking here is a little raw.  In the future, we might want to consider moving header parsing up the stack so that we have an actual `NatsHeaders` object here.

stebet:
Does this impþementation kill the subscription? It's important in my opinion that the subscription be left intact for custom implementations of request-reply. My use case for example needs the subscription to stay intact for performance reasons (avoids the overhead of a new inbox + subscription for every request).

caleblloyd:
> Does this impþementation kill the subscription?

Yes. But you can explicitly set NatsSubOpts.ThrowIfNoResponders=false and perform the check for a 503 code in the headers yourself.

caleblloyd:
Note that Request-Reply uses the muxed Inbox Subscription by default, which muxes all subscriptions over a single persistent NATS Subscription. So there's no additional NATS Subscription overhead on Inbox Subscriptions (except for the first one established).

mtmk:
I'm mainly concerned about auto unsubs. especially empty message check silently unsubs which could be source of hard to find bugs. also no-responders check may be OK but that only happens on reply inbox subscriptions and not all subscriptions.

mtmk:
If you chose to implement this feel free to get rid of #275

src/NATS.Client.Core/NatsConnection.RequestReply.cs
```diff
@@ -111,13 +98,34 @@ string Throw()
 
     private NatsSubOpts SetReplyOptsDefaults(NatsSubOpts? replyOpts)
     {
-        var opts = replyOpts ?? DefaultReplyOpts;
+        var opts = replyOpts ?? new NatsSubOpts { MaxMsgs = 1 };
```
mtmk, line 101: why aren't we avoiding an extra object allocation here?

src/NATS.Client.Core/NatsConnection.RequestReply.cs
```diff
@@ -111,13 +98,34 @@ string Throw()
...
+        return SetBaseReplyOptsDefaults(opts);
+    }
+
+    private NatsSubOpts SetReplyManyOptsDefaults(NatsSubOpts? replyOpts)
+    {
+        var opts = replyOpts ?? new NatsSubOpts();
```
mtmk, line 107: ..and here

src/NATS.Client.Core/NatsSubBase.cs
mtmk: `SequenceEqual` should work on span, no?

src/NATS.Client.Core/NatsSubOpts.cs
```diff
@@ -42,6 +42,26 @@ public record NatsSubOpts
...
+    /// </summary>
+    /// <remarks>
+    /// If not set, all published messages will be received until explicitly
+    /// unsubscribed or disposed.
+    /// </remarks>
+    public bool? StopOnEmptyMsg { get; init; }
+
+    /// <summary>
+    /// If true, throw an exception if a no responders status message is received.
+    /// The no responders status message will not be delivered to the subscription.
```
mtmk, line 57: can we add a note about unsub here?

mtmk (commented):
a few nits

src/NATS.Client.Core/NatsSubBase.cs
caleblloyd: What if it's multi-span?  Or is it guaranteed to never be, so could just check for this on the first span?

src/NATS.Client.Core/NatsConnection.RequestReply.cs
caleblloyd: I will revert...  I guess I was making the (likely wrong) assumption that the compiler would optimize

src/NATS.Client.Core/NatsSubOpts.cs
caleblloyd: added

src/NATS.Client.Core/NatsSubBase.cs
caleblloyd: ah... `ToSpan` instead of `ToArray` should address - made that change

mtmk (approved):
LGTM

mtmk:
Thank you @caleblloyd! was there any other change you wanted to make before merge?

caleblloyd:
Nope- just merged!

PR #283, opened 2023-12-12
Read certificates as PEM strings or from files

mtmk:


caleblloyd:
What do you think about offering a couple of callbacks instead?  Then some static methods to help configure the callbacks for common cases?  I can see many more requests in this area on the horizon - PFX certificates, private keys with passwords, etc

```
public sealed record NatsTlsOpts {
    public Func<ValueTask<X509Certificate2>>? LoadClientCert { get; init; }

    public Func<ValueTask<X509Certificate2Collection>>? LoadCaCerts { get; init; }

    public static Func<ValueTask<X509Certificate2>> LoadClientCertFromPemString(string certPem, string keyPem)
    {
        var clientCert = X509Certificate2.CreateFromPem(certPem, keyPem);
        return () => ValueTask.FromResult(clientCert);
    }

    public static Func<ValueTask<X509Certificate2Collection>> LoadCaCertsFromPemString(string caPem)
    {
        var caCerts = new X509Certificate2Collection();
        caCerts.ImportFromPem(caPem);
        return () => ValueTask.FromResult(caCerts);
    }
}
```

Note that this is not a full replacement for `LocalCertificateSelectionCallback` and `RemoteCertificateValidationCallback` - it's just a way to load other X509 Certificates and use them with our default `LocalCertificateSelectionCallback` and `RemoteCertificateValidationCallback` implementations.  I can also see being asked to provide overrides for `LocalCertificateSelectionCallback` and `RemoteCertificateValidationCallback` in the future, but of course we can address that later.

mtmk:
@caleblloyd why don't you just hack it into the branch? Not sure exactly what you mean.

caleblloyd:
Sure, will open a PR to this one

caleblloyd (approved):
LGTM

PR #284, opened 2023-12-12
move string-based PEM loading to callbacks

caleblloyd:
Note - PR is to branch opened for #283 

Suggestion to move string-based PEM loading to callbacks, which will enable other functionality in callbacks too - different cert formats, private keys with passwords, load from external store, etc

### DIFF
--- a/src/NATS.Client.Core/NatsTlsOpts.cs
+++ b/src/NATS.Client.Core/NatsTlsOpts.cs
@@ -1,3 +1,4 @@
+using System.Security.Cryptography.X509Certificates;
 using NATS.Client.Core.Internal;
 
 namespace NATS.Client.Core;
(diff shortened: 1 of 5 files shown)


src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -45,47 +46,37 @@ public sealed record NatsTlsOpts
...
-    /// <remarks>
-    /// You can use <see cref="NatsTlsOpts"/> to load the private key from a file
-    /// or you can pass the PEM-encoded private key directly as string. If the value
-    /// looks like a PEM-encoded pattern (for example if it starts with "-----BEGIN PRIVATE KEY-----")
-    /// it would be used as PEM-encoded string otherwise it would be treated as file path.
-    /// </remarks>
     public string? KeyFile { get; init; }
 
+    /// Callback that loads Client Certificate
+    public Func<ValueTask<X509Certificate2>>? LoadClientCert { get; init; }
```
caleblloyd, line 57: Is it typical to use the `Async` suffix on a callback that returns a Task/ValueTask?  Should this be `LoadClientCertAsync`?

caleblloyd:
Odd to be getting a test failure here - https://github.com/nats-io/nats.net.v2/actions/runs/7184322773/job/19565121688?pr=284#step:7:138

@mtmk is there anything special that needs to be done timing-wise - like wait for server shutdown?

mtmk:
> Odd to be getting a test failure here - https://github.com/nats-io/nats.net.v2/actions/runs/7184322773/job/19565121688?pr=284#step:7:138
> 
> @mtmk is there anything special that needs to be done timing-wise - like wait for server shutdown?

it looked like an intermittent server start error. Other versions passed so, not sure. Running again will have a closer look if it persists.

src/NATS.Client.Core/NatsTlsOpts.cs
mtmk: Not sure about the naming. I wouldn't've thought so.

btw why do we need a callback? can it be just `X509Certificate2`?

caleblloyd: Callback is so we can support reloading certs - for example if they expire.  Right now `TlsCerts` is constructed on every initial connect in `NatsConnection.InitialConnectAsync`

Although this behavior should probably be documented.

mtmk: of course! makes sense.

mtmk: > Right now `TlsCerts` is constructed on every initial connect in `NatsConnection.InitialConnectAsync`

should we reload on reconnect? (not for this PR)

mtmk (approved):
LGTM thank you @caleblloyd 💯

PR #287, opened 2023-12-14
TLS Check Certificate Revocation option

mtmk:
Looks like OCSP must-staple support is implemented in .NET7 and should be in .NET8. Not sure if .NET6 received a patch as well. https://github.com/dotnet/runtime/issues/33377

### DIFF
--- a/src/NATS.Client.Core/NatsTlsOpts.cs
+++ b/src/NATS.Client.Core/NatsTlsOpts.cs
@@ -77,6 +77,8 @@ public sealed record NatsTlsOpts
     /// <summary>When true, skip remote certificate verification and accept any server certificate</summary>
     public bool InsecureSkipVerify { get; init; }
 
+    /// <summary>Certificate revocation mode for certificate validation.</summary>
+    /// <value>One of the values in <see cref="T:System.Security.Cryptography.X509Certificates.X509RevocationMode" />. The default is <see langword="NoCheck" />.</value>
     public X509RevocationMode CertificateRevocationCheckMode { get; init; }
 
     /// <summary>TLS mode to use during connection</summary>
--- a/src/NATS.Client.Core/NatsTlsOpts.cs
+++ b/src/NATS.Client.Core/NatsTlsOpts.cs
@@ -77,7 +77,7 @@ public sealed record NatsTlsOpts
     /// <summary>When true, skip remote certificate verification and accept any server certificate</summary>
     public bool InsecureSkipVerify { get; init; }
 
-    public bool CheckCertificateRevocation { get; init; }
+    public X509RevocationMode CertificateRevocationCheckMode { get; init; }
 
     /// <summary>TLS mode to use during connection</summary>
     public TlsMode Mode { get; init; }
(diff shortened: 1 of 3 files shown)


tests/NATS.Client.Core.Tests/TlsClientTest.cs
```diff
@@ -22,6 +22,25 @@ public async Task Client_connect_using_certificate()
...
+            new NullOutputHelper(),
+            new NatsServerOptsBuilder()
+                .UseTransport(TransportType.Tls, tlsVerify: true)
+                .Build());
+
+        var clientOpts = server.ClientOpts(NatsOpts.Default with { Name = "tls-test-client" });
+        clientOpts = clientOpts with { TlsOpts = clientOpts.TlsOpts with { CheckCertificateRevocation = true } };
+        await using var nats = new NatsConnection(clientOpts);
+
+        // At the moment I don't know of a good way of checking if the revocation check is working
```
caleblloyd, line 38: We should probably try and come up with a real test.  If it's not supported in .NET 6 then we should probably conditionally compile the option.  I wonder how hard it'd be to use a self-signed CA that supports online revocations.

mtmk: > We should probably try and come up with a real test.

I don't really know how OCSP must-staple work. checked what mongodb driver is doing looked fairly involved to setup https://github.com/search?q=repo%3Amongodb%2Fmongo-csharp-driver%20ocsp&type=code but the c# code is just enabling this option.

> If it's not supported in .NET 6 then we should probably conditionally compile the option. I wonder how hard it'd be to use a self-signed CA that supports online revocations.

I was referring to ocsp stapling for .net6 AFAIU other types of online checks would still work. no need to exclude the option.

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -77,6 +77,8 @@ public sealed record NatsTlsOpts
     /// <summary>When true, skip remote certificate verification and accept any server certificate</summary>
     public bool InsecureSkipVerify { get; init; }
 
+    public bool CheckCertificateRevocation { get; init; }
```
caleblloyd, line 80: Should we change this to be 

```
public X509RevocationMode CertificateRevocationCheckMode {get; init;}
```

By making it a bool that flips between `NoCheck` and `Online`, we are not allowing the option for the caller to select `Online`

tests/NATS.Client.Core.Tests/TlsClientTest.cs
caleblloyd: > I was referring to ocsp stapling for .net6 AFAIU other types of online checks would still work. no need to exclude the option.

Oh good point

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -77,6 +77,8 @@ public sealed record NatsTlsOpts
     /// <summary>When true, skip remote certificate verification and accept any server certificate</summary>
     public bool InsecureSkipVerify { get; init; }
 
+    public X509RevocationMode CertificateRevocationCheckMode { get; init; }
```
caleblloyd, line 80:
```suggestion
    /// <summary>Certificate revocation mode for certificate validation.</summary>
    /// <value>One of the values in <see cref="T:System.Security.Cryptography.X509Certificates.X509RevocationMode" />. The default is <see langword="NoCheck" />.</value>
    public X509RevocationMode CertificateRevocationCheckMode { get; init; }
```

caleblloyd (approved):
Added a suggestion for a summary on the new option.  Then LGTM

PR #288, opened 2023-12-14
Handle JetStream publish no-responders

mtmk:
During cluster reconnect we often get no-responders exceptions. This change makes sure JetStream publish method will retry in that case too.

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSContext.cs
+++ b/src/NATS.Client.JetStream/NatsJSContext.cs
@@ -135,22 +135,29 @@ public partial class NatsJSContext
                         Timeout = Connection.Opts.RequestTimeout,
 
                         // If JetStream is disabled, a no responders error will be returned
+                        // No responders error might also happen when reconnecting to cluster
                         ThrowIfNoResponders = true,
                     },
                     cancellationToken)
                 .ConfigureAwait(false);
 
-            while (await sub.Msgs.WaitToReadAsync(cancellationToken).ConfigureAwait(false))
+            try
             {
-                while (sub.Msgs.TryRead(out var msg))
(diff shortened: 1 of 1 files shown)


caleblloyd:
So does this happen when we reconnect to a server that is running, but hasn't started its JetStream service yet?  I think making it retryable is OK - how many retries do we try by default, and how long do we wait?

Ideally this would be solved by the infrastructure layer not marking a JetStream server as ready and serving traffic to that before JetStream has started.

mtmk:
> So does this happen when we reconnect to a server that is running, but hasn't started its JetStream service yet? I think making it retryable is OK - how many retries do we try by default, and how long do we wait?

by default two retries with 250ms gaps (mimicking Go client if I read correctly https://github.com/nats-io/nats.go/blob/bb64e1b6e378dcafbbffe33145db0b7ee064485a/js.go#L548)
 
> Ideally this would be solved by the infrastructure layer not marking a JetStream server as ready and serving traffic to that before JetStream has started.

I see this in my local test with an r3 cluster kill -9 and restart servers at random.

caleblloyd (approved):
LGTM

caleblloyd:
Ah, the comments in the Go client also mentions leadership changes.  So maybe some of these handlers only run on the Leader, and the time spent isn't waiting for JetStream to start, but it's waiting for a leadership change.

mtmk:
> Ah, the comments in the Go client also mentions leadership changes. So maybe some of these handlers only run on the Leader, and the time spent isn't waiting for JetStream to start, but it's waiting for a leadership change.

yes that could be it. on consume side we see the specific leadership change events but no when publishing.

PR #296, opened 2023-12-21
Fixed serialization problem in PublishModel/PublishHeaders examples

robibobi:
Currently the example projects `Example.Core.PublishModel`, `Example.Core.SubscribeModel`, `Example.Core.PublishHeaders` and `Example.Core.SubscribeHeaders` are not working because the default serializer can not handle the custom `Bar` type. To make the programs work, we need to provide a serializer for that type. 

To achieve this, i added a project reference to `Nats.Client.Serializers.Json` and configured the `NatsJsonSerializerRegistry` as SerializerRegistry in the mentioned projects.

The same could be achieved by using the AOT ready `NatsJsonContextSerializer` but with we would need to create a custom `JsonSerializerContext` for the `Bar` type in every project. To keep the required code as small as possible, i chose the reflection based serializer.


mtmk (approved):
LGTM thank you @robibobi

PR #297, opened 2023-12-22
Suggested fixes to interfaces

sspates:
Fixes to ensure all public methods on the class are available on the interface for INatsConnection.
Test added to prove fixes

NatsConnectionPool adjustment to ensure pooled connections can't be accidentally disposed of.

### DIFF
--- a/src/NATS.Client.Core/INatsConnection.cs
+++ b/src/NATS.Client.Core/INatsConnection.cs
@@ -15,6 +15,7 @@ public interface INatsConnection : IAsyncDisposable
     NatsOpts Opts { get; }
 
     NatsConnectionState ConnectionState { get; }
+
     NatsHeaderParser HeaderParser { get; }
 
 
--- a/src/NATS.Client.Core/INatsConnection.cs
+++ b/src/NATS.Client.Core/INatsConnection.cs
@@ -16,9 +16,6 @@ public interface INatsConnection : IAsyncDisposable
 
     NatsConnectionState ConnectionState { get; }
 
-    NatsHeaderParser HeaderParser { get; }
-
-
     /// <summary>
     /// Send PING command and await PONG. Return value is similar as Round Trip Time (RTT).
     /// </summary>
@@ -177,5 +174,8 @@ public interface INatsConnection : IAsyncDisposable
     /// </summary>
     ValueTask ConnectAsync();
 
+    /// <summary>
+    /// Gets the statistics for the current Nats Connection
+    /// </summary>
     NatsStats GetStats();
 }
--- a/src/NATS.Client.Core/INatsConnection.cs
+++ b/src/NATS.Client.Core/INatsConnection.cs
@@ -2,14 +2,6 @@ namespace NATS.Client.Core;
 
 public interface INatsConnection : IAsyncDisposable
 {
-    event EventHandler<string>? ConnectionDisconnected;
-
-    event EventHandler<string>? ConnectionOpened;
-
-    event EventHandler<string>? ReconnectFailed;
-
-    event EventHandler<INatsError>? OnError;
-
     INatsServerInfo? ServerInfo { get; }
 
     NatsOpts Opts { get; }
@@ -173,9 +165,4 @@ public interface INatsConnection : IAsyncDisposable
     /// Connect socket and write CONNECT command to nats server.
     /// </summary>
     ValueTask ConnectAsync();
-
-    /// <summary>
-    /// Gets the statistics for the current Nats Connection
-    /// </summary>
-    NatsStats GetStats();
 }
--- a/src/NATS.Client.Core/INatsConnection.cs
+++ b/src/NATS.Client.Core/INatsConnection.cs
@@ -1,7 +1,23 @@
 namespace NATS.Client.Core;
 
-public interface INatsConnection
+public interface INatsConnection : IAsyncDisposable
 {
+    event EventHandler<string>? ConnectionDisconnected;
+
+    event EventHandler<string>? ConnectionOpened;
+
+    event EventHandler<string>? ReconnectFailed;
+
+    event EventHandler<INatsError>? OnError;
+
+    INatsServerInfo? ServerInfo { get; }
+
+    NatsOpts Opts { get; }
+
+    NatsConnectionState ConnectionState { get; }
+    NatsHeaderParser HeaderParser { get; }
+
+
     /// <summary>
     /// Send PING command and await PONG. Return value is similar as Round Trip Time (RTT).
     /// </summary>
@@ -154,4 +170,11 @@ public interface INatsConnection
         NatsPubOpts? requestOpts = default,
         NatsSubOpts? replyOpts = default,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Connect socket and write CONNECT command to nats server.
+    /// </summary>
+    ValueTask ConnectAsync();
+
+    NatsStats GetStats();
 }
--- a/src/NATS.Client.Core/NatsPooledConnection.cs
+++ b/src/NATS.Client.Core/NatsPooledConnection.cs
@@ -8,4 +8,6 @@ internal class NatsPooledConnection : NatsConnection
     }
 
     public override ValueTask DisposeAsync() => ValueTask.CompletedTask;
+
+    internal ValueTask ForceDisposeAsync() => base.DisposeAsync();
 }
--- a/src/NATS.Client.Core/NatsPooledConnection.cs
+++ b/src/NATS.Client.Core/NatsPooledConnection.cs
@@ -1,0 +1,11 @@
+namespace NATS.Client.Core;
+
+internal class NatsPooledConnection : NatsConnection
+{
+    public NatsPooledConnection(NatsOpts opts)
+        : base(opts)
+    {
+    }
+
+    public override ValueTask DisposeAsync() => ValueTask.CompletedTask;
+}
--- a/src/NATS.Client.Core/NatsPooledConnection.cs
+++ b/src/NATS.Client.Core/NatsPooledConnection.cs
@@ -1,6 +1,6 @@
 namespace NATS.Client.Core;
 
-internal class NatsPooledConnection : NatsConnection
+internal sealed class NatsPooledConnection : NatsConnection
 {
     public NatsPooledConnection(NatsOpts opts)
         : base(opts)
--- a/src/NATS.Client.JetStream/INatsJSContext.cs
+++ b/src/NATS.Client.JetStream/INatsJSContext.cs
@@ -1,4 +1,3 @@
-using System.Runtime.CompilerServices;
 using NATS.Client.Core;
 using NATS.Client.JetStream.Models;
 
--- a/src/NATS.Client.JetStream/INatsJSContext.cs
+++ b/src/NATS.Client.JetStream/INatsJSContext.cs
@@ -1,3 +1,4 @@
+using System.Runtime.CompilerServices;
 using NATS.Client.Core;
 using NATS.Client.JetStream.Models;
 
@@ -203,4 +204,12 @@ public interface INatsJSContext
     IAsyncEnumerable<INatsJSStream> ListStreamsAsync(
         string? subject = default,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// List stream names.
+    /// </summary>
+    /// <param name="subject">Limit the list to streams matching this subject filter.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
+    /// <returns>Async enumerable list of stream names to be used in a <c>await foreach</c> loop.</returns>
+    IAsyncEnumerable<string> ListStreamNamesAsync(string? subject = default, [EnumeratorCancellation] CancellationToken cancellationToken = default);
 }
--- a/src/NATS.Client.JetStream/INatsJSContext.cs
+++ b/src/NATS.Client.JetStream/INatsJSContext.cs
@@ -211,5 +211,5 @@ public interface INatsJSContext
     /// <param name="subject">Limit the list to streams matching this subject filter.</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
     /// <returns>Async enumerable list of stream names to be used in a <c>await foreach</c> loop.</returns>
-    IAsyncEnumerable<string> ListStreamNamesAsync(string? subject = default, [EnumeratorCancellation] CancellationToken cancellationToken = default);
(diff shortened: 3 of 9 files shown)


src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -1,7 +1,24 @@
...
+
+    event EventHandler<string>? ReconnectFailed;
+
+    event EventHandler<INatsError>? OnError;
+
+    INatsServerInfo? ServerInfo { get; }
+
+    NatsOpts Opts { get; }
+
+    NatsConnectionState ConnectionState { get; }
```
mtmk, line 17: fyi we discussed opening this up before here #127

src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -1,7 +1,24 @@
 namespace NATS.Client.Core;
 
-public interface INatsConnection
+public interface INatsConnection : IAsyncDisposable
```
mtmk, line 3: Why is it necessary to make the interface disposable? Is it because of a DI issue or to follow a general pattern of connection pooling as in database connections?

src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -1,7 +1,24 @@
...
+
+    event EventHandler<INatsError>? OnError;
+
+    INatsServerInfo? ServerInfo { get; }
+
+    NatsOpts Opts { get; }
+
+    NatsConnectionState ConnectionState { get; }
+
+    NatsHeaderParser HeaderParser { get; }
```
mtmk, line 19: This should've been internal perhaps.

src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -154,4 +171,11 @@ public interface INatsConnection
         NatsPubOpts? requestOpts = default,
         NatsSubOpts? replyOpts = default,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Connect socket and write CONNECT command to nats server.
+    /// </summary>
+    ValueTask ConnectAsync();
+
+    NatsStats GetStats();
```
mtmk, line 180: We have been quite careful about which methods we make public on the interface. Can we focus on the ones you need rather than extracting every public method into the interface?

src/NATS.Client.JetStream/INatsJSContext.cs
```diff
@@ -204,4 +204,12 @@ public interface INatsJSContext
...
         string? subject = default,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// List stream names.
+    /// </summary>
+    /// <param name="subject">Limit the list to streams matching this subject filter.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
+    /// <returns>Async enumerable list of stream names to be used in a <c>await foreach</c> loop.</returns>
+    IAsyncEnumerable<string> ListStreamNamesAsync(string? subject = default, [EnumeratorCancellation] CancellationToken cancellationToken = default);
```
mtmk, line 214: This was missed out. thanks.
Not sure if we need the annotation on the interface: `The 'EnumeratorCancellation' attribute applied to parameter 'cancellationToken' has no effect. The attribute is only effective on a parameter of type 'CancellationToken' in an async-iterator method returning 'IAsyncEnumerable<>'.`

mtmk (commented):
Looks good in general but design wise there has been a few discussions of creating interfaces to better support unit testing and I believe the consensus was to keep the public interfaces as small as possible (because we're going to have to support them forever), just want to see if that's changed. cc @caleblloyd @jasper-d @to11mtm

jasper-d:
I only skimmed through the changes, but they look good to me. I've found myself casting INatsConnection to NatsConnection many times because I needed some member that isn't on the interface.

I'm not concerned about the size of the interface. No matter if a public member is on `NatsConnection` or `INatsConnection` altering it is a breaking change anyway, so having it on the interface too doesn't really change anything I presume.

src/NATS.Client.Core/NatsPooledConnection.cs
```diff
@@ -0,0 +1,13 @@
+namespace NATS.Client.Core;
+
+internal class NatsPooledConnection : NatsConnection
```
jasper-d, line 3: This should be `sealed`. Alternatively, `NatsConnection` could be sealed and this could be a readonly struct, e.g.

```
readonly struct PooledNatsConnection : INatsConnection {
  private readonly NatsConnection _conn;
  // delegate all calls to _conn, except for DisposeAsync
}
```

src/NATS.Client.Core/INatsConnection.cs
jasper-d: Should we just make it internal now? It's a breaking change but hopefully there aren't many people relying on it (yet) and there are more breaking changes in this PR anyways.

src/NATS.Client.Core/INatsConnection.cs
caleblloyd: I think that since only the creator of the concrete class should be disposing it, that the interface probably shouldn't be disposable?

src/NATS.Client.Core/INatsConnection.cs
caleblloyd: `HeaderParser` should be internal

mtmk:
@sspates-starbucks please sign your commits going forward.

src/NATS.Client.Core/INatsConnection.cs
sspates-starbucks: @mtmk @caleblloyd Example usages:

```async using INatsConnection con = new NatsConnection()```

```async using INatsConnection con = connectionPool.GetConnection()```

File: src/NATS.Client.Core/INatsConnection.cs (hunk 4 above)
Comment (sspates-starbucks) [reply src/NATS.Client.Core/INatsConnection.cs:180]: @mtmk We probably don't need this one, but it is public on the class.   We don't know every method and property we will need yet since we can't yet use the new library due to missing interface injection. For interfaces, they should be swappable for the object instance so if the instance has the method public, it should be in the interface.

File: src/NATS.Client.Core/NatsPooledConnection.cs (hunk 6 above)
Comment (sspates-starbucks) [reply src/NATS.Client.Core/NatsPooledConnection.cs:3]: @jasper-d this can't be a struct, as it needs to inherit from NatsConnection, not the interface in order to prevent disposing of a pooled connection.

**Comment by sspates-starbucks** [comment]
@mtmk @caleblloyd @jasper-d FYI... your .editorconfig does not match the GitHub linter.  

example: https://github.com/nats-io/nats.net.v2/blob/aa32a49168cdaec31fb9bb00841bac5f79cf70e7/.editorconfig#L27C31-L27C31

File: src/NATS.Client.Core/INatsConnection.cs (hunk 2 above)
Comment (sspates-starbucks) [reply src/NATS.Client.Core/INatsConnection.cs:3]: @mtmk @caleblloyd our current v1 code does something like this:
```
public class NatsConnectionWrapper : IAsyncDisposable
{
    public INatsConnection Connection { get; }

    internal NatsConnectionWrapper(INatsConnection connection) //For Unit Testing Only
    {
        Connection = connection;
    }

    public NatsConnectionWrapper() : this(new NatsConnection()) { }

    public async ValueTask DisposeAsync() => await Connection.DisposeAsync().ConfigureAwait(false);
}
```

to11mtm:
> Looks good in general but design wise there has been a few discussions of creating interfaces to better support unit testing and I believe the consensus was to keep the public interfaces as small as possible (because we're going to have to support them forever), just want to see if that's changed. cc @caleblloyd @jasper-d @to11mtm

(I had the day off so I apologize in advance for the length of this reply)

If we are really trying to support unit testing of `NatsConnection` (Personally I prefer having basic integration tests for your impl and then using docker to run everything)... Interfaces _DO_ in the unhappy path have about two indirects... _YES_ the Jitter can optimize the call out and the ways to 'break' things are getting smaller... But again we have to support everything.

The 'yak shave' option, would be to have e.x. a _base_ `abstract` form of `NatsConnection` (i.e. `NatsConnectionBase : INatsConnection` [0]) with the methods one would want to have exposed be `abstract`, and `NatsConnection : NatsConnectionBase` still has all current functionality. Fakes can easily be entirely mocked, or wrap a 'real' `NatsConnection` for parts of the overrides if they want to. If we want to be extra nice we can have a `.Test` nuget package that provides a default 'fake' implementation for `NatsConnectionBase` that has no-ops for everything but are `virtual`. [1]

What's the benefit to all of that?

 - In general, the worst case for a virt lookup is better than the worst case for an interface lookup.
   - OTOH in general, the Jitter is better at devirting a (clean [2]) abstract base implementation.
   - We can't make `NatsConnection` sealed b/c it's breaking (even though we have no virts, it still would prevent inheriting.
 - It's a bit easier (but still requires consideration) to version methods in classes rather than interfaces.
   - e.x. `[ObsoleteAttribute("This method will be removed in the next major release", false)]` -> `[ObsoleteAttribute("This method is Deprecated. Consider OtherMethod instead", true)]`

Note that my first point is based on anecdotal experience with the NET6 Runtime/Jitter; In general, I'd expect that if the change was done, using `NatsConnectionBase` would be faster than using `INatsConnection`, but just using `NatsConnection` should perform the way it did before.

[0] - Again, we kinda have to support at least whatever `INatsConnection` already has...

[1] - The other value of such however, is it would hopefully guide users into using the base rather than interfaces for testing.

[2] - Cleaner the better. `abstract` works better than `virtual` noops/`throw new NotImplementedException()` in my experience, also ideally you have a single actual implementation (in runtime, that's why test fakes are OK) that is sealed (can't do that but oh well). It needs to stay pretty shallow though

to11mtm:
If there's one concern I have here, it's this:

(from PR Comments):

> @mtmk We probably don't need this one, but it is public on the class. We don't know every method and property we will need yet since we can't yet use the new library due to missing interface injection. For interfaces, they should be swappable for the object instance so if the instance has the method public, it should be in the interface.

This raises some concerns for me as while I can understand the need to be able to move things over to v2 from v1/AlterNats this is a bit of heavy surgery with the potential for hard-to-remove tech debt due to interface versioning challenges.

How difficult would it be to wrap `NatsConnection` and it's calls into some sort of virtual-wrapper to see what is needed to be overridden? That way we could have a better understanding of what users need for mocks and why, and make sure to provide a stable abstraction rather than risk exposing things that could prove difficult to remove in future.

sspates:
@to11mtm We do plan on wrapping the NatsConnection/INatsConnection ourselves, but we need to be able to write unit tests for our wrapper class.  That is our primary concern as the current interface implementation doesn't provide the ability to pass the interfaces into the Jetstream Context and other classes.   Using a base class makes sense if it provides that capability and probably would allow for the elements that need to be internal from a NATS library perspective to remain that way.  This PR is really just to fix a few of the areas where the interface doesn't match the real object implementation.  There are probably a dozen ways of making this work and be unit test friendly (abstracts, factories, builders, larger interfaces, virtual methods, etc.).

src/NATS.Client.Core/INatsConnection.cs
mtmk: That's fair enough I should've been open about my motive here; I think we should start using .NET `Metrics` for Open telemetry. could we make `GetSTats()` internal as well?

src/NATS.Client.Core/INatsConnection.cs
mtmk: consistency is good enough reason. I can't see a downside personally.

mtmk:
> This raises some concerns for me as while I can understand the need to be able to move things over to v2 from v1/AlterNats this is a bit of heavy surgery with the potential for hard-to-remove tech debt due to interface versioning challenges.

Just as a general note: this is a very good point. We've consciously separated ourselves from previous implementations to have freedom and the best chance to implement a modern .NET interface. we should not try to accommodate upgrades in this codebase but rather provide examples for application code if required.

mtmk:
> The 'yak shave' option, would be to have e.x. a _base_ `abstract` form of `NatsConnection` (i.e. `NatsConnectionBase : INatsConnection` [0]) with the methods one would want to have exposed be `abstract`, and `NatsConnection : NatsConnectionBase` still has all current functionality. Fakes can easily be entirely mocked, or wrap a 'real' `NatsConnection` for parts of the overrides if they want to. If we want to be extra nice we can have a `.Test` nuget package that provides a default 'fake' implementation for `NatsConnectionBase` that has no-ops for everything but are `virtual`. [1]

sounds like a good idea especially the `.Test` package. We could expose more constructs in a *test* package since it can be set to see internals as well. wouldn't want to clog this PR though.

sspates:
@mtmk Just a note, we don't have any expectations of backwards compatibility or following of the V1 conventions. If the v2 library has abstracts instead of interfaces or used virtual methods or functional injection it would all work fine for us as long as the public interface (the exposed features) match between the classes and the Dependency Inversion abstraction elements (Interface, Abstract, Func, Event, Action).

sspates:
@mtmk one other note, the connection events (e.g. ```event EventHandler<string>? ConnectionDisconnected;``) are not Async compatible and can cause performance issues on multithreaded systems because they are blocking.  You might think of switching to async handlers or an async eventing system like MediatR. 

https://medium.com/@a.lyskawa/the-hitchhiker-guide-to-asynchronous-events-in-c-e9840109fb53

mtmk:
> the connection events (e.g. `event EventHandler<string>? ConnectionDisconnected;`) are not Async compatible and can cause performance issues on multithreaded systems because they are blocking. You might think of switching to async handlers or an async eventing system like MediatR.

this is a great coincidence #299 started discussing that. Shall we yield to #299 and keep the events out of the interface for this PR in the interest of making progress?

mtmk:
I feel like we can merge this PR pending couple of changes:
* Exclude events from the interface
* make GetStats() internal

mtmk (approved):
LGTM thanks @sspates

nuzolx:
> @to11mtm We do plan on wrapping the NatsConnection/INatsConnection ourselves, but we need to be able to write unit tests for our wrapper class. That is our primary concern as the current interface implementation doesn't provide the ability to pass the interfaces into the Jetstream Context and other classes. 

After the merge and the release of version 2.1, I thought it would be possible to build a NatsJSContext with INatsConnection instead of NatsConnection, but that doesn't seem to be the case.
Have I misunderstood something?

Thank you for your work @sspates!

mtmk:
> > @to11mtm We do plan on wrapping the NatsConnection/INatsConnection ourselves, but we need to be able to write unit tests for our wrapper class. That is our primary concern as the current interface implementation doesn't provide the ability to pass the interfaces into the Jetstream Context and other classes.
> 
> After the merge and the release of version 2.1, I thought it would be possible to build a NatsJSContext with INatsConnection instead of NatsConnection, but that doesn't seem to be the case. Have I misunderstood something?
> 
> Thank you for your work @sspates!

you can use `NatsJSContextFactory`

PR #303, opened 2023-12-29
Use System.IO.Pipelines for sending

caleblloyd:
Refactors command sending to use `System.IO.Pipelines`

- Serializes full command into Pipe Writer, returning any Serialization Exceptions to `PublishAsync`
- Kicks off Flush to Pipe Reader
- At this point, the Command is considered "Enqueued for Sending", and no further Cancellation is possible.  A successful status is returned to the caller

There is a possibility that the Connection drops before the Pipe Reader is read and actually sends the bytes on the Network.  If so, the bytes remain in the Pipe Reader and will be sent if the Connection reconnects.  So the only way to ensure that the bytes were actually sent to the server is to perform a round-trip with `PingAsync`.  I believe that this behavior is similar to the Go client.

## Performance

Performance is similar to `WaitUntilSent=false` but a little better due to some reduced allocations

**Before**:

| Method             | Iter | Mean        | Error       | StdDev    | Gen0    | Allocated |
|------------------- |----- |------------:|------------:|----------:|--------:|----------:|
| **WaitUntilSentTrue**  | **64**   |    **981.8 μs** |   **945.78 μs** |  **51.84 μs** |  **1.9531** |   **6.82 KB** |
| WaitUntilSentFalse | 64   |    119.6 μs |    61.53 μs |   3.37 μs |       - |    1.1 KB |
| **WaitUntilSentTrue**  | **512**  |  **7,139.8 μs** | **6,944.54 μs** | **380.65 μs** | **15.6250** |  **52.21 KB** |
| WaitUntilSentFalse | 512  |    549.3 μs |   851.19 μs |  46.66 μs |  2.9297 |  10.43 KB |
| **WaitUntilSentTrue**  | **1024** | **16,415.3 μs** | **5,228.92 μs** | **286.61 μs** | **31.2500** | **104.26 KB** |
| WaitUntilSentFalse | 1024 |  1,015.8 μs | 1,783.37 μs |  97.75 μs | 13.6719 |  43.74 KB |

**After**

| Method       | Iter | Mean     | Error       | StdDev   | Allocated |
|------------- |----- |---------:|------------:|---------:|----------:|
| **PublishAsync** | **64**   | **115.5 μs** |    **68.81 μs** |  **3.77 μs** |     **510 B** |
| **PublishAsync** | **512**  | **439.4 μs** |   **778.86 μs** | **42.69 μs** |     **535 B** |
| **PublishAsync** | **1024** | **806.7 μs** | **1,379.06 μs** | **75.59 μs** |     **542 B** |

## Backwards Compatibility

I aimed to make the changes as Backward Compatible as possible, but there are some considerations:

- `NatsPubOpts.WaitUntilSent` and `NatsPubOpts.ErrorHandler` have been marked as Obsolete
  - Those were the only 2 options in `NatsPubOpts`
  - There still could be future options in `NatsPubOpts`, such as a per-publish `CommandTimeout` override.  So we probably shouldn't get rid of it
- `CommandTimeout` was lowered from 1 minute by default to 5 seconds.  It is used to throw a Timeout exception if anything in the path of sending the command times out, including:
  - Connection does not enter Connected State
  - Cannot acquire Semaphore required to access Pipe Writer
  - Pipe is full and cannot Flush the last command that was enqueued
- JetStream `AckOpts` were not in a format that matched `NatsPubOpts` and `NatsSubOpts` so I standardized those and got rid of `WaitUntilSent`


mtmk:
I just started a long running test with JS publisher and consumer just to make sure there are no regressions.

stebet:
It is also useful and simplifies a lot just to use a BoundedChannel as the "Semaphore" for the pipeline writer. It saves you from having to use a Semaphore and a bunch of try/finallies to synchronize the pipe flushes. So publish commands to the channel (it can have a pretty low size), and have a worker thread pull stuff off from it and send to the PipeWriter.

caleblloyd:
> It is also useful and simplifies a lot just to use a BoundedChannel instead of SemaphoreSlim to synchronize writes to the PipeWriter, since SemaphoreSlim, despite it's name, is not really a lightweight synchronization primitive

We could, but we would have to introduce another layer of memory and copying.  Right now Command Writer is writing directly to the Pipe Writer's buffers, which are the same buffers that get flushed to the Pipe Reader and sent on the wire.  Access to the Pipe Writer's buffers must be serialized.

Worse case scenario for adding a `BoundedChannel<Memory<byte>>` is another (Max Message Size * Bounded Channel Size) bytes of memory, and adding a copy to the Pipe Writer for every message.  It could end up being faster for highly parallel environments, would have to benchmark.  Maybe we save that for a potential future improvement though?

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -0,0 +1,627 @@
...
+internal sealed class CommandWriter : IAsyncDisposable
+{
+    private readonly ConnectionStatsCounter _counter;
+    private readonly TimeSpan _defaultCommandTimeout;
+    private readonly Action<PingCommand> _enqueuePing;
+    private readonly NatsOpts _opts;
+    private readonly PipeWriter _pipeWriter;
+    private readonly ProtocolWriter _protocolWriter;
+    private readonly ChannelWriter<QueuedCommand> _queuedCommandsWriter;
+    private readonly SemaphoreSlim _sem;
```
mtmk, line 32: Would something like `_semLock` be a better name indicating its purpose?

src/NATS.Client.Core/Internal/BufferWriterExtensions.cs
```diff
@@ -1,25 +1,25 @@
...
-using System.Text;
 using NATS.Client.Core.Commands;
 
 namespace NATS.Client.Core.Internal;
 
 internal static class BufferWriterExtensions
 {
-    private const int MaxIntStringLength = 10; // int.MaxValue.ToString().Length
+    private const int MaxIntStringLength = 9; // https://github.com/nats-io/nats-server/blob/28a2a1000045b79927ebf6b75eecc19c1b9f1548/server/util.go#L85C8-L85C23
+    private static readonly StandardFormat MaxIntStringFormat = new('D', MaxIntStringLength);
```
mtmk, line 11: Should we reconsider this approach? Do we gain big enough/measurable performance benefit doing this vs. writing unnecessary stuff to the wire?

mtmk (changes requested):
Looks good at first glance. Important thing is the bit about pipe reader try-finally thing. Others are nits.

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -0,0 +1,627 @@
...
+    private Task? _flushTask;
+    private bool _disposed;
+
+    public CommandWriter(NatsOpts opts, ConnectionStatsCounter counter, Action<PingCommand> enqueuePing, TimeSpan? overrideCommandTimeout = default)
+    {
+        _counter = counter;
+        _defaultCommandTimeout = overrideCommandTimeout ?? opts.CommandTimeout;
+        _enqueuePing = enqueuePing;
+        _opts = opts;
+        var pipe = new Pipe(new PipeOptions(pauseWriterThreshold: opts.WriterBufferSize, resumeWriterThreshold: opts.WriterBufferSize / 2, minimumSegmentSize: 65536, useSynchronizationContext: false));
```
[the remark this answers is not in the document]
caleblloyd: done

src/NATS.Client.Core/Internal/NatsPipeliningWriteProtocolProcessor.cs
```diff
@@ -230,6 +242,11 @@ private async Task WriteLoopAsync()
         {
             // ignore, will be handled in read loop
         }
+        catch (Exception ex)
```
mtmk, line 245: is it worth also handling some socket exceptions as well? e.g. I sometimes see `System.Net.Sockets.SocketException (10054): An existing connection was forcibly closed by the remote host.` when reconnecting. maybe a warning, since we kind of expecting it on reconnects.

caleblloyd: Meaning ignore `SocketException` similarly to how we ignore `SocketClosedException`?  I think the socket exceptions bubble up through the read loop, not the write loop

mtmk: we do get it in writer loop too sometimes:
```
2024-01-04 17:26:48.644 +00:00 fail: NATS.Client.Core.Internal.NatsPipeliningWriteProtocolProcessor[0] Unexpected error occured in write loop System.Net.Sockets.SocketException (10054): An existing connection was forcibly closed by the remote host.    at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.CreateException(SocketError error, Boolean forAsyncThrow)    at NATS.Client.Core.Internal.NatsPipeliningWriteProtocolProcessor.WriteLoopAsync() in C:\Users\mtmk\src\nats.net.v2\src\NATS.Client.Core\Internal\NatsPipeliningWriteProtocolProcessor.cs:line 174    at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)    at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.MoveNext(Thread threadPoolThread)    at System.Threading.ThreadPoolWorkQueue.Dispatch()    at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart() --- End of stack trace from previous location ---    at NATS.Client.Core.Internal.NatsPipeliningWriteProtocolProcessor.WriteLoopAsync() in C:\Users\mtmk\src\nats.net.v2\src\NATS.Client.Core\Internal\NatsPipeliningWriteProtocolProcessor.cs:line 174
```

it's quite rare. we can leave it as it is for this PR. looking at main I think this was the case already.

src/NATS.Client.Core/Internal/BufferWriterExtensions.cs
caleblloyd: Added logic in write loop to trim out the unneeded bytes

src/NATS.Client.Core/Commands/PingCommand.cs
```diff
@@ -1,68 +1,8 @@
-using NATS.Client.Core.Internal;
-
 namespace NATS.Client.Core.Commands;
 
-internal sealed class PingCommand : CommandBase<PingCommand>
+internal sealed class PingCommand
```
to11mtm, line 3: Is there value to making this a struct since it doesn't require inheritance?

src/NATS.Client.Core/Commands/ProtocolWriter.cs
```diff
@@ -48,70 +48,131 @@ public void WritePong()
...
+        {
+            // len =          header len         +' '+ payload len
+            lenSpan = span[..(MaxIntStringLength + 1 + MaxIntStringLength)];
+            span = span[lenSpan.Length..];
         }
 
-        _writer.WriteNewLine();
-    }
+        span[0] = (byte)'\r';
+        span[1] = (byte)'\n';
```
to11mtm, line 122: If we do `Int` for others, maybe `short` here?

src/NATS.Client.Core/Commands/ProtocolWriter.cs
```diff
@@ -48,70 +48,131 @@ public void WritePong()
...
 
         // Consider null as empty payload. This way we are able to transmit null values as sentinels.
         // Another point is serializer behaviour. For instance JSON serializer seems to serialize null
         // as a string "null", others might throw exception.
         if (value != null)
         {
-            serializer.Serialize(_bufferPayload, value);
+            var initialCount = _writer.UnflushedBytes;
+            serializer.Serialize(_writer, value);
+            totalLength += _writer.UnflushedBytes - initialCount;
```
to11mtm, line 140: Pardon my complete ignorance here (haven't pulled the branch down) but is there any chance `UnflushedBytes` could be racy?

src/NATS.Client.Core/Commands/ProtocolWriter.cs
```diff
@@ -48,70 +48,131 @@ public void WritePong()
 
     // https://docs.nats.io/reference/reference-protocols/nats-protocol#pub
     // PUB <subject> [reply-to] <#bytes>\r\n[payload]\r\n
-    public void WritePublish(string subject, string? replyTo, NatsHeaders? headers, ReadOnlySequence<byte> payload)
+    // or
+    // https://docs.nats.io/reference/reference-protocols/nats-protocol#hpub
+    // HPUB <subject> [reply-to] <#header bytes> <#total bytes>\r\n[headers]\r\n\r\n[payload]\r\n
+    //
+    // returns the number of bytes that should be skipped when writing to the wire
+    public int WritePublish<T>(string subject, T? value, NatsHeaders? headers, string? replyTo, INatsSerialize<T> serializer)
```
to11mtm, line 56: Have mixed feelings about this, would need to check what else has changed.

Main reason being, it looks like we've lowered a generic call further down (i.e. this is now `<T>` rather than `ReadOnlySequence<byte>`) and that may have implications around codegen for struct types.

src/NATS.Client.Core/Commands/ProtocolWriter.cs
```diff
@@ -48,70 +48,131 @@ public void WritePong()
...
+        span[1] = (byte)'\n';
+        _writer.Advance(2);
+
+        // write the length
+        var lenWritten = 0;
+        if (headers != null)
+        {
+            if (!Utf8Formatter.TryFormat(headersLength, lenSpan, out lenWritten))
+            {
+                throw new NatsException("Can not format integer.");
```
to11mtm, line 154: NIT/Future: Consider wrapping throw in call to encourage JITter.

src/NATS.Client.Core/Commands/PingCommand.cs
```diff
@@ -1,68 +1,8 @@
...
-        writer.WritePing();
-    }
+    public DateTimeOffset WriteTime { get; } = DateTimeOffset.UtcNow;
 
-    protected override void Reset()
-    {
-        WriteTime = null;
-        _connection = null;
-    }
+    public TaskCompletionSource<TimeSpan> TaskCompletionSource { get; } = new(TaskCreationOptions.RunContinuationsAsynchronously);
```
to11mtm, line 7: If we do this It would be nice to have a github issue to try to pool in future (since I'd assume ping is easy to do that with)

to11mtm (commented):
Still need to look at some things closer but wanted to leave what feedback I have

to11mtm:
I'm still reviewing but wanted to try to add some flavor to all of this;

> We could, but we would have to introduce another layer of memory and copying. Right now Command Writer is writing directly to the Pipe Writer's buffers, which are the same buffers that get flushed to the Pipe Reader and sent on the wire. Access to the Pipe Writer's buffers must be serialized.

At this point I probably -should- pull down the PR and look, How does SemaphoreSlim help here? Main reason I ask is that it does not, AFAIK, have FIFO behavior as a guarantee.

> Worse case scenario for adding a `BoundedChannel<Memory<byte>>` is another (Max Message Size * Bounded Channel Size) bytes of memory, and adding a copy to the Pipe Writer for every message. It could end up being faster for highly parallel environments, would have to benchmark. Maybe we save that for a potential future improvement though?

It's possible that it was being suggested, rather than use a `BoundedChannel<Memory<byte>>`... but instead to use `Channel<T>` (in the past I've done `Channel<Exception>` with sentinels or `Channel<int>` with numbers for state... [0]) with  in a specific way to mimic a `SemaphoreSlim`.

Getting more specific, as a semi-contrived example, `SubscribeAsync` now does a `TryWait`, if that fails it then dumps into the state machine. but I'd have to guess that the `TryWait`->`WaitAsync` will have more interlocked/volatile operations than a `ReadAsync` or `WaitToReadAsync` on a `ChannelReader`. [1]

[0] - In fact, I have done so on more than one occasion to 'gate' `NatsConnection` and various things on disconnect/reconnect.

[1] - I've seen this myself, where `WaitToReadAsync` on a `BoundedChannel` worked better than `SemaphoreSlim`.

src/NATS.Client.Core/Commands/ProtocolWriter.cs
caleblloyd: I don't think so, since access to the writer is serialized.  And we ensure that the previous Flush has completed every time prior to writing a new comand

src/NATS.Client.Core/Commands/PingCommand.cs
caleblloyd: Absolutely, if this merges we should make a follow-up to pool this in the future

src/NATS.Client.Core/Commands/ProtocolWriter.cs
caleblloyd: Not familiar with that- can you paste a quick example?

src/NATS.Client.Core/Commands/ProtocolWriter.cs
```diff
@@ -48,70 +49,95 @@ public void WritePong()
...
-            var headersLength = _bufferHeaders.WrittenSpan.Length;
-            _writer.WriteNumber(CommandConstants.NatsHeaders10NewLine.Length + headersLength);
-            _writer.WriteSpace();
-            var total = CommandConstants.NatsHeaders10NewLine.Length + headersLength + payload.Length;
-            _writer.WriteNumber(total);
+            span[0] = (byte)'H';
+            span[1] = (byte)'P';
+            span[2] = (byte)'U';
+            span[3] = (byte)'B';
+            span[4] = (byte)' ';
```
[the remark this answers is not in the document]
caleblloyd: I think that `BinaryPrimitives.WriteUInt32LittleEndian` has a slight edge over setting index directly for 4 bytes - but probably not a big enough difference to sacrifice the readability.  `span.CopyTo` is the slowest.  see [gist](https://gist.github.com/caleblloyd/bcfcbdee77757c54c00a226082debb4b)

src/NATS.Client.Core/Commands/ProtocolWriter.cs
caleblloyd: Setting the index directly is actually faster for 2 bytes than `BinaryPrimitives.WriteUInt16LittleEndian`, see [gist](https://gist.github.com/caleblloyd/bcfcbdee77757c54c00a226082debb4b)

mtmk (approved):
LGTM

src/NATS.Client.Core/Commands/PingCommand.cs
mtmk: see #321

src/NATS.Client.Core/Commands/PingCommand.cs
caleblloyd: :+1: done

mtmk (approved):

PR #304, opened 2024-01-02
Explicitly set service listener serialization

mtmk:
This is necessary when the default serializer isn't using a Raw serialization for NatsMemory type which is used by the service listener.


caleblloyd (approved):
LGTM

PR #310, opened 2024-01-05
Release 2.0.3

mtmk:
Bug fix release

* Svc info subs do not use queue group (#308)
* Explicitly set service listener serialization (#304)
* KV get keys should not hang with deleted keys (#293)
* NatsMemoryOwner empty payload should not throw (#295)
* Fixed serialization problem in PublishModel/PublishHeaders examples (#296)


caleblloyd (approved):
LGTM

Issue #312, opened 2024-01-06
Consider Exposing ReplyTo for NatsJSMsg


rickdotnet:
### Proposed change

NatsMsg and NatsSvcMsg both expose the ReplyTo property. For consistency (and selfishly, my use-case) I'd like to see this property exposed for NatsJSMsg as well.

https://github.com/nats-io/nats.net.v2/blob/1f7010c7422e049faea6417031a9ea29eb7e9e54/src/NATS.Client.JetStream/NatsJSMsg.cs#L179

### Use case

I rely on the ReplyTo subject instead of the Reply method for NatsCore messages. This property would allow me reply to JetStream messages as well.

### Contribution

This looks to be a two line adjustment, but I don't mind so long as someone tells me the desired approach. For instance, I would probably add this to the interface as well, but I am not involved enough to know if there are design concerns/contracts that aren't obvious.

mtmk:
note that for example Go client already had this exposed.

PR #313, opened 2024-01-06
Expose reply-to for JS msg

mtmk:
This is to have parity with other clients


mtmk:
apologies I missed the issue #312 I shall yield to @rickdotnet for implementation. sorry about that.

PR #324, opened 2024-01-11
Async event handlers

mnmr:
Changed events to use AsyncEventHandler. Modified all event subscribers internally. Changed event publishing in NatsConnection to instead write to a channel and used a background task to do the actual publishing.

All the tests (except the memory tests, which I didn't run) run successfully, however, it should be noted that events now fire outside locks. Which is probably a good thing as NatsConnection was formerly relying on event handlers to be well-behaved.


src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -594,6 +613,42 @@ private async void ReconnectLoop()
...
     }
 
+    private async Task PublishEventsAsync()
+    {
+        try
+        {
+            while (!_disposedCancellationTokenSource.IsCancellationRequested)
+            {
+                await _eventChannel.Reader.WaitToReadAsync().ConfigureAwait(false);
+                if (_eventChannel.Reader.TryRead(out var eventData))
```
mtmk, line 623: Should we not check return value of `WaitToReadAsync()` and use while loop for `TryRead()`?

mtmk (commented):
Look good 👍 Just a quick question about channel reader.

mtmk:
there were a few test failures due to the way we pull in nats-server binaries (nothing to do with actual tests). Rerunning some of the m now...

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -84,16 +94,25 @@ public NatsConnection(NatsOpts opts)
...
             AllowSynchronousContinuations = false,
         };
+
+        // push consumer events to a channel so handlers can be awaited (also prevents user code from blocking us)
+        _eventChannel = Channel.CreateUnbounded<(NatsEvent, object)>(new UnboundedChannelOptions
+        {
+            AllowSynchronousContinuations = false,
+            SingleWriter = false,
+            SingleReader = true,
+        });
```
to11mtm, line 104: I'm actually super happy to see this, as in the past it was a hurdle for some heavy consumer cases.

to11mtm (commented):
I think this is a good change overall.

Two thoughts/concerns: 

First, `Task` rather than `ValueTask` from the standpoint of allowing high perf scenarios without breaking APIs in future. We may want to think about this. (although....)

Secondly, whether we should consider instead doing something along the lines of (genuinely apologize for late night napkin concept but wanted to have the thought out there, both for the sake of avoiding breaking things for current users as well as making it potentially easier for consumers of all types to wire up):

```
//again, forgive (probably some) incorrectness

//TODO: Need to allow for ordering of execution.
//NOTE: Due to ordering, may need an ordered dictionary with locks 
//           rather than ConcurrentDictionary
private readonly ConcurrentDictionary<object, Func<object, string,ValueTask> _disconnectHandlers = new();

    //Assumes AsyncEventHandler is now ValueTask delegate type.
    public event AsyncEventHandler<string>? ConnectionDisconnectedAsync
    {
        add { _disconnectHandlers.TryAdd(value, value); }
        remove { _disconnectHandlers.TryRemove(value, out _); }
    }
    public event EventHandler<string>? ConnectionDisconnected
    {
        add { _disconnectHandlers.TryAdd(value, 
                 v=> {
                            return  (send,arg) => 
                                    { 
                                        v(send,arg); 
                                        return ValueTask.CompletedTask}; 
                                    }
                         }
        remove 
        { 
            _disconnectHandlers.TryRemove(value, out _); 
        }    
    }
    //Could also have Task variant that converts to ValueTask
```

Again, I'm not certain whether this would have a perf benefit [0] but at least it would help avoid breaking changes of various sorts...

Not sure if doing such would break things like outside callers trying to raise but in general that is a questionable pattern.


[0] - It very well might, between avoiding the casts as well as the cost of [`.GetInvocationList()`'s internal cloning of the list](https://github.com/dotnet/runtime/blob/22ba7d607bb1d9caa0db9afcdc47eb5cef641fcb/src/coreclr/System.Private.CoreLib/src/System/MulticastDelegate.cs#L416-L434), as well as the ability to get rid of casts in the main event dispatch loop.

src/NATS.Client.Core/AsyncEventHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace NATS.Client.Core;
+
+/// <summary>
+/// An asynchronous event handler.
+/// </summary>
+/// <param name="sender">The sender of the event.</param>
+/// <param name="args">Event arguments.</param>
+/// <returns>A task whose completion signals handling is finished.</returns>
+public delegate Task AsyncEventHandler(object? sender, EventArgs args);
```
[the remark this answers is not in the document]
mtmk: I don't see a downside personally. We might as well start with `ValueTask` to avoid breaking it later.

src/NATS.Client.Core/NatsConnection.cs
mtmk: Good point a short comment would be nice.

mtmk (commented):
I think @to11mtm's point about supporting both event handler types is well worth considering. We won't break and invocation list performance would be a nice bonus.

src/NATS.Client.Core/NatsConnection.cs
mnmr: There is technically already a while loop, and since this is only for event notifications I don't think it's worth adding an inner while loop to empty the queue. But using the return value makes sense, I will update the PR.

src/NATS.Client.Core/AsyncEventHandler.cs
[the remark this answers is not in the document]
mnmr: I think using ValueTask is a mistake. This code is not on a hot path (if we had a MessageDelivered event, this would be a different story), and the awaiting code is outside the library - you must never await a ValueTask more than once, and this provides a gotcha for consumers of the library.

I'm not opposed to making it ValueTask per se (and will happily adjust the PR), but don't see this as an area that needs any optimization. There's only 4 events and none of them are expected to fire very often; if they do, you have more serious concerns to deal with.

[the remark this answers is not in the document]
mtmk: Good point about not being on hot path. In general we do use ValueTask pretty much everywhere though, hot path or not. If you don't see any technical problems using it (apart from double await) do you think we should consider using it for consistency as well?

mnmr:
> I think @to11mtm's point about supporting both event handler types is well worth considering. We won't break and invocation list performance would be a nice bonus.

It's a clever idea, but not without drawbacks (e.g. how to provide ordering). ConcurrentDictionary is slow to iterate and using a dictionary for something that is likely 1-3 entries might also not be ideal. If we really want to go down this route, I think it should be possible to use Interlocked and a fixed sized array to store the delegates (i.e. there would be a max subscriber limit).

That said, as I understand it, these events will be fired quire rarely (perhaps with the exception of OnError?), and since it's now being done using a separate task, it's not performance critical at all. Also, since this code calls out to user-provided code, it can be as clever as it want and still not provide any guarantees about timeliness (since an earlier subscriber could be taking a long time to complete; this would of course be bad, but is also not something the library can protect against).

I could of course be mistaken about this, or you could have plans for events that could be more frequent than the current set, but if not then I don't think we need cleverness here at all.

mnmr:
As an aside, I think we should consider introducing custom EventArgs for all events, and in particular for the 3 events that currently just pass a string.

Motivation: If we at some point decide to add a timestamp (since the handlers are invoked some time after the occurrence, this might be useful) or a concurrency version/token (to detect if an event relates to an earlier state of the connection), then these fields can simply be added to the EventArgs - no breaking change for existing subscribers as the signature would still be the same.

src/NATS.Client.Core/AsyncEventHandler.cs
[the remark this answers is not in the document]
mnmr: I would use ValueTask sparingly, unless it was internally or on a hot path. Stephen Toub has a great [guide](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/) on the topic.

mtmk:
> I think we should consider introducing custom EventArgs for all events, and in particular for the 3 events that currently just pass a string.

This sounds good to me. passing strings here is a bit odd.

mnmr:
> This sounds good to me. passing strings here is a bit odd.

If there are no objections from anyone else, I will apply this change to the PR.

Let me know what the decision ends up being for ValueTask/Task and "optimized storage with dual sync/async support". For now, I will not be applying any of these changes.

mnmr:
I've pulled the latest code from main into the branch, and applied some changes:

- Added NatsEventArgs and renamed MessageDroppedError to NatsMessageDroppedEventArgs
- Changed events to use EventArgs and adjusted internal subscribers
- Renamed OnError to MessageDropped, as that better reflects what error is signaled and also is naming-wise more consistent with the other events

The error event was only signaled in one place, so it makes sense to rename it I think, but this was not discussed earlier so feel free to shout and have me revert if you dislike it :)

mtmk:
> The error event was only signaled in one place, so it makes sense to rename it I think, but this was not discussed earlier so feel free to shout and have me revert if you dislike it :)

nah it makes sense but I think we need to keep it as it is not because of matter of taste :) but I think we should keep it inline with other NATS clients: https://docs.nats.io/using-nats/developer/connecting/events/slow#detect-a-slow-consumer-and-check-for-dropped-messages

mnmr:
> > The error event was only signaled in one place, so it makes sense to rename it I think, but this was not discussed earlier so feel free to shout and have me revert if you dislike it :)
> 
> nah it makes sense but I think we need to keep it as it is not because of matter of taste :) but I think we should keep it inline with other NATS clients: https://docs.nats.io/using-nats/developer/connecting/events/slow#detect-a-slow-consumer-and-check-for-dropped-messages

But there's not really much consensus between the SDKs when it comes to error handling. The Java version supplies a "slowConsumerDetected" handler to the errorListener, but that's both backwards and unsafe (a special-purpose method is used to handle a generic event, rather than vice versa). And correcting a mistake has to start somewhere, right? ;)

Alternatively, we could keep both. That is, if we re-introduce Error but with a generic NatsErrorEventArgs argument (with an enum to specify the kind of error), we have an event in place for any future errors. Handlers for the generic Error event would need to inspect the args and cast it to the specific args type if they want more information.

I will of course change it back to Error (even OnError, although the OnXxx methods are usually reserved for the methods triggering the event) if you insist, but I think a bit of push-back here is in order ;)

mnmr:
I see that some tests are failing in CI, but they run fine locally. Is this something to be worried about, or is this a known issue with how the tests are executed?

mtmk:
> I see that some tests are failing in CI, but they run fine locally. Is this something to be worried about, or is this a known issue with how the tests are executed?

There are a few flappers. I need to fix them at some stage. Just re-running them now.

mtmk:
> I will of course change it back to Error (even OnError, although the OnXxx methods are usually reserved for the methods triggering the event) if you insist, but I think a bit of push-back here is in order ;)

Nothing like a good debate :) If you look at [the examples](https://docs.nats.io/using-nats/developer/connecting/events/slow#detect-a-slow-consumer-and-check-for-dropped-messages) all languages are accepting some form of error callback thorough a generic error handler then deducing (which would be an `is` type check for us) the nature of the error in that callback (Java is a bit odd but same idea applies).

Good point about the name though? Just `Error` or `ErrorReceived`?

mnmr:
> Nothing like a good debate :) If you look at [the examples](https://docs.nats.io/using-nats/developer/connecting/events/slow#detect-a-slow-consumer-and-check-for-dropped-messages) all languages are accepting some form of error callback thorough a generic error handler then deducing (which would be an `is` type check for us) the nature of the error in that callback (Java is a bit odd but same idea applies).

Don't most of these exist in their present form because they originate from designs in languages that are less expressive? If you can only pass one error handler, it needs to be unspecific. This is not consumer friendly, as you need to know what types you'll encounter at runtime and then do appropriate casting. It doesn't feel like a good fit for a language where each event and handler type can be declared separately.

> Good point about the name though? Just `Error` or `ErrorReceived`?

If we do go back to this we should probably use the verbified edition, so ErrorReceived/ErrorOccurred/etc

mtmk:
> > Nothing like a good debate :) If you look at [the examples](https://docs.nats.io/using-nats/developer/connecting/events/slow#detect-a-slow-consumer-and-check-for-dropped-messages) all languages are accepting some form of error callback thorough a generic error handler then deducing (which would be an `is` type check for us) the nature of the error in that callback (Java is a bit odd but same idea applies).
> 
> Don't most of these exist in their present form because they originate from designs in languages that are less expressive? If you can only pass one error handler, it needs to be unspecific. This is not consumer friendly, as you need to know what types you'll encounter at runtime and then do appropriate casting. It doesn't feel like a good fit for a language where each event and handler type can be declared separately.

That's a good point. This one is kind of a gray area maybe, but our general approach is that we should be using what's natural to C# over general NATS client parity.

mtmk:
LGTM thanks @mnmr 💯 

I shall merge in a day or so to give everyone a chance to have one last look.

cc @sspates @to11mtm @oising @caleblloyd 

PS I'll create a discussion for keeping the old handlers (@to11mtm's suggestion) once merged.

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -594,6 +612,42 @@ private async void ReconnectLoop()
         }
     }
 
+    private async Task PublishEventsAsync()
+    {
+        try
+        {
+            while (!_disposedCancellationTokenSource.IsCancellationRequested)
+            {
+                var hasData = await _eventChannel.Reader.WaitToReadAsync(_disposedCancellationTokenSource.Token).ConfigureAwait(false);
```
caleblloyd, line 621: What's the benefit of serializing events over a channel and awaiting them, as opposed to just firing them off, wrapping them in a try/catch with a logger

I guess if there were a large amount of events we could be creating a lot of demand on the thread pool, but I don't think that is all too common?

mnmr: They're not serialized, just passed through a queue. One advantage of this intermediate step is that event delivery will be in-order, which cannot be guaranteed for fire-and-forget tasks that may run in random order (this could be mitigated by giving events a sequence number and/or timestamp, but recipients would then need to re-order events). It also allows for future pre-processing, e.g. to throttle error events, without blocking the connection.

caleblloyd: Probably a poor use of the word `serializing` - I meant what is the benefit of running one-at-a-time, awaiting each one.  In-order delivery certainly is an advantage, but this solution could suffer from head-of-line blocking

mtmk: Related note: One of the reasons for using a channel to deliver events is to be able to fire synchronously using `TryWrite` e.g. from subscriber's channel full `Action`.

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -594,6 +612,42 @@ private async void ReconnectLoop()
...
+                if (hasData && _eventChannel.Reader.TryRead(out var eventArgs))
```
to11mtm, line 622: Apologies for not noticing earlier; this probably could be a `while` to limit thrashing on `WaitToReadAsync` calls.

mnmr: It's fixed.

src/NATS.Client.Core/AsyncEventHandler.cs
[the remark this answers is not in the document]
caleblloyd: I wanted to revive this thread - I just scanned our public API and I think we use `ValueTask` everywhere.  It might be best to stick to that for consistency's sake.  External code wouldn't be awaiting events anyways, we control when the events run and are awaited.

src/NATS.Client.Core/NatsConnection.cs
mnmr: > In-order delivery certainly is an advantage, but this solution could suffer from head-of-line blocking

I suspect that slow-running handler code would have been much worse with the old code, as at least some were fired while locks were held. 

There are some things we could consider to perhaps improve the situation:
- assign a sequence number to events and fire them off asynchronously (but this needs task tracking)
- use a bounded channel with some discard strategy to avoid flooding consumers with events (but this may cause lost events for slow handlers)
- deduplication for queued events, i.e. only publish the last of a given type (not entirely trivial)
- throttling for error events (min-interval between reports, possibly with scaled delays; also not trivial)

However, I got the impression that we would move forward with the current solution and then iterate on it after discussing options in a follow-up thread (mostly for considering the benefits of whether to manage the delegates manually, so that we could support both sync and async event syntax, and same some allocations). Maybe we need to have this discussion before this PR is merged?

caleblloyd: I'm fine to move forward with the current solution, I think how the events are invoked can be changed in a non-breaking fashion.

caleblloyd:
I think it's almost ready - I did revive the `ValueTask` comment thread - https://github.com/nats-io/nats.net.v2/pull/324#discussion_r1456721464 - but that's the only outstanding comment I have

caleblloyd:
I think that we should stick with `ValueTask`, since the public API is currently consistent in using `ValueTask`.  @mtmk what do you think?

mtmk:
Let's change that to `ValueTask`. That was my opinion too. Also, I think that's the consensus as well.

@mnmr I think your main point against using `ValueTask` is that care needs to be taken because of its pitfalls like not being able to double await; which is valid but for our library. I think we're are already making extensive use of `ValueTask`; so from our library's design perspective `ValueTask` fits in better because it's similar to other API calls as @caleblloyd mentioned, also it might open up opportunities for finer optimizations in the future as @to11mtm suggested.

mnmr:
I have changed the delegate to return ValueTask and adjusted the rest of the code.

I also investigated if we could get rid of the GetInvocationList() copy (other than by caching the result), and came across this interesting [issue](https://github.com/dotnet/runtime/issues/41849) - so it looks like this is something that can be improved in the future.

mtmk:
Added open comments to #348 (as far as I could gather, feel free to add/correct bits I missed)

mtmk (approved):
LGTM thanks @mnmr

mtmk:
Apologies, I forgot to check and remind you for this PR @mnmr please sign your commits next time.

PR #314, opened 2024-01-06
Expose ReplyTo on NatsJSMsg

rickdotnet:
This PR addresses #312 , exposing the ReplyTo property on NatsJSMsg.

PR #326, opened 2024-01-11
Release 2.1.0-preview.1

mtmk:
* Use System.IO.Pipelines for sending (#303)
* Suggested fixes to interfaces (#297)
* JS Message reply-to exposed (#314)
* NatsSub cancellation token race fix


caleblloyd (approved):
LGTM

PR #327, opened 2024-01-12
Object store name check fix

mtmk:
Since object names are encoded when stores they can be anything as long as not empty.


caleblloyd (approved):
LGTM

PR #332, opened 2024-01-14
Nullability review for models

mnmr:
Made optional reference properties nullable unless a meaningful default value is  provided. 

Removed redundant default assignments and cleaned up some code generated namespace prefixes.

Marked required properties (derived from attribute annotations and code inspection). Any use of "required" is in a conditional code block, as the JSON source generator for 6.0 doesn't understand the keyword and thus produces invalid code. Maybe this can be solved by using the code generated for 8.0 also for the 6.0 build.

Adjusted code in a few places where required properties had to be set and some nullable properties had to be checked before use.
 
Excluded SC1206 warnings from editorconfig (couldn't manage to use the required keywork with it enabled). This is likely fixable but apparently beyond my skills :)

As the model signatures have changed, this PR is a breaking change.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -149,7 +149,7 @@ dotnet_diagnostic.SA1202.severity = warning  # Elements should be ordered by acc
 dotnet_diagnostic.SA1203.severity = warning  # Constants should appear before fields
 dotnet_diagnostic.SA1204.severity = warning  # Static elements should appear before instance elements
 dotnet_diagnostic.SA1205.severity = warning  # Partial elements should declare access
-dotnet_diagnostic.SA1206.severity = warning  # Declaration keywords should follow order
+dotnet_diagnostic.SA1206.severity = none     # Declaration keywords should follow order
 dotnet_diagnostic.SA1207.severity = warning  # Protected should come before internal
 dotnet_diagnostic.SA1208.severity = warning  # System using directives should be placed before other using directives
 dotnet_diagnostic.SA1209.severity = warning  # Using alias directives should be placed after other using directives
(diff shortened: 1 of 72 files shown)


src/NATS.Client.JetStream/NatsJSContext.Streams.cs
```diff
@@ -17,6 +17,7 @@ public partial class NatsJSContext
         StreamConfig config,
         CancellationToken cancellationToken = default)
     {
+        ArgumentNullException.ThrowIfNull(nameof(config.Name));
```
mtmk, line 20: missing argument.

.editorconfig
```diff
@@ -149,7 +149,7 @@ dotnet_diagnostic.SA1202.severity = warning  # Elements should be ordered by acc
 dotnet_diagnostic.SA1203.severity = warning  # Constants should appear before fields
 dotnet_diagnostic.SA1204.severity = warning  # Static elements should appear before instance elements
 dotnet_diagnostic.SA1205.severity = warning  # Partial elements should declare access
-dotnet_diagnostic.SA1206.severity = warning  # Declaration keywords should follow order
+dotnet_diagnostic.SA1206.severity = none     # Declaration keywords should follow order
```
mtmk, line 152: a lot of change already. fair enough :) I will spend some time on it on a follow up PR (probably just add the disable pragmas wherever it's not happy)

mtmk (commented):
LGTM just a minor fix. thanks @mnmr

src/NATS.Client.JetStream/NatsJSContext.Streams.cs
mnmr: Good catch. I had two other places with the same mistake! .. at least I'm consistently wrong ;)

.editorconfig
mnmr: I spent a fair bit of time trying to get it to work, but I don't think it is possible as long as 6.0 is also a target.

mtmk (approved):

PR #339, opened 2024-01-17
TLS auth timeout

### LINKED ISSUE #337: ConnectTimeout does not cancel upgrade of TcpConnection to SslConnection
### Discussed in https://github.com/nats-io/nats.net.v2/discussions/335

<sup>Originally posted by **joao-genebra** January 16, 2024</sup>
When opening a connection, after the connection has been created and before it is upgraded to ssl [here](https://github.com/nats-io/nats.net.v2/blob/3e3c86745008cf190e6f16d83a08a6a839adfda4/src/NATS.Client.Core/NatsConnection.cs#L288C21-L294C22), if some kind of hiccup occurs in our network, the timeout value set in `ConnectTimeout` of `NatsOpts` is not stopping the connection upgrade, which is causing the call to `ConnectAsync` to stay blocked.

We know that the connection has already been opened because we can find [this](https://github.com/nats-io/nats.net.v2/blob/3e3c86745008cf190e6f16d83a08a6a839adfda4/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs#L392) log.</div>

mtmk:
Resolves #337

### DIFF
--- a/src/NATS.Client.Core/Internal/SslStreamConnection.cs
+++ b/src/NATS.Client.Core/Internal/SslStreamConnection.cs
@@ -85,29 +85,12 @@ internal sealed class SslStreamConnection : ISocketConnection
         var options = SslClientAuthenticationOptions(uri);
         try
         {
-            using var ctsTimer = new CancellationTokenSource();
-            using var ctsAuth = new CancellationTokenSource();
-
-            var taskSsl = _sslStream.AuthenticateAsClientAsync(options, ctsAuth.Token);
-            var taskTimeout = Task.Delay(timeout, ctsTimer.Token);
-
-            var task = await Task.WhenAny(taskSsl, taskTimeout).ConfigureAwait(false);
-            if (task == taskTimeout)
-            {
-#if NET6_0
-                ctsAuth.Cancel();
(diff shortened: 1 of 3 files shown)


src/NATS.Client.Core/Internal/SslStreamConnection.cs
```diff
@@ -80,12 +80,34 @@ public void SignalDisconnected(Exception exception)
...
 
-    public async Task AuthenticateAsClientAsync(NatsUri uri)
+    public async Task AuthenticateAsClientAsync(NatsUri uri, TimeSpan timeout)
     {
         var options = SslClientAuthenticationOptions(uri);
         try
         {
-            await _sslStream.AuthenticateAsClientAsync(options).ConfigureAwait(false);
+            using var ctsTimer = new CancellationTokenSource();
+            using var ctsAuth = new CancellationTokenSource();
```
caleblloyd, line 89: why not use `ctsAuth = new CancellationTokenSource(timeout);` and avoid the 2nd task?

caleblloyd (approved):
LGTM

PR #343, opened 2024-01-19
Log connection failure as warning

### LINKED ISSUE #342: Log connection failures as warnings
### Proposed change

Log connection failures in `ReconnectLoop` as warning instead of error.
https://github.com/nats-io/nats.net.v2/blob/199c2426d6a3385788d9c6a68ad74af6c004b942/src/NATS.Client.Core/NatsConnection.cs#L567

### Use case

Since nats-server instances can come and go, this leads to quite a few errors being logged (and thus noise), i.e. when redeploying nats-server.

### Contribution

Yes

jasper-d:
Fixes #342

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -564,7 +564,7 @@ public partial class NatsConnection : INatsConnection
             {
                 if (url != null)
                 {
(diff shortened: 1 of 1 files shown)


mtmk:
Failed tests are flappers and tracked in #344

PR #346, opened 2024-01-22
Send buffer changes

mtmk:
This PR
| Method       | Iter | Mean     | Error     | StdDev   | Allocated |
|------------- |----- |---------:|----------:|---------:|----------:|
| PublishAsync | 64   | 156.4 us |  20.34 us |  1.12 us |     488 B |
| PublishAsync | 512  | 299.8 us | 146.19 us |  8.01 us |     520 B |
| PublishAsync | 1024 | 482.9 us | 218.91 us | 12.00 us |     488 B |

Main
| Method       | Iter | Mean     | Error     | StdDev  | Allocated |
|------------- |----- |---------:|----------:|--------:|----------:|
| PublishAsync | 64   | 165.9 us |  39.41 us | 2.16 us |     488 B |
| PublishAsync | 512  | 375.1 us |  81.65 us | 4.48 us |     488 B |
| PublishAsync | 1024 | 607.0 us | 152.22 us | 8.34 us |     488 B |


src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -70,589 +91,522 @@ public async ValueTask DisposeAsync()
...
-            }
-            finally
+            PipeWriter bw;
+            lock (_lock)
             {
-                EnqueueCommand(success);
+                if (_pipeWriter == null)
+                    ThrowOnDisconnected();
+                bw = _pipeWriter!;
             }
```
to11mtm: Given the re-use of this pattern, is there any potential value in us trying to consolidate shared logic of the lock/get (and possibly more) into a method call to lower maintenance concerns? (Maybe not at this stage but though it was worth asking.)

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -70,589 +91,522 @@ public async ValueTask DisposeAsync()
...
-            return UnsubscribeStateMachineAsync(true, sid, cancellationToken);
-        }
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static void ThrowOnDisconnected() => throw new NatsException("Connection hasn't been established yet.");
 
+    [AsyncMethodBuilder(typeof(PoolingAsyncValueTaskMethodBuilder))]
+    private async ValueTask PublishLockedAsync(string subject, string? replyTo,  NatsPooledBufferWriter<byte> payloadBuffer, NatsPooledBufferWriter<byte>? headersBuffer, CancellationToken cancellationToken)
+    {
+        // Interlocked.Add(ref _counter.PendingMessages, 1);
+        await _semLock.WaitAsync(cancellationToken).ConfigureAwait(false);
```
to11mtm, line 274: Speaking of, putting the `WaitAsync` is probably 'better' here overall (the old `Wait(0)` -> statemachine pattern has some thrashing concerns) but it does raise some interesting questions in my head again as to the driver.


(Below is a language-lawyer disclaimer and... well, hopefully we aren't worried about it, but if we are, wanted to bring it up.)
This is going to be more fair than the old method (b/c it's always `WaitAsync`, but it is worth noting that `SemaphoreSlim` like most locking primitives in .NET do not guarantee fairness as to which thread gets released first [citation](https://github.com/dotnet/runtime/issues/19336#issuecomment-259982210). 

If we -do- need stronger ordering/fairness on these, we will need to consider an alternative... and also remember the caveats mentioned in the citation; they are certainly not 'absolutes' however they may require consideration in a solution (again, only if needed, this can be a rabbit hole.)

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -70,589 +91,522 @@ public async ValueTask DisposeAsync()
...
-        {
-            if (!await _semLock.WaitAsync(_defaultCommandTimeout, cancellationToken).ConfigureAwait(false))
-            {
-                throw new TimeoutException();
-            }
-        }
+        // 8520 should fit into 6 packets on 1500 MTU TLS connection or 1 packet on 9000 MTU TLS connection
+        // assuming 40 bytes TCP overhead + 40 bytes TLS overhead per packet
+        const int maxSendMemoryLength = 8520;
+        var sendMemory = new Memory<byte>(new byte[maxSendMemoryLength]);
```
to11mtm, line 318: Question: Does this get re-started between connections?
And if yes is there a way to possibly reuse?

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -70,589 +91,522 @@ public async ValueTask DisposeAsync()
...
     {
-        if (Interlocked.Increment(ref _disposed) == 1)
+        var minimumSize = (uint)_index + (uint)sizeHint;
+
+        // The ArrayPool<T> class has a maximum threshold of 1024 * 1024 for the maximum length of
+        // pooled arrays, and once this is exceeded it will just allocate a new array every time
+        // of exactly the requested size. In that case, we manually round up the requested size to
+        // the nearest power of two, to ensure that repeated consecutive writes when the array in
+        // use is bigger than that threshold don't end up causing a resize every single time.
+        if (minimumSize > 1024 * 1024)
```
to11mtm, line 599: <3 the exposition here! Great breakdown of the why.

to11mtm (commented):
Left some first-pass comments/questions, will try to review further and pull down for a look.

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -70,589 +91,522 @@ public async ValueTask DisposeAsync()
...
-        {
-            return ConnectStateMachineAsync(false, connectOpts, cancellationToken);
-        }
-
-        if (_flushTask is { IsCompletedSuccessfully: false })
-        {
-            return ConnectStateMachineAsync(true, connectOpts, cancellationToken);
-        }
-
+        Interlocked.Add(ref _counter.PendingMessages, 1);
```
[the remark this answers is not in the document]
mtmk: copy/paste error. (Was there an argument about Add() being faster?🤷‍♂️)
Fixed.

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -70,589 +91,522 @@ public async ValueTask DisposeAsync()
...
+            lock (_lock)
             {
-                EnqueueCommand(success);
+                if (_pipeWriter == null)
+                    ThrowOnDisconnected();
+                bw = _pipeWriter!;
             }
+
+            _protocolWriter.WriteConnect(bw, connectOpts!);
+            await bw.FlushAsync(cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
mtmk: _pipeWriter is replaced on reconnect in `Reset()`. I wanted to avoid any possible stale reads. afaik in this case semaphore or channel 'async locks' would provide any memory consistency guarantees.

src/NATS.Client.Core/Commands/CommandWriter.cs
mtmk: just swapped this out with a channel. WDYT? (@stebet also raised concerns about `SemaphoreSlim` before)

src/NATS.Client.Core/Commands/CommandWriter.cs
mtmk: it lives as long as the `CommandWriter` object (and in turn `NatsConnection` object). On reconnect pipeline is renewed (not pooled at the moment) but the read loop stays in place.

src/NATS.Client.Core/Commands/CommandWriter.cs
mtmk: not me unfortunately 😅 originally taken from the community toolkit (as mentioned in NatsBufferWriter.cs class). I'll make sure to put an attribution comment later.

src/NATS.Client.Core/Commands/CommandWriter.cs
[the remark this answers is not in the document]
to11mtm: >  (Was there an argument about Add() being faster?🤷‍♂️)

It _might_ be?

As always devil is in the details, I can't comment for ARM but for x86/x64 I did start to look at the tables on [Agner Fog](https://www.agner.org/optimize/instruction_tables.pdf). However,

 - It has been a minute since I have peeked at these tables
 - They do not list `LOCK INC` 
 - It's entirely possible that the JIT decides to pick the best option regardless...

However digging did lead me to [*this*](https://c9x.me/x86/html/file_module_x86_id_140.html) which claims that INC doesn't update the carry flag compared to `ADD`. Whether that still makes a difference in modern arches, I can't say...

This does almost certainly qualify as a micro opt either way, `.Increment()`/`.Decrement()` feel more readable but am happy with either option.

to11mtm (commented):
Left food for thought, primarily around the channel changes.

In summary:

As we are using the channel as a semaphore this way (i.e. using full channel to signal waiting as opposed to an empty channel to signal waiting,) it is better to use `WriteAsync` throughout *if* we want to have better ordering fairness and minimize thrashing on concurrent writes.

If we -do- need to 'prioritize' certain locks, we may want to consider how to do (same problem exists with a semaphore

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -1,67 +1,87 @@
...
+        _cts = new CancellationTokenSource();
+        _readerLoopTask = Task.Run(ReaderLoopAsync);
     }
 
-    public PipeReader PipeReader { get; }
-
-    public ChannelReader<QueuedCommand> QueuedCommandsReader { get; }
+    public void Reset(ISocketConnection? socketConnection)
+    {
+        var pipe = new Pipe(new PipeOptions(
```
caleblloyd, line 57: The current implementation does not reset the Pipe when a socket connects.  The reason for this is that a command may already be in the Pipe, but not sent.

This was OK - since the Reader would mark data as Examined but not Processed, the next socket would read from the same Pipe, and pick up that command that was still in the Pipe but had never been sent.

By resetting the pipe here, I think we are potentially losing that unsent command?

mtmk: It is bit of a pain swapping out reader and writer as well. In that case to avoid data loss, additionally I guess we should set buffer position for read advance after socket send?

src/NATS.Client.Core/Commands/PingCommand.cs
```diff
@@ -1,12 +1,56 @@
+using System.Threading.Tasks.Sources;
+using NATS.Client.Core.Internal;
+
 namespace NATS.Client.Core.Commands;
 
-internal struct PingCommand
+internal class PingCommand : IValueTaskSource<TimeSpan>, IObjectPoolNode<PingCommand>
```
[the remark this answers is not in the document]
mtmk: should we pull this into a separate PR? It's kind of irrelevant here and this PR might linger a bit if it ever goes in that is.

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -274,385 +296,374 @@ public ValueTask SubscribeAsync(int sid, string subject, string? queueGroup, int
...
 
-        try
-        {
-            if (_disposed)
-            {
-                throw new ObjectDisposedException(nameof(CommandWriter));
-            }
+internal sealed class NatsPooledBufferWriter<T> : IBufferWriter<T>, IObjectPoolNode<NatsPooledBufferWriter<T>>
+{
+    private const int DefaultInitialBufferSize = 256;
```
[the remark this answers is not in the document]
mtmk: can we derive a value from write buffer size for example? I thinking it might not be relevant if the implementation changes in the future.

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -189,39 +201,54 @@ public ValueTask PongAsync(CancellationToken cancellationToken = default)
...
+            serializer.Serialize(payloadBuffer, value);
+
+        return PublishLockedAsync(subject, replyTo, payloadBuffer, headersBuffer, cancellationToken);
+    }
+
+    public async ValueTask SubscribeAsync(int sid, string subject, string? queueGroup, int? maxMsgs, CancellationToken cancellationToken)
+    {
+        Interlocked.Increment(ref _counter.PendingMessages);
+
+        while (!_channelLock.Writer.TryWrite(1))
```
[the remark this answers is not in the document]
mtmk: TBH I was aiming at high throughput on a tight loop, but that's just my test and I am actually not seeing much change between TryWrite/loop vs. WriteAsync. Happy to go with WriteAsync here.

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -70,31 +90,30 @@ public async ValueTask DisposeAsync()
...
+                await _pipeWriter.CompleteAsync().ConfigureAwait(false);
+            if (_pipeReader != null)
+                await _pipeReader.CompleteAsync().ConfigureAwait(false);
+
+            await _readerLoopTask.ConfigureAwait(false);
         }
         finally
         {
-            _semLock.Release();
+            while (!_channelLock.Reader.TryRead(out _))
```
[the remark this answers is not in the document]
mtmk: so would you say using ReadAsync() would be better?

src/NATS.Client.Core/Commands/CommandWriter.cs
caleblloyd: Lots of the logic in `NatsPipeliningWriteProtocolProcessor` dealt with figuring out command boundaries, and only marking data as Consumed in the Pipe Reader after the entire command was sent.  Would need to revive some of that logic - maintain an Unbounded Channel of Command Sizes that have been Flushed to the Pipe Reader, and Peek / Read those after each successful `Send` to figure out how much data can be marked as Consumed

mtmk: need to think about it a bit but my first thought is it would not matter because writer flush and reader read are always in sync but that would mean getting rid of the sendMemory limitation which might not be a bad idea anyway. write buffer size (I think set to 64K at the moment - which is pipelines default) would be the largest we write to socket.

src/NATS.Client.Core/Commands/CommandWriter.cs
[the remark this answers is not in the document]
to11mtm: > TBH I was aiming at high throughput on a tight loop, but that's just my test and I am actually not seeing much change between TryWrite/loop vs. WriteAsync. Happy to go with WriteAsync here.

How hard would it be to add some concurrent tests where we see what happens when we have a lot of threads trying to do the needful?

src/NATS.Client.Core/Commands/PingCommand.cs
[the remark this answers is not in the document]
to11mtm: Separate PR wouldn't hurt if it's not painful to do in the existing codebase, if we have benchmarks to be sure the alloc savings doesn't cost us, even bettter?

src/NATS.Client.Core/Commands/CommandWriter.cs
to11mtm: > need to think about it a bit but my first thought is it would not matter because writer flush and reader read are always in sync but that would mean getting rid of the sendMemory limitation which might not be a bad idea anyway. write buffer size (I think set to 64K at the moment - which is pipelines default) would be the largest we write to socket.

Two questions, semi related:

1: Do we have tests and/or benchmarks to test differences and functionality for larger payloads? I know it's... far from ideal but sometimes it happens and it may help us see opportuntities for improvement
2: Per comment on buffer size, should we have a follow up to add configurability?

src/NATS.Client.Core/Commands/CommandWriter.cs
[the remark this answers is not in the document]
to11mtm: I think a derived value would be fine if the math can work out.

src/NATS.Client.Core/Commands/CommandWriter.cs
caleblloyd: The main issue I am trying to highlight is a partially sent command. For example, a 5K publish. The first 2K is sent, then the socket drops. That 2K needs to be re-sent next time the socket connects, and the remaining 3K also needs to be sent.

So Bytes shouldn't be marked as consumed right after they are sent. But once an entire command has been sent, its bytes can be marked as consumed. So the reader needs knowledge of the commands sizes ahead of time.

caleblloyd: > writer flush and reader read are always in sync

Socket Send doesn't always send the full buffer, it returns the number of bytes actually sent, which could be less than the supplied buffer.

Plus large commands could be bigger than the Flush size, and Flush is called once but multiple Reads need to happen.

src/NATS.Client.Core/Commands/CommandWriter.cs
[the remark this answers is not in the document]
to11mtm: 🤔 

1. Need to think more about the `.Dispose()` but probably yes.
2. I think in general, `ReadAsync` should be used (unless a strong argument exists for thrashing on non-critical ops)
3. I'll make a sample helper for the pattern I'm thinking of (or pull branch and provide commit link or draft pr against your branch if that's ok and I get the opportunity to)

Where using a `BoundedChannel` vs a `SemaphoreSlim` can be challenging, is in deciding, for lack of better words, 'the right pivot' on how to use the channel.

Depending on the need, and how the underlying synchronization abstraction is implemented, can have a sizeable impact on performance.

Unless there is a compelling reason to do otherwise I think I'm settled on `.ReadAsync` being a better idea for what we are doing here.

If we need more, we can go deeper; `BoundedChannel` is lower alloc than `SemaphoreSlim` for _many_ cases, however there may be future opportunities for a better pooled abstraction[0]


[0] - In general, if a channel has a waiting value on read, it's no alloc. if it's the FIRST waiter with nondefault cancellation token, it gets a pooled instance. if there's a token or the pooled waiter is used, alloc. A _little_ better than semaphoreslim and has in some ways more reliable semantics, but it can be tricky at first.

src/NATS.Client.Core/Commands/PingCommand.cs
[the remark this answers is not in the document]
mtmk: ping command moved here #358

src/NATS.Client.Core/Commands/CommandWriter.cs
mtmk: ok I get it. This is the complication I wanted to avoid hence dropping the buffer on reconnect. I think data loss may be acceptable since we're in at most once territory. Also some messages might fail on a new connection.

caleblloyd: I don't know if dropping is preferable if we can help it. I get that we are in at-most once territory but we also say that sent commands means accepted to a send buffer, and that we transparently handle reconnects. In my mind that means that a reconnect should not cause a command in the send buffer to be dropped during a reconnect.

caleblloyd: > Also some messages might fail on a new connection

I think you're right, maybe PING and SUB should be skipped if the socket cycles?  It definitely complicates the logic...

stebet:
Taking a look at this now

src/NATS.Client.Core/Commands/CommandWriter.cs
mtmk: one thing we should do though that we should drop the message from the buffer whether the socket sent was successful or not to ensure at-most-once. I'm not sure but there may be cases part of the data is sent across even though send threw an exception (might be totally wrong here).

stebet:
This might also be something to take a look at to make things simpler for locking: https://dotnet.github.io/dotNext/features/threading/exclusive.html

src/NATS.Client.Core/Commands/CommandWriter.cs
caleblloyd: I added a test in #360 to capture the current behavior, should we want to preserve that

mtmk:
> This might also be something to take a look at to make things simpler for locking: https://dotnet.github.io/dotNext/features/threading/exclusive.html

This library now net8.0 only. tried pulling the code in but there is way too much dependency I gave up at 6 KLOC.

caleblloyd:
I added a new benchmark to test Parallel Publish in #367 and here are the results:

`main`:

| Method               | Concurrency | Mean     | Error     | StdDev   | Gen0       | Allocated |
|--------------------- |------------ |---------:|----------:|---------:|-----------:|----------:|
| **PublishParallelAsync** | **1**           | **109.7 ms** | **185.20 ms** | **10.15 ms** |  **1000.0000** |   **3.05 MB** |
| **PublishParallelAsync** | **2**           | **250.2 ms** |  **96.15 ms** |  **5.27 ms** | **21000.0000** |  **62.59 MB** |
| **PublishParallelAsync** | **4**           | **260.8 ms** |  **76.89 ms** |  **4.21 ms** | **21000.0000** |  **64.04 MB** |


This PR:

| Method               | Concurrency | Mean     | Error      | StdDev   | Gen0      | Allocated |
|--------------------- |------------ |---------:|-----------:|---------:|----------:|----------:|
| **PublishParallelAsync** | **1**           | **368.4 ms** | **1,667.7 ms** | **91.41 ms** | **1000.0000** |   **3.05 MB** |
| **PublishParallelAsync** | **2**           | **182.7 ms** |   **111.6 ms** |  **6.12 ms** | **2000.0000** |   **6.64 MB** |
| **PublishParallelAsync** | **4**           | **213.2 ms** |   **652.4 ms** | **35.76 ms** | **4000.0000** |  **14.03 MB** |

Looking great!

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -262,8 +297,92 @@ public async ValueTask UnsubscribeAsync(int sid, int? maxMsgs, CancellationToken
...
+                                while (totalSize < totalSent)
+                                {
+                                    int size;
+                                    while (!channelSize.Reader.TryRead(out size))
+                                    {
+                                        await channelSize.Reader.WaitToReadAsync(cancellationToken).ConfigureAwait(false);
+                                    }
+
+                                    totalSize += size;
+                                }
```
caleblloyd, line 356: Don't want to pop off channel before the command has been completely sent.  So I think this needs to be more like:

```
while (true) {

peek channel

if (totalSize + peek > totalSent) {
  break;
}

read channel
totalSize += size

}
```

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -262,8 +297,92 @@ public async ValueTask UnsubscribeAsync(int sid, int? maxMsgs, CancellationToken
...
+
+                                // make sure to mark the buffer only at message boundaries.
+                                // if there was a message sent only partially, we consider it sent even if it might not be
+                                // due to a socket error in following socket send iteration. this is to avoid re-sending
+                                // the same message again, ensuring at-most-once delivery.
+                                completed = buffer.GetPosition(totalSize);
```
caleblloyd, line 362: Need to store the `examined` and `examinedOffset` here, which should be

```
var examined = buffer.GetPosition(totalSent);
var examinedOffset = buffer.GetPosition(totalSent - totalSize`);
```

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -262,8 +297,92 @@ public async ValueTask UnsubscribeAsync(int sid, int? maxMsgs, CancellationToken
...
+            while (true)
+            {
+                var result = await pipeReader.ReadAsync(cancellationToken).ConfigureAwait(false);
+
+                if (result.IsCanceled)
+                {
+                    break;
+                }
+
+                var buffer = result.Buffer;
```
caleblloyd, line 323: need to slice the Examined Offset off the front of the buffer here.  the Examined Offset are the bytes that were sent last iteration, but not marked as consumed

```
var buffer = result.Buffer.Slice(examinedOffset)
```

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -262,8 +297,92 @@ public async ValueTask UnsubscribeAsync(int sid, int? maxMsgs, CancellationToken
...
+                        finally
+                        {
+                            ArrayPool<byte>.Shared.Return(bytes);
+                        }
+                    }
+                }
+                finally
+                {
+                    // Always examine to the end to potentially unblock writer
+                    pipeReader.AdvanceTo(completed, buffer.End);
```
caleblloyd, line 374: should be `pipeReader.AdvanceTo(completed, examined);`

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -262,8 +297,92 @@ public async ValueTask UnsubscribeAsync(int sid, int? maxMsgs, CancellationToken
...
+            {
+                var result = await pipeReader.ReadAsync(cancellationToken).ConfigureAwait(false);
+
+                if (result.IsCanceled)
+                {
+                    break;
+                }
+
+                var buffer = result.Buffer;
+                var completed = buffer.Start;
```
caleblloyd, line 324: `consumed` may be a better name - since `pipeReader.AdvanceTo` takes 2 args - `consumed` and `examined`
need to track `examined` also

```
var examined = buffer.Start
```

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -322,19 +322,18 @@ private static async Task ReaderLoopAsync(ILogger<CommandWriter> logger, ISocket
...
-                var examined = buffer.Start;
+                var examined = buffer.GetPosition(examinedOffset);
 
                 try
                 {
                     if (!buffer.IsEmpty)
                     {
-                        var bufferLength = (int)buffer.Length;
+                        var readBuffer = buffer.Slice(examinedOffset);
+                        var bufferLength = (int)readBuffer.Length;
```
caleblloyd, line 332: I _think_ (but I'm not entirely sure) that an empty write may possibly cause the `readBuffer` to be empty here, so probably good to check for `!readBuffer.IsEmpty` just to be careful

caleblloyd (approved):
LGTM :tada:

PR #347, opened 2024-01-22
 pipe reader: don't mark commands as consumed until pending=0 

caleblloyd:
Includes test from #346 (e90d588d9c2732ffd64729d177bdfe93646e9505) - thanks for recreating @mtmk 

Previously, commands that were partially trimmed but not sent had marked their data from the pipe reader as Consumed.  On a Reconnect, the same amount of data would get Consumed again.

Refactors the Pipe Reader to only mark data as consumed once an entire command has been processed (pending == 0)

Also adds a bunch of comments to help explain the flow


mtmk (approved):
LGTM

PR #354, opened 2024-01-24
Fix unnecessary error log on dispose

mtmk:
Ignore operation cancel error in event handling loop since it'd be cancelled on connection dispose.
This is causing unclean server shutdown in some cases.

original thread https://natsio.slack.com/archives/C1V81FKU6/p1706128132853999

cc @mnmr


caleblloyd (approved):
LGTM

PR #359, opened 2024-01-26
Text matrix CI server versions

mtmk:
When using binaries.nats.dev, branch names `latest` and `main` were both picking up the same development version. Also release now picks the latest 2.9.x

Now we test for: 2.9, 2.10 and main

**Edit:** I was wrong, this isn't working. We still need to find a way to test 2.10.x latest. `main` and `latest` tags are giving us the same binary.

**Edit2:**: We'll use the exact versions until above tags are supported by binaries service.

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -13,8 +13,8 @@ jobs:
       fail-fast: false
       matrix:
         config:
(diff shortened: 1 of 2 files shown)


caleblloyd (approved):
LGTM

PR #361, opened 2024-01-26
Fixed build warnings

mtmk:


tests/NATS.Client.JetStream.Tests/NatsJsContextFactoryTest.cs
```diff
@@ -82,6 +82,7 @@ public void Create_Context_WithMockConnection_Test()
 
     public class MockConnection : INatsConnection
     {
+#pragma warning disable CS0067
```
caleblloyd, line 85: does it need to be re-enabled after the handlers? or at the end of the class?

or is it auto re-enabled at the end of the file?

caleblloyd (approved):
LGTM

PR #362, opened 2024-01-28
Prevent KeyValueStore error when MaxAge is set below the default DuplicateWindow value

rickdotnet:
When setting a MaxAge less than the default DuplicateWindow value, you receive the following error. This PR fixes this.

```
NATS.Client.JetStream.NatsJSApiException: duplicates window can not be larger then max age
   at NATS.Client.JetStream.Internal.NatsJSResponse`1.EnsureSuccess() in C:\SourceControl\Random\nats.net.v2\src\NATS.Client.JetStream\Internal\NatsJSResponse.cs:line 27
   at NATS.Client.JetStream.NatsJSContext.JSRequestResponseAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken) in C:\SourceControl\Random\nats.net.v2\src\NATS.Client.JetStream\NatsJSContext.cs:line 185
   at NATS.Client.JetStream.NatsJSContext.CreateStreamAsync(StreamConfig config, CancellationToken cancellationToken) in C:\SourceControl\Random\nats.net.v2\src\NATS.Client.JetStream\NatsJSContext.Streams.cs:line 21
   at NATS.Client.KeyValueStore.NatsKVContext.CreateStoreAsync(NatsKVConfig config, CancellationToken cancellationToken) in C:\SourceControl\Random\nats.net.v2\src\NATS.Client.KeyValueStore\NatsKVContext.cs:line 115
   at Program.<Main>$(String[] args) in C:\SourceControl\Random\nats.net.v2\sandbox\Example.KeyValueStore.Watcher\Program.cs:line 18
   at Program.<Main>(String[] args)
```

Modified `Example.KeyValueStore.Watcher` code to reproduce:
```cs
var nats = new NatsConnection();

var kv = new NatsKVContext(js);

var config = new NatsKVConfig("e1")
{
    MaxAge = TimeSpan.FromSeconds(30),
};

var store = await kv.CreateStoreAsync(config);

await foreach (var entry in store.WatchAsync<int>())
{
    Console.WriteLine($"[RCV] {entry}");
}
```

### DIFF
--- a/src/NATS.Client.KeyValueStore/NatsKVContext.cs
+++ b/src/NATS.Client.KeyValueStore/NatsKVContext.cs
@@ -108,8 +108,12 @@ public class NatsKVContext : INatsKVContext
             // MirrorDirect =
             // Mirror =
             Retention = StreamConfigRetention.Limits, // from ADR-8
-            DuplicateWindow = TimeSpan.FromMinutes(2), // 120_000_000_000ns, from ADR-8
+            DuplicateWindow = config.MaxAge != default ? config.MaxAge : TimeSpan.FromMinutes(2), // 120_000_000_000ns, from ADR-8
         };
+
+        // https://github.com/nats-io/nats.go/blob/98430acd80423b776149f29d625d158f490ac3c5/jetstream/kv.go#L334-L342
+        if (streamConfig.MaxAge > TimeSpan.Zero && streamConfig.MaxAge < streamConfig.DuplicateWindow)
+            streamConfig.DuplicateWindow = streamConfig.MaxAge;
 
         var stream = await _context.CreateStreamAsync(streamConfig, cancellationToken);
 


mtmk:
Our general policy is not to fix validation issues but throw validation exception.

rickdotnet:
Interesting. I wouldn't expect JS exceptions when operating with the K/V store. But, can fix on my end. No problem.

mtmk:
I didn't mean to close the PR. it's a good use case. I think we should validate the config in ` kv.CreateStoreAsync(config)` before sending it to the server. I personally like your approach but throwing an exception is our validation approach in NATS clients in general.

mtmk:
Actually what you've done seems to be the correct thing:
https://github.com/nats-io/nats.go/blob/98430acd80423b776149f29d625d158f490ac3c5/jetstream/kv.go#L334-L342

I didn't realised duplicate window wasn't actually part of KV config.

src/NATS.Client.KeyValueStore/NatsKVContext.cs
```diff
@@ -108,7 +108,7 @@ public async ValueTask<INatsKVStore> CreateStoreAsync(NatsKVConfig config, Cance
             // MirrorDirect =
             // Mirror =
             Retention = StreamConfigRetention.Limits, // from ADR-8
-            DuplicateWindow = TimeSpan.FromMinutes(2), // 120_000_000_000ns, from ADR-8
+            DuplicateWindow = config.MaxAge != default ? config.MaxAge : TimeSpan.FromMinutes(2), // 120_000_000_000ns, from ADR-8
```
mtmk, line 111: also check against default duplicate window per Go client: https://github.com/nats-io/nats.go/blob/98430acd80423b776149f29d625d158f490ac3c5/jetstream/kv.go#L334-L342

rickdotnet: Added this check.
```
// https://github.com/nats-io/nats.go/blob/98430acd80423b776149f29d625d158f490ac3c5/jetstream/kv.go#L334-L342
if (streamConfig.MaxAge > TimeSpan.Zero && streamConfig.MaxAge < streamConfig.DuplicateWindow)
    streamConfig.DuplicateWindow = streamConfig.MaxAge;
```

mtmk (approved):
LGTM thanks @rickdotnet 💯

src/NATS.Client.KeyValueStore/NatsKVContext.cs
```diff
@@ -108,9 +108,13 @@ public async ValueTask<INatsKVStore> CreateStoreAsync(NatsKVConfig config, Cance
...
```
mtmk, line 111: my fault didn't check setting duplicate-window to max-age here should've been removed
```diff
- DuplicateWindow = config.MaxAge != default ? config.MaxAge : TimeSpan.FromMinutes(2), // 120_000_000_000ns, from ADR-8
+ DuplicateWindow = TimeSpan.FromMinutes(2), // 120_000_000_000ns, from ADR-8
```

cc @scottf

Background: When setting max age default window is always set equals to max age. e.g. if we set max-age=30 minute default windows should still stay at 2 minutes.

PR #367, opened 2024-01-31
MicroBenchmark for PublishParallel

caleblloyd:
Add a micro benchmark for publishing in parallel

Uses JSON serialization in an attempt to make serialization more CPU intensive


to11mtm (approved):
<3

We may want to consider a 'no serializer' case if it's low effort, that said this is gr8!

mtmk (approved):
LGTM

PR #368, opened 2024-02-02
throw away bytes in send buffer after a failed send

caleblloyd:
we have no idea if bytes in send buffer were actually sent on a failed `SendAsync`


mtmk (approved):
LGTM 🎉 thank you :)

Issue #344, opened 2024-01-19
Fix test flappers


mtmk:
* https://github.com/nats-io/nats.net.v2/actions/runs/7568146589/job/20608743720#step:7:81
```
[xUnit.net 00:00:10.38]     NATS.Client.Core.Tests.SlowConsumerTest.Slow_consumer [FAIL]
[xUnit.net 00:00:10.38]       Expected Volatile.Read(ref count) to be less than 20, but found 20.
```

* https://github.com/nats-io/nats.net.v2/actions/runs/7583600734/job/20655578375?pr=343#step:8:253
```
[xUnit.net 00:00:42.23]     NATS.Client.JetStream.Tests.ConsumerFetchTest.Fetch_dispose_test [FAIL]
[xUnit.net 00:00:42.24]       System.TimeoutException : Took too long (00:00:30) waiting until ack pending 0
```

* https://github.com/nats-io/nats.net.v2/actions/runs/7647000095/job/20837000272
```
Test Run Aborted.
Total tests: Unknown
     Passed: 135
    Skipped: 4
 Total time: 5.0011 Minutes
Error: Process completed with exit code 1.
```

* Protocol_parser_under_load
```
[xUnit.net 00:00:12.07]     NATS.Client.Core.Tests.ProtocolTest.Protocol_parser_under_load(size: 1024) [FAIL]
[xUnit.net 00:00:12.07]       Expected counts.Count to be greater than or equal to 3, but found 1.
[xUnit.net 00:00:12.07]       Stack Trace:
[xUnit.net 00:00:12.07]            at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
[xUnit.net 00:00:12.07]            at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)
[xUnit.net 00:00:12.07]            at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message)
[xUnit.net 00:00:12.07]            at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
[xUnit.net 00:00:12.07]            at FluentAssertions.Numeric.NumericAssertions`2.BeGreaterThanOrEqualTo(T expected, String because, Object[] becauseArgs)
[xUnit.net 00:00:12.07]            at FluentAssertions.Numeric.NumericAssertions`2.BeGreaterOrEqualTo(T expected, String because, Object[] becauseArgs)
[...]

PR #371, opened 2024-02-04
Expose Headers on NatsSvcMsg

### LINKED ISSUE #370: Expose headers on NatsSvcMsg
### Proposed change

Hi Guys,

I was working on some nats service code, and got stuck when I found out the headers on the NatsSvcMsg is not exposed

Is there a deeper reason to why the headers on [NatsSvcMsg Object](https://github.com/nats-io/nats.net.v2/blob/main/src/NATS.Client.Services/NatsSvcMsg.cs) is not exposed from the underlying NatsMsg ?

### Use case

Use headers on service requests :smile

### Contribution

If it's ok to expose the headers, and it's just a matter of forwarding them from the NatsMsg, then sure, I can do a pr 👍

niklasfp:
Thi pr adds the following

* Exposes Heades on NatsSvcMsg (forwarded from the underlying NastMsg)
* A simple service test to make sure headers gets passed to and from service endpoints.

Fixes #370

### DIFF
--- a/src/NATS.Client.Services/NatsSvcMsg.cs
+++ b/src/NATS.Client.Services/NatsSvcMsg.cs
@@ -48,6 +48,11 @@ public readonly struct NatsSvcMsg<T>
     public string? ReplyTo => _msg.ReplyTo;
 
     /// <summary>
+    /// Pass additional information using name-value pairs.
+    /// </summary>
+    public NatsHeaders? Headers => _msg.Headers;
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


mtmk:
note test failures are flappers #344

mtmk:
@niklasfp looks good + love the test! Just an admin we need: could you please sign your commits?

mtmk (approved):
LGTM thanks @niklasfp

PR #378, opened 2024-02-07
Reject payloads over the threshold set by server

mtmk:
Proactively reject payloads over the threshold set by server.

Go client reference:
https://github.com/nats-io/nats.go/blob/f42d2be72a8e21f8f684f002a870eefb852ec937/nats.go#L3761


caleblloyd (approved):
LGTM

PR #379, opened 2024-02-07
Validate stream name parameters in JSContext methods.

### LINKED ISSUE #357: Add validation for stream name when contains illegal characters
### Proposed change

Based on jetStream [Docs](https://docs.nats.io/nats-concepts/jetstream/streams#configuration), `Name` cannot contain whitespace, `.`, `*`, `>`, path separators (forward or backwards slash), and non-printable characters. I think it will be easy to add such validation with more descriptive message.
Thanks

### Use case

When I'm creating a stream with name containing invalid characters like`.` I'm receiving the following stack trace which does not give much information about the problem.
`NATS.Client.JetStream.NatsJSApiNoResponseException: No API response received from the server
   at NATS.Client.JetStream.NatsJSContext.JSRequestAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.JSRequestAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.JSRequestResponseAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.CreateStreamAsync(StreamConfig config, CancellationToken cancellationToken)`

### Contribution

yes I can try :)

niklasfp:
This pr adds validation of stream names in the JS context.

The validation checks for: `' ', and '.'` if any of these characters are present in the stream name, an  `ArgumentException` will be thrown. This at least gets us past the `No API response received from the server` error, but it still leaves us with the fact that we'll leave it up to the server to validate the rest of the invalid characters like specified in the [docs](https://docs.nats.io/nats-concepts/jetstream/streams#configuration) 

So in the situation where we have a stream name NOT containing `' ', or '.'` but having one of the other illegal ones,  e.g. `'>'`, it will throw an exception like this: 

```code
NATS.Client.JetStream.NatsJSApiException: stream name is required and can not contain '.', '*', '>'`
```

In essence that means that some invalid stream names will throw `ArgumentException(s)`, others (server validated) will throw `NatsJSApiException`. 

Oh and I've added `.DS_Store`to .gitignore, it's a folder attribute file that MacOS adds all over the place 😄 

Fixes #357

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSContext.cs
+++ b/src/NATS.Client.JetStream/NatsJSContext.cs
@@ -175,16 +175,12 @@ public partial class NatsJSContext
 
     internal static void ThrowIfInvalidStreamName([NotNull] string? name, [CallerArgumentExpression("name")] string? paramName = null)
     {
-#if NET8_0_OR_GREATER
-        ArgumentNullException.ThrowIfNullOrEmpty(name);
-#else
         ArgumentNullException.ThrowIfNull(name, paramName);
 
         if (name.Length == 0)
         {
             ThrowEmptyException(paramName);
         }
-#endif
 
         var nameSpan = name.AsSpan();
         if (nameSpan.IndexOfAny(" .") >= 0)


src/NATS.Client.JetStream/NatsJSContext.cs
```diff
@@ -172,6 +173,26 @@ public NatsJSContext(NatsConnection connection, NatsJSOpts opts)
         throw new NatsJSPublishNoResponseException();
     }
 
+    internal static void ThrowIfInvalidStreamName([NotNull] string? name, [CallerArgumentExpression("name")] string? paramName = null)
+    {
+#if NET8_0_OR_GREATER
+        ArgumentNullException.ThrowIfNullOrEmpty(name);
```
mtmk, line 179: looks like net6 and net8 are going to return different messages when the empty.

mtmk:
> @mtmk PTAL.
> 
> I would love som feedback on the ArgumentException(s) vs NatsJSApiException part 😄

Yea it looks good I think (only had one minor comment). Do you think we shouldn't use ArgumentExceptions?

mtmk:
> > > @mtmk PTAL.
> > > I would love som feedback on the ArgumentException(s) vs NatsJSApiException part 😄
> > 
> > 
> > Yea it looks good I think (only had one minor comment). Do you think we shouldn't use ArgumentExceptions?
> 
> I meant the difference in exceptions that we will get, ' ' and '.' will throw argument exceptions because we validate them, any other invalid characters will be handled by the server, and will be a nats js api exception. Looking at the different client implememtations there's is no common way, java does a lot client side, go does not 😄

We should probably go with Go 😃 
https://github.com/nats-io/nats.go/blob/94eaa2ac8a7a3deb9dd8841e736a2c368b3f97a0/jetstream/jetstream.go#L772-L780

Server seems to check a few more
https://github.com/nats-io/nats-server/blob/e841c030db03fe31b695b4ec992abd1f242d83c4/server/jetstream.go#L2812-L2817

So, I guess if we stick to the same way of thinking (i.e. client only checks for most common errors) what you're checking (space and '.') covers the most common mistakes, I'd say.

src/NATS.Client.JetStream/NatsJSContext.cs
mtmk: if the name is empty wouldn't they throw totally different exception messages? net8 isn't going to have the custom message set in net6, no?

mtmk: > The message used for .net 6 Is copied from .net 8 framework messages, [this is what .net 8 does](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/ArgumentException.cs,126) But I should probably change it to use ArrgumentException.ThrowIfNullOrEmpty for .net 8, it's the same static though 😄
> 

ah nice! didn't realise that

> As a side note, I just tried using characters like `\x7 \x8 \x9 \r\n` in different combinations when creating streams, and it gives all kinds of different behaviour , some ends with no response from server, others with the server creating the stream but silently removing the character, and in one case replacing it with an 'x' 😄 so we should probably consider making all the characters below '\1F' "illegal" as well.

What about 'space'? Is the server replacing that with 'x'? Maybe we should stick to minimal checks as someone might be relying on the current behaviour somehow 😅

mtmk: btw, as mentioned in the issue, docs says:
> Names cannot contain whitespace, ., *, >, path separators (forward or backwards slash), and non-printable characters.

https://docs.nats.io/nats-concepts/jetstream/streams#configuration

mtmk: I'm leaning towards less validation (i.e. leave as it is) even though server not responding in some cases is not dev friendly, they should be very rare. cc @Jarema

caleblloyd (approved):
LGTM

mtmk (approved):
LGTM thanks @niklasfp 💯 🥇

PR #380, opened 2024-02-08
Send buffer fix

mtmk:
If server drops the connection halfway through sending a large message pipelines reader was not recovering after reconnect. With this fix we consider the pending message sent if there were any socket send errors.

PS I could not reproduce the issue connecting to nats-server (hence used the mock server) but the error can be reproduced if proactive max_payload check is commented out. Just try to publish a payload larger than 1MB then try to publish or ping after that. Even though client reconnects you'd see pipeline exceptions: `System.InvalidOperationException: The examined position cannot be less than the previously examined position.`

### DIFF
--- a/src/NATS.Client.Core/Commands/CommandWriter.cs
+++ b/src/NATS.Client.Core/Commands/CommandWriter.cs
@@ -43,6 +43,7 @@ internal sealed class CommandWriter : IAsyncDisposable
     private readonly PipeReader _pipeReader;
     private readonly PipeWriter _pipeWriter;
     private readonly SemaphoreSlim _semLock = new(1);
+    private readonly PartialSendFailureCounter _partialSendFailureCounter = new();
     private ISocketConnection? _socketConnection;
     private Task? _flushTask;
     private Task? _readerLoopTask;
@@ -92,6 +93,7 @@ internal sealed class CommandWriter : IAsyncDisposable
                     _pipeReader,
                     _channelSize,
                     _consolidateMem,
+                    _partialSendFailureCounter,
                     _ctsReader.Token)
                 .ConfigureAwait(false);
             });
@@ -423,12 +425,12 @@ internal sealed class CommandWriter : IAsyncDisposable
         }
     }
 
-    private static async Task ReaderLoopAsync(
-        ILogger<CommandWriter> logger,
+    private static async Task ReaderLoopAsync(ILogger<CommandWriter> logger,
         ISocketConnection connection,
         PipeReader pipeReader,
         Channel<int> channelSize,
         Memory<byte> consolidateMem,
+        PartialSendFailureCounter partialSendFailureCounter,
         CancellationToken cancellationToken)
     {
         try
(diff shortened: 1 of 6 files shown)


src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -500,10 +500,18 @@ internal async Task TestStallFlushAsync(TimeSpan timeSpan)
                             }
 
                             // don't mark the message as complete if we have more data to send
-                            if (totalSize + pending > sent)
+                            if (sendEx == null)
```
caleblloyd, line 503: I don't think this is the correct behavior.  `sent` is already set correctly if there was a `sendEx`; it acts as if all of the bytes were sent during an exception.

So the code that marks which commands were consumed afterwards is the same, whether there was a Send Exception or not.  Any command considered fully sent is consumed.

Lastly, the `sendEx` is thrown, which will cause this task to eventually get recreated, and any examined data will be available to be sent again

caleblloyd:
> PS I could not reproduce the issue connecting to nats-server (hence used the mock server) but the error can be reproduced if proactive max_payload check is commented out.

I think what is happening when the proactive max_payload check is disabled is that `nats-server` is closing the connection when it gets the Payload Size that exceeds its settings, then the code gets stuck in a reconnect loop where it keeps trying to send that message again.

mtmk:
> I think what is happening when the proactive max_payload check is disabled is that `nats-server` is closing the connection when it gets the Payload Size that exceeds its settings, then the code gets stuck in a reconnect loop where it keeps trying to send that message again.

Actually no, pipeline reader fails with `The examined position cannot be less than the previously examined position.` 🤷‍♂️

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -529,12 +525,6 @@ internal async Task TestStallFlushAsync(TimeSpan timeSpan)
 
                         // slice the buffer for next iteration
                         buffer = buffer.Slice(sent);
-
-                        // throw if there was a send failure
-                        if (sendEx != null)
-                        {
-                            throw sendEx;
```
caleblloyd, line 536: Bring this back

src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -474,17 +466,22 @@ internal async Task TestStallFlushAsync(TimeSpan timeSpan)
...
                             sent = await connection.SendAsync(sendMem).ConfigureAwait(false);
                         }
-                        catch (Exception ex)
+                        catch
                         {
                             // we have no idea how many bytes were actually sent, so we have to assume they all were
                             // this could result in message loss, but is consistent with at-most once delivery
-                            sendEx = ex;
-                            sent = sendMem.Length;
+                            if (pending > 0 && buffer.Length >= pending)
```
caleblloyd, line 477: revert this block, and move this logic to the `if (sendEx != null)` block

caleblloyd (approved):
LGTM

PR #374, opened 2024-02-04
Fix NatsSvcServer start time format so that it's no longer culture aware

niklasfp:
Changes the started time format in NatsSvcServer to `yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffffff'Z'`  note the  added ' (ticks)  in  the format, this makes sure its NOT affected by current culture, see  [DateTimeFormatInfo.UniversalSortableDateTimePattern Property](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.datetimeformatinfo.universalsortabledatetimepattern?view=net-8.0) on how the "iso" format is specified in .Net

Fixes #372

PR #373, opened 2024-02-04
Publish - avoid async state machine when possible

caleblloyd:
Brings back optimizations from prior to #346 

- Logic to skip async state machine when possible
- Locks switched back from channel-based to semaphore-based, but it appears they are faster
- Selective send memory consolidation
- Bigger send buffer sizes

Performed some benchmarks on Linux:

## Send 10 Million Messages Serially:

`main`:

```
Completed: 9999701, Awaited: 299

1,412,795 msgs/sec ~ 172.46 MB/sec
```

This PR:

```
Completed: 9999980, Awaited: 20

2,688,705 msgs/sec ~ 328.21 MB/sec
```

So around 1.9x more throughput.

## Publish Serial Microbench

`main`:

| Method       | Iter | Mean     | Error      | StdDev    | Allocated |
|------------- |----- |---------:|-----------:|----------:|----------:|
| **PublishAsync** | **64**   | **126.0 μs** |   **246.1 μs** |  **13.49 μs** |     **360 B** |
| **PublishAsync** | **512**  | **584.0 μs** | **2,328.1 μs** | **127.61 μs** |     **333 B** |
| **PublishAsync** | **1024** | **839.2 μs** |   **365.5 μs** |  **20.03 μs** |     **358 B** |

This PR:

| Method       | Iter | Mean     | Error      | StdDev    | Allocated |
|------------- |----- |---------:|-----------:|----------:|----------:|
| **PublishAsync** | **64**   | **127.1 μs** |   **258.8 μs** |  **14.19 μs** |     **321 B** |
| **PublishAsync** | **512**  | **377.8 μs** |   **677.4 μs** |  **37.13 μs** |     **336 B** |
| **PublishAsync** | **1024** | **750.7 μs** | **1,845.1 μs** | **101.14 μs** |     **339 B** |

Looks pretty similar

## Publish Parallel Microbench

`main`:

| Method               | Concurrency | Mean    | Error    | StdDev   | Gen0       | Gen1      | Allocated |
|--------------------- |------------ |--------:|---------:|---------:|-----------:|----------:|----------:|
| **PublishParallelAsync** | **1**           | **1.708 s** | **4.9311 s** | **0.2703 s** | **10000.0000** |         **-** |  **30.52 MB** |
| **PublishParallelAsync** | **2**           | **1.486 s** | **1.6187 s** | **0.0887 s** | **17000.0000** | **2000.0000** | **105.66 MB** |
[...]

PR #382, opened 2024-02-08
Release 2.1.0-preview.7

mtmk:
* Send buffer fix (#380)
* Publish - avoid async state machine when possible (#373)
* Reject payloads over the threshold set by server (#378)
* Fix NatsSvcServer start time format so that it's no longer culture aware (#374)
* Expose Headers on NatsSvcMsg (#371)

### DIFF
--- a/version.txt
+++ b/version.txt
@@ -1,1 +1,1 @@
-2.1.0-preview.6
+2.1.0-preview.7


caleblloyd (approved):
LGTM

PR #386, opened 2024-02-09
Hosting keyed services fixes

mtmk:
Keyed service was not returning different instances due to factory method not using the key.


caleblloyd (approved):
LGTM

PR #387, opened 2024-02-09
Fixed base64 decode buffer size

mtmk:
Convert.TryFromBase64String output buffer was potentially too small. Fixed to use the base64 string length which would be more than enough to fit decoded data.


caleblloyd (approved):
Lgtm

PR #395, opened 2024-02-13
NuGet icon update

mtmk:
* Sized 128x128 - https://learn.microsoft.com/en-us/nuget/reference/nuspec#icon
* Transparent bg for dark themes
* Taken from https://github.com/cncf/artwork/blob/main/projects/nats/icon/color/nats-icon-color.png


caleblloyd (approved):
LGTM

PR #401, opened 2024-02-18
Use fetch no wait for empty streams

mtmk:
With this PR we formalize the use of Fetch no_wait since even though usage is discouraged there are valid scenarios we need to use fetch with no_wait e.g.:

* Empty streams or where we just want to get a snapshot of messages
* IoT applications where fetch is performed at a very low frequency

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
+++ b/src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
@@ -199,6 +199,32 @@ public class NatsJSOrderedConsumer : INatsJSConsumer
             _fetchConsumerName = string.Empty;
     }
 
+    /// <inheritdoc />
+    public async IAsyncEnumerable<NatsJSMsg<T>> FetchNoWaitAsync<T>(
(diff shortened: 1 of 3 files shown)


src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
```diff
@@ -199,6 +199,32 @@ await foreach (var msg in consumer.FetchAsync(opts, serializer, cancellationToke
...
+        await foreach (var msg in consumer.FetchNoWaitAsync(opts, serializer, cancellationToken))
+        {
+            if (msg.Metadata is not { } metadata)
+                continue;
+
+            _fetchSeq = metadata.Sequence.Stream;
+            yield return msg;
+        }
+
+        var deleted = await TryDeleteConsumer(_fetchConsumerName, cancellationToken);
```
caleblloyd, line 222: should `TryDeleteConsumer` be run inside of a `finally` - for example if the `CancellationToken` throws an `OperationCanceledException`?

caleblloyd (approved):
LGTM

PR #406, opened 2024-02-23
Gracefully close sockets

caleblloyd:
When watching wireshark, I noticed a lot of `RST` instead of graceful `FIN`

So looking into it more

- TCP Socket - calling `_socket.Shutdown(SocketShutdown.Both);` seems to resolve the RST
- SSL - I guess we forgot to Dispose the SSLStream?  oops.  adding the Dispose still results in a RST on Linux in some cases though, but I'm not sure there is much we can do about it
- WebSocket - needed to call `CloseAsync` which looks like it actually does send another application-level packet, so I included a timeout on that.  But I am guessing if the socket is already disconnected that will return immediately with an exception


mtmk:
nice! one test seems to be failing false positive

mtmk (approved):
LGTM

PR #407, opened 2024-02-25
Message error processing

mtmk:
Some operations are silently failing because of serialization exception. e.g. when using KV with ad hoc JSON. With this PR we're exposing these exceptions through an `Error` property on `NatsMsg` object so that decisions can be made in application code rather than client ending the subscription and hiding the exception until the subscription object is disposed, which is only done after ending the enumeration (in turn which might not happen until the application terminates)

Error property is actually kept on the header object to not use any additional space on `NatsMsg` struct.

Now application code can check for errors during enumeration:

```csharp
    await foreach (var msg in nats.SubscribeAsync<int>("test"))
    {
        if (msg.Error != null)
        {
            Console.WriteLine($"Error: {msg.Error}");
            continue;
        }
        
        Console.WriteLine(msg);
    }
```


caleblloyd:
Why are they silently failing - since it calls `SetException` and ends the subscription, shouldn't that exception bubble up to the caller?

mtmk:
I noticed this with KV watchers. it's probably a deeper issue with how KV watchers work in that case.

mtmk:
do you think we should rather buble the exception up to the user in KV serialization?

I still think this log entry is valuable since there is a chance subscription may just hang because of a serialization exception and since the exception only propagated when the subscription is disposed.

caleblloyd:
I think right now we typically only log errors that happen in the background, and not errors that result in exceptions thrown to the caller.

If possible, we should try to bubble it up.

If you think it's also valuable to log serialization errors we could do that too.

mtmk:
I'm mainly worried about silent failures, not serialization errors. in this case while using KV watch, it failed to work (just hang) without showing any errors despite logging being active. not just KV watch but another similar thing could happen in production, making it impossible to identify or even be aware of any issues.

In KV watcher case we can do what we're doing with services: Use binary serialization at subscription, serialize at KV level and add Exception property to KV entry object and/or throw serialization exceptions from await foreach loops, potentially optionally.

caleblloyd:
Yes, we want to avoid silent failures, if we can, by having the exceptions bubble up and showing up in user code.  Not everyone will have the logger turned on or pay attention to it, so logging shouldn't be the _only_ way to catch a silent failure.

mtmk:
> My 2 cents, I agree that exceptions like serialization should not be "log only". If I'm watching e.g. KV, then I would expect the async enumeration to explode and bubble the exception.

👍

Would you also say it may be expected to be optional and have an exception property on the entry if it's opt-in?

> But e.g reconnects would be nice if it is logged (don't know if that already is the case), I guess that could happen behind the scenes and actually end up in a good state again ?

yes, reconnects are logged at the connection level and KV watchers should recover and carry on yielding where they left off.

mtmk:
thinking about it we have the same issue with all the consumers as well!

src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -259,3 +299,29 @@ private void CheckReplyPreconditions()
         }
     }
 }
+
+public class NatsMsgError
```
caleblloyd, line 303: Seems a little redundant to have an error class that just ends up holding a couple different exceptions.  Would it be better to :

```csharp
public class NatsDeserializeException : NatsException {
  public NatsDeserializeException(inner): NatsException("Exception during deserilaiztion", inner)

  public byte[] Data { get; internal set; }

}

public class NatsHeaderParseException : NatsException {
  public NatsHeaderParseException(inner): NatsException("Exception parsing headers", inner)

  public byte[] Data { get; internal set; }
}
```

mtmk: I like that. But we 'might' have two exceptions at the same time and we'd only have one property on the message i.e. `msg.Error`.

Should we not handle header parser exceptions here?

They are unlikely to happen (only on some client producing malform headers) which would end up in log / on sub dispose.

caleblloyd: Would the 2 exceptions be a Header Parser exception and a De-serialize exception?  Or a Serialization exception and some other exception?

If it is Header Parser exception could we just hit that one stop, not even attempt the de-serialize?

mtmk: yep it's Header Parser exception and a De-serialize exception.

sounds good!

caleblloyd (approved):
LGTM

PR #410, opened 2024-02-26
Allow to pass ReplyTo Topic

ChristophKuper:
I think If someone wants to use nats authentication with request reply the reply topic needs to configureable.

@mtmk would it be possible to check the change / suggest a different implementation. 

Thanks in advance

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.RequestReply.cs
+++ b/src/NATS.Client.Core/NatsConnection.RequestReply.cs
@@ -34,7 +34,7 @@ public partial class NatsConnection
         CancellationToken cancellationToken = default)
     {
         replyOpts = SetReplyOptsDefaults(replyOpts);
-        await using var sub = await RequestSubAsync<TRequest, TReply>(subject, data, headers, requestSerializer, replySerializer, requestOpts, replyOpts, cancellationToken)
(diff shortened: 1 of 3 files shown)


mtmk:
looks reasonable to me but could you explain the use case a little? is this to do with permissions on the server? I might be missing something.

caleblloyd:
`NewInbox()` generates a completely random Inbox for every request-reply

It has the format:

```
{InboxPrefix}.{Random ID for this Client}.{Random ID for this Request}
```

So every time Request or RequestMany is called, it generates a new ID.  That way the response will only go to the requester.

To tighten the permissions on inboxes, so that other clients can't see each other's inboxes, use a unique permission-constrained Inbox Prefix.

caleblloyd:
That is the Request subject.

## For the requesting side:

A connection making a Request to those subjects can already do that.  Then the response will come on the `replyTo` subject, which will be dynamically generated based off the InboxPrefix

## For the serving side

Consider using [NATS.Client.Services](https://nats-io.github.io/nats.net.v2/documentation/services/intro.html) for serving these endpoints

ChristophKuper:
I will close this PR with the inbox prefix user can configure authentication.
Only limitation is that this has to be shared across connection.

PR #412, opened 2024-02-26
[OLD] Add OpenTelemetry Instrumentation

mtmk:
This PR here as a reference for a bit.
* replaced by #437

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,9 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 end_of_line = lf
 
+[*.json]
+indent_size = 2
+
 [src/NATS.Client.Core/NaCl/**.cs]
 generated_code = true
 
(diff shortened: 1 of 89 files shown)


Cooksauce:
@mtmk I think we should hold off on this. I actually rewrote most of this to a different approach which:
- is much less invasive to the internal API
- has behavioral fixes
- has meaningful tests added
- etc...

I think it's a better approach, and so should be considered before merging the code in this branch here. I can probably find time to get it to a mergable state by the end of the week.

[Branch is here](https://github.com/Cooksauce/nats.net.v2/tree/53-latest) (I have some local changes not pushed as well)

cc'ing #124 for visibility

mtmk:
@Cooksauce, that sounds good to me. feel free to create a fresh PR.

mtmk:
see #437

PR #418, opened 2024-02-28
Fix test flappers

mtmk:
fixed core tests. in JS Fetch_dispose_test still sometimes fails. I shall leave that to another PR


caleblloyd (approved):
LGTM

PR #430, opened 2024-03-04
Added KV flow control

mtmk:


caleblloyd (dismissed):
LGTM

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
```diff
@@ -193,6 +193,11 @@ private async Task CommandLoop()
                                         _ => operation,
                                     };
                                 }
+
+                                if (headers is { Code: 100, MessageText: "FlowControl Request" })
+                                {
+                                    await msg.ReplyAsync(cancellationToken: _cancellationToken);
```
caleblloyd, line 199: need `continue;` afterwards?  Or is it correct to fall-through and continue processing?

caleblloyd (approved):
LGTM

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
mtmk: nothing happens after that point but makes sense to continue

PR #438, opened 2024-03-08
NatsKVStore.WatchAsync : Added optional Action which gets called ever…

darkwatchuk:
Added optional Action parameter.  Previously, you could only detect if the enumeration was up-to-date by checking the delta, this is not possible if the KVStore was empty in the first place.  The action, if specified now gets called whenever it's up to date and has reached the eod of the current data.


src/NATS.Client.KeyValueStore/INatsKVStore.cs
```diff
@@ -77,19 +77,21 @@ public interface INatsKVStore
     /// <param name="serializer">Serializer to use for the message type.</param>
     /// <param name="opts">Watch options</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
+    /// <param name="endOfCurrentData">An Action callback for when there is currently no more data to process</param>
     /// <typeparam name="T">Serialized value type</typeparam>
     /// <returns>An asynchronous enumerable which can be used in <c>await foreach</c> loops</returns>
-    IAsyncEnumerable<NatsKVEntry<T>> WatchAsync<T>(string key, INatsDeserialize<T>? serializer = default, NatsKVWatchOpts? opts = default, CancellationToken cancellationToken = default);
+    IAsyncEnumerable<NatsKVEntry<T>> WatchAsync<T>(string key, INatsDeserialize<T>? serializer = default, NatsKVWatchOpts? opts = default, CancellationToken cancellationToken = default, Action endOfCurrentData = null);
```
mtmk, line 83: we should put the action in the options. this is breaking the interface.

src/NATS.Client.KeyValueStore/INatsKVStore.cs
```diff
@@ -14,33 +15,30 @@ public interface INatsKVStore
     /// </summary>
     /// <param name="key">Key of the entry</param>
     /// <param name="value">Value of the entry</param>
-    /// <param name="serializer">Serializer to use for the message type.</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
     /// <typeparam name="T">Serialized value type</typeparam>
-    ValueTask<ulong> PutAsync<T>(string key, T value, INatsSerialize<T>? serializer = default, CancellationToken cancellationToken = default);
```
mtmk, line 20: this file shouldn't change. we don't want to break the interface.

mtmk:
see #439

PR #439, opened 2024-03-08
Added OnNoData Async Func to NatsKVOpts

### LINKED ISSUE #436: Enhancement to KV WatchAsync to notify if the watch has been started on an empty store
### Proposed change

For the KV WatchAsync command, provide some sort of mechanism to say when the original set of data has been returned, currently this can only be detected when there is data coming back from the original watch. However if KV is empty, you won't know.  Possibly just add an optional parameter to WatchAsync like an **Action onInitialComplete** parameter that fires on the first go. Or **Action endOfData** which fires every time delta has hit 0 if there is no data currently left to come in.. (probably the latter is better?)

### Use case

When using the latest .Net client library (ver 2) and doing a kvStore.WatchAsync, I currently release a semaphore the first time delta hits 0, this denotes that I've received all values to date, the watch then continues....  I tend to start this watch at the beginning of the program and want to make sure that I've received all current KV values before the rest of the program continues..   This works great, all current values come in, semaphore gets triggered, and further events come through and get processed.  However, when the KV is initially empty, nothing comes back in the WatchAsync (obviously!).... and the program can't continue.....   In V1 of the library there was an event that came through to say, first lot of data is back.  

### Contribution

_No response_

darkwatchuk:
Added async Func for callback for when there is no more data left to process

* resolves #436


mtmk:
looking good. just a couple of small things:
1) could you run dotnet format to make sure code format is inline with the editorconfig
1) could you also see the warnings are cleared in the build

mtmk:
Could you also write a test? it's a small change but I think we should have a test for it.

mtmk:
Looking at the tests I just had a thought! should we use the return value of the callback to potentially yield break out of the loop?

darkwatchuk:
> Looking at the tests I just had a thought! should we use the return value of the callback to potentially yield break out of the loop?

There's no data to loop on, so how to yield break? Unless you wait, insert data later and then break?  Seems a bit pointless though...  Happy for you to change though?

mtmk:
> > Looking at the tests I just had a thought! should we use the return value of the callback to potentially yield break out of the loop?
> 
> There's no data to loop on, so how to yield break? Unless you wait, insert data later and then break? Seems a bit pointless though... Happy for you to change though?

I mean change the callback signature to `Func<CT, Task<bool>>` then if callback returns `true` yield break:

```csharp
if (await opts.EndOfCurrentData(cancellationToken))
    yield break;
```

darkwatchuk:
> > > Looking at the tests I just had a thought! should we use the return value of the callback to potentially yield break out of the loop?
> > 
> > 
> > There's no data to loop on, so how to yield break? Unless you wait, insert data later and then break? Seems a bit pointless though... Happy for you to change though?
> 
> I mean change the callback signature to `Func<CT, Task<bool>>` then if callback returns `true` yield break:
> 
> ```cs
> if (await opts.EndOfCurrentData(cancellationToken))
>     yield break;
> ```

Got ya, ok.... So a configurable optional break. Good idea. Let me change.....

mtmk (approved):
LGTM thanks @darkwatchuk

src/NATS.Client.KeyValueStore/NatsKVOpts.cs
```diff
@@ -36,6 +36,11 @@ public record NatsKVWatchOpts
     /// Meta only to get the metadata of the entries
     /// </summary>
     public bool MetaOnly { get; init; } = false;
+
+    /// <summary>
+    /// Async function called when the enumerator reaches the end of data. Return True to break the async enumeration, False to allow the enumeration to continue.
```
caleblloyd, line 41: The name and description does not make sense to me... What is the "end of data"? I can see from the code that it is NumPending==0, but the name/description do not indicate that.

Also, why does it need to be a callback? Would it be better to make it a Boolean `StopWhenNumPendingZero` or something along those lines?

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -315,11 +315,27 @@ public async IAsyncEnumerable<NatsKVEntry<T>> WatchAsync<T>(string key, INatsDes
...
             {
                 yield return entry;
+
+                if (entry.Delta == 0 && opts?.EndOfCurrentData != null)
+                {
+                    if (await opts.EndOfCurrentData(cancellationToken))
+                    {
+                        yield break;
+                    }
+                }
```
caleblloyd, line 338: This could already be done in regular enumeration - checking `entry.Delta` and optionally running a funciton

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -315,11 +315,27 @@ public async IAsyncEnumerable<NatsKVEntry<T>> WatchAsync<T>(string key, INatsDes
     {
         await using var watcher = await WatchInternalAsync<T>(key, serializer, opts, cancellationToken);
 
+        if (watcher.InitialConsumer.Info.NumPending == 0 && opts?.EndOfCurrentData != null)
+        {
+            if (await opts.EndOfCurrentData(cancellationToken))
+            {
+                yield break;
+            }
+        }
```
caleblloyd, line 324: This is the part that I agree we need an option for, and it may even be good to have a consistent option in `NatsKVOpts` and `NatsConsumeOpts`

mtmk: unfortunately this 'trick' only applies to push consumers which are only used for KV. for other consumers solution is to use fetch-no-wait. pull consumers would have a race between the consumer creation and pull request i.e. you might think there is data but by the time pull request is issued that might not be the case anymore causing enumeration to hang.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
mtmk: @darkwatchuk, I'm going to remove this bit if that's ok with you? We can always check this in the loop as @caleblloyd suggested, and also callback can be just for one thing; there was no data.

mtmk:
Also naming wise I guess we should go with the convention of `OnXxxx`. @oising would you say that would be more inline with the naming conventions?

darkwatchuk:
Hi, Sorry, have been away for several days.  Have changed the callback function name and removed from the inner loop.

caleblloyd (approved):
LGTM

PR #440, opened 2024-03-11
Added DuplicateWindow to NatsKVConfig

darkwatchuk:
The DuplicateWindow TimeSpan needs to be exposed for KVStores in order to facilitate certain patterns like leader election within a reasonable time period.  Currently it's fixed at 2 minutes, so you can't reduce MaxAge to less than this.

This PR just exposes DuplicateWindow via NatsKVConfig as an optional parameter with the default left at the current 2 minutes.

### DIFF
--- a/src/NATS.Client.KeyValueStore/NatsKVConfig.cs
+++ b/src/NATS.Client.KeyValueStore/NatsKVConfig.cs
@@ -61,11 +61,6 @@ public record NatsKVConfig
     /// </summary>
     public bool Compression { get; init; }
 
-    /// <summary>
-    /// The time window to track duplicate messages for
-    /// </summary>
-    public TimeSpan? DuplicateWindow { get; init; } = null;
-
     // TODO: Bucket mirror configuration.
     // pub mirror: Option<Source>,
     // Bucket sources configuration.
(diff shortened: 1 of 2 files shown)


mtmk:
this might potentially be a contentious change and not inline with the spec. @ripienaar is this acceptable?

ripienaar:
Duplicate window should == MaxAge when MaxAge is below 2 minutes.

https://github.com/nats-io/nats.go/blob/52a2d33b791b6e819aacdaeef2c0b79c2cfbf81b/jetstream/kv.go#L573

mtmk:
thanks @ripienaar 

So, it looks like we just need to implement the validation but there is not reason why you can't make the duplicate window configurable.

OK just realised we did have a related change in #362. Just merged that, so you need to fix conflicts, sorry about that.

darkwatchuk:
Have mimicked the Go Client.  Go client options don't allow you to configure the duplicate window.  It just sets duplicate window to max age if max age is < 2 mins....

mtmk:
@darkwatchuk I thought you also wanted to make duplicate window configurable, no?

ripienaar:
I'd prefer not to make it configurable - we dont want to over expose stream internals

darkwatchuk:
> @darkwatchuk I thought you also wanted to make duplicate window configurable, no?

No.  I just needed a MaxAge < 2 mins.  The only way to do this was reduce the window.  Am happy with the Go client implementation of auto setting both to be the same.

darkwatchuk:
Ok - fine if just merged.  I'll close this.

mtmk:
> Ok - fine if just merged. I'll close this.

sorry about the confusion. thanks @darkwatchuk

PR #444, opened 2024-03-16
Added source generated helper method for registering service endpoints.

adamreed90:
I have outlined a method for Attribute based registration of Service Endpoints. 

This is done via a source generated extension method:

```
Task RegisterEndpointsAsync(this INatsSvcServer service, CancellationToken cancellationToken = default)
```

Methods are designated with the attribute `ServiceEndpoint`

Example:
```
 public class MyService
 {
     [ServiceEndpoint(name:"e1", queueGroup:null )]
     public async ValueTask<int> Endpoint1(NatsSvcMsg<int> m, CancellationToken cancellationToken)
     {
         if (m.Data == 7)
         {
             await m.ReplyErrorAsync(m.Data, $"Error{m.Data}", cancellationToken: cancellationToken);
             return 0;
         }

         if (m.Data == 8)
         {
             throw new NatsSvcEndpointException(m.Data, $"Error{m.Data}");
         }

         if (m.Data == 9)
         {
             throw new Exception("this won't be exposed");
         }

         await m.ReplyAsync(m.Data * m.Data, cancellationToken: cancellationToken);
         return m.Data * m.Data;
     }
 }
 ```
 
 Then the extension method is called on the created service.
 
Example:
```
  await using var s1 = await svc.AddServiceAsync("s1", "1.0.0", cancellationToken: cancellationToken);

  await s1.RegisterEndpointsAsync(cancellationToken);
  ```
  
  I was unable to get the source generated code to work with XUnit, however the code is very simple and straightforward.

### DIFF
--- a/src/NATS.Client.Services.EndpointGenerator/NATS.Client.Services.EndpointGenerator.csproj
+++ b/src/NATS.Client.Services.EndpointGenerator/NATS.Client.Services.EndpointGenerator.csproj
@@ -1,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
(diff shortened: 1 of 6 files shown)


adamreed90:
Looks like this will need more work, Source Generators must target .NET Standard :(

Issue #450, opened 2024-03-22
JetStream publish performance


safayatborhan:
### Observed behavior

I have a dotnet 6 application running on nats.net. If I migrate to nats.net.v2, the performance is degrading drastically. Here is the code I am running with [nats.net.v2](https://github.com/nats-io/nats.net.v2) library: 

``` csharp
var natsOptions = NatsOpts.Default;
natsOptions = natsOptions with { Url = "localhost" };
await using var nats = new NatsConnection(natsOptions);
var js = new NatsJSContext(nats);
var config = new StreamConfig(name: "EVENTS", subjects: new[] { "events.>" });
config.Storage = StreamConfigStorage.File;
var stream = await js.CreateStreamAsync(config);

var tasks = new List<Task>();
for (int i = 0; i < 1000; i++)
{
    var task = Task.Run(async () =>
    {
        while (true)
        {
            var sw = Stopwatch.StartNew();
            for (var i = 0; i < 2; i++)
            {
                await js.PublishAsync<object>(subject: "events.page_loaded", data: null);
                await js.PublishAsync<object>(subject: "events.mouse_clicked", data: null);
                await js.PublishAsync<object>(subject: "events.mouse_clicked", data: null);
                await js.PublishAsync<object>(subject: "events.page_loaded", data: null);
                await js.PublishAsync<object>(subject: "events.mouse_clicked", data: null);
                await js.PublishAsync<object>(subject: "events.input_focused", data: null);
            }
            Console.WriteLine($"Total time taken: {sw.Elapsed.TotalSeconds}");
        }
    });
    tasks.Add(task);
}

await Task.WhenAll(tasks);
```

Sample output: 
![image](https://github.com/nats-io/nats.net.v2/assets/13792572/b653bfb5-2d30-44f0-b9df-e5cddd2d1f30)

Here is the same code I am running with [nats.net](https://github.com/nats-io/nats.net) library: 

``` csharp
Options opts = ConnectionFactory.GetDefaultOptions("localhost");
ConnectionFactory connectionFactory = new ConnectionFactory();
var conn = connectionFactory.CreateConnection(opts);
IJetStream jetStream = conn.CreateJetStreamContext();
IJetStreamManagement jetStreamManagement = conn.CreateJetStreamManagementContext();

jetStreamManagement.AddStream(StreamConfiguration.Builder()
                .WithName("EVENTS")
                .WithStorageType(StorageType.File)
                .WithSubjects("events.>")
                .Build());

[...]

PR #453, opened 2024-03-24
Request reply baseline benchmark

mtmk:
Request reply performance improvements: The basic idea is to create a shortcut to avoid additional subscriber (albeit muxed) creation so that we can save on additional allocations and potentially work. reply many can stay the same since creating a subscriber makes sense there.

This is a baseline benchmark we can merge now so that when the actual performance work is done we can have an easy way to compare results:

| Method            | Mean     | Error    | StdDev  | Allocated |
|------------------ |---------:|---------:|--------:|----------:|
| RequestReplyAsync | 217.4 us | 41.71 us | 2.29 us |   3.68 KB |

* see also #450


stebet:
@mtmk One thing I did in my app that uses the v2 lib is to not use the built-in request-reply since it creates a new subscription for every request. I just use one subscription generating a random prefix (inbox if you will) and then reuse that subscription for all requests, appending a unique ID to the prefix, f.ex:
`_somerandominboxid.randomrequestid`.

That way my custom subscriber just has to have one active long-running ephemeral subscription to get each response for `_somerandominboxid.*`.

It performs a lot better than the built-in `request-reply` mechanism.

Is there a reason it currently creates a new subscription each time?

mtmk:
@stebet what you're doing is sensible and we should handle it more efficiently in the library. It's not as bad at the moment since we are using a 'mux' inbox as you described, but still creating somewhat a dummy 'inbox subscription' object to handle it. So the communication with the server is optimal but internal handling still a bit costly.

The reason for this approach is to generalize mux subscriptions so application code can leverage that too. e.g. if you create a subscription with the current inbox prefix, that would not create a new subscription on the server. However, this is not documented and I'm not sure if it's used at all.

So the idea in this PR to make a special case for Request-Reply since it's a very common pattern (especially when using JetStream for example) and improve its 'internal' performance. Any ideas, critique welcome, as always.

mtmk:
keeping this PR simple to a benchmark only.

caleblloyd (approved):
Lgtm

PR #454, opened 2024-03-25
Add client certificate chain support and SSL context generation

jmcrumb:
## Context
Current implementation of client certificate loading does not allow for use of chained client certificates, either from the base options of filepath or the `LoadClientCertificate` hook in `NatsTlsOpts`.  The current behavior results in the first certificate in the file to be read, and any subsequent certificates contained are ignored.  For clients which use certificate chaining, this forces server allowlisting of lowest-level intermediate certificates or failure to connect.

## Resolution
Change the SDK to use certificate collections rather than individual certificates as well as expose a hook for custom loading of said collection.  Some of the functionality of this fix, specifically use of `SslStreamCertificateContext` during construction of `SslClientAuthenticationOptions`, is only available in the latest stable .NET major version (.NET 8), so this feature will only have support that the .NET 8 build of the NATS SDK.

### Changes
1. [BREAKING] Change `NatsTlsOpts.LoadCertificate(X509Certificate2)` to `NatsTlsOpts.LoadCertificates(X509Certificate2Collection)`
1. Add support for Certificate Chains
3. Add helper functions for loading certificates from string or file
4. Add `SslStreamCertificateContext` and `X509ChainPolicy` creation for use during SSL handshake

## Additional Notes
During the RCA for this issue another bug was identified.
When attempting to connect from a Windows 10 Enterprise (19045.4046) machine, SSL connection was failing when `NatsTlsOpts.InsecureSkipVerify = false`.  The same code ran without error on dockerized Linux.  I suspect this is because of windows certificate handling (the same logic prompted special handling [here](https://github.com/nats-io/nats.net.v2/blob/main/src/NATS.Client.Core/Internal/TlsCerts.cs#L54-L65), and may need to be applied to [SslStreamConnection.RcsCbCaCertChain](https://github.com/nats-io/nats.net.v2/blob/main/src/NATS.Client.Core/Internal/SslStreamConnection.cs#L108).

In addition, during the failure above, a null pointer exception would occur at [SetupReaderWriterAsync](https://github.com/nats-io/nats.net.v2/blob/a5cee3a6c3c8fd3a5481c216cb862b604b9d81c2/src/NATS.Client.Core/NatsConnection.cs#L385) since `_currentConnectUri` was never set due to a caught error during[ SSL connection and authentication](https://github.com/nats-io/nats.net.v2/blob/a5cee3a6c3c8fd3a5481c216cb862b604b9d81c2/src/NATS.Client.Core/NatsConnection.cs#L319-L324).


mtmk:
> Please note that this PR is meant to highlight the implementation, and as such tests have been omitted as they used credentials which cannot be checked into this repository.

can we create test versions of these with a shell script?

jmcrumb:
@mtmk Updated generator script with the openssl generation of the chained certs and added test cases for a multi-PEM file input

mtmk:
checks seems to be failing. having a look.

mtmk:
@jmcrumb could you sign your commits please?

mtmk:
> Please note that this PR is meant to highlight the implementation, and as such tests have been omitted as they used credentials which cannot be checked into this repository.

is this still true?

jmcrumb:
> is this still true?
No.  Just revised the description.  
And thanks for addressing the test issue!

mtmk:
@jmcrumb, I updated the docs a little. if you can please check the last commit and let me know if you're happy with it then we can merge this PR I think.

src/NATS.Client.Core/Internal/SslStreamConnection.cs
```diff
@@ -170,6 +170,42 @@ private SslClientAuthenticationOptions SslClientAuthenticationOptions(NatsUri ur
             rcsCb = RcsCbCaCertChain;
         }
 
+#if NET8_0_OR_GREATER
+        X509ChainPolicy? policy = null;
+        SslStreamCertificateContext? streamCertificateContext = null;
+        if (_tlsCerts?.ClientCerts != null && _tlsCerts.CaCerts != null && _tlsCerts.ClientCerts.Count >= 1)
```
caleblloyd, line 176: Why check for `_tlsCerts.CaCerts != null` here?  Shouldn't CA Certs be handled separately, if they aren't set here then check the OS store.  Could still have a client cert with intermediates and a CA Cert in the OS store.

jmcrumb (commented):
@mtmk Looks good to me!

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -62,7 +63,7 @@ public sealed record NatsTlsOpts
     /// <summary>
     /// Callback that loads Client Certificate
     /// </summary>
-    public Func<ValueTask<X509Certificate2>>? LoadClientCert { get; init; }
+    public Func<ValueTask<X509Certificate2Collection>>? LoadClientCerts { get; init; }
```
caleblloyd, line 66: I don't think we should necessarily pluralize `LoadClientCerts` - it is still just 1 cert plus intermediates.  Plural to me indicates the list of Client Certs will be tried based off some property of the cert such as common name or thumbprint.  Should expand the doc here to explain why it's `X509Certificate2Collection`

src/NATS.Client.Core/Internal/SslStreamConnection.cs
mtmk: good point. Also if CA is in the list of client certs, would we still need to provide CA separately (if not is OS store)?

caleblloyd: The CA list is for validating the Server's certs, so we want the distinction between Client Certs and CA Certs and should not cross them.  The Client and Server certs could technically be signed by different trust roots.

src/NATS.Client.Core/NatsTlsOpts.cs
caleblloyd: Also - should this callback return a [SslStreamCertificateContext](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslstreamcertificatecontext.create?view=net-5.0) instead of an `X509Certificate2Collection`, since it appears that is the correct API for building a client certificate chain with intermediates

mtmk: what about `X509ChainPolicy` it needs that as well

src/NATS.Client.Core/Internal/SslStreamConnection.cs
mtmk: then don't we need clientCAs + serverCAs?

caleblloyd: I think there is no such thing as a Client CA really...  just a Client Cert + optional Client intermediaries

jmcrumb: Its a good callout that the root CA may not be the same for client chain and server chain. @caleblloyd / @mtmk is it a fair assumption that 
1. all behavior around server certificate chain validation will occur in `RcsCbCaCertChain`?
2. Server will always have knowledge of the client chain's root CA?

src/NATS.Client.Core/NatsTlsOpts.cs
jmcrumb: Hard for me to weigh in on this with my limited vision of customers beyond my organization (and what they need).  The most extreme and simple would be to expose the hook for `SslClientAuthenticationOptions` or for a `Func<SslClientAuthenticationOptions,SslClientAuthenticationOptions>` which allows for customizable overrides of the default built in `SslStreamConnection`.  Between that and the current state just seems like different levels of tradeoff between flexibility and consistency.

src/NATS.Client.Core/Internal/SslStreamConnection.cs
caleblloyd: My understanding is "yes" to both of those assumptions

src/NATS.Client.Core/NatsTlsOpts.cs
mtmk: ok, trying the tests without `CertificateChainPolicy` doesn't seem to make any difference. Also looks like it might mess up the OCPS (I might've remembered the acronym wrong) feature:

> If not null, [CertificateRevocationCheckMode](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslclientauthenticationoptions.certificaterevocationcheckmode?view=net-8.0#system-net-security-sslclientauthenticationoptions-certificaterevocationcheckmode) and [SslCertificateTrust](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust?view=net-8.0) are ignored.

src/NATS.Client.Core/Internal/SslStreamConnection.cs
mtmk: OK, got it! So client bundle would include it's root CA for the client. So `SslStreamCertificateContext` doesn't need to know about the `caCert` in our settings at all since our `caCert` is for the server chain validation. and yes server will have to trust at least one of the client's CAs, typically root.

src/NATS.Client.Core/NatsTlsOpts.cs
mtmk: > Also - should this callback return a [SslStreamCertificateContext](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslstreamcertificatecontext.create?view=net-5.0) instead of an X509Certificate2Collection, since it appears that is the correct API for building a client certificate chain with intermediates

the only problem is that `ClientCertificateContext` option (which gets assigned) is not available in net6.0

caleblloyd: Yes, that API popped up in net8, there is a mega thread about it here: https://github.com/dotnet/runtime/issues/26323

caleblloyd:
@mtmk can we merge this to a staging branch?

I think I'd like to undo the breaking change, and split it to:

```
    /// NET6 and NET8, set Obsolete after NET6 EOL
    /// <summary>
    /// Callback that loads Client Certificate
    /// </summary>
    public Func<ValueTask<X509Certificate2>>? LoadClientCert { get; init; }
    
    // NET8 ONLY
    /// <summary>
    /// Callback that loads Client Certificate
    /// </summary>
    public Func<ValueTask<SslStreamCertificateContext>>? LoadClientCertContext { get; init; }
```

I have something staged but I can't push to this PR because it's on a fork

mtmk:
thanks @jmcrumb! we're staging now. we should be ready to merge to main then release very soon.

jmcrumb:
Great to hear, thank you!

PR #455, opened 2024-03-27
Deserialize empty payloads

### LINKED ISSUE #449: Deserializer is skipped for 0 length payloads
### Observed behavior

Create a deserializer that returns a object on empty payloads. 
Send a message with an empty payload.
The  result is null instead of the expected object.

The code directly returns the default value which is null for reference types.

### Expected behavior

The deserializer should be called.
for some reference types, especially in F#, null is not the default value (for instance Map<'a,'b>). 
A deserializer that returns Map.empty is provided, but the client is skipping the deserializer and returns null instead, which is not a valid value for Map<'a,'b>

### Server and client version

Nats.net 2.1.2

### Host environment

not specific

### Steps to reproduce

Create a deserializer that returns a object on empty payloads. 
Send a message with an empty payload.
The  result is null instead of the expected object.

mtmk:
Remove the payload empty check when deserializing so that empty payloads are also passed to the deserializer for it to make the decision. This way, deserializers will have the option of what to do with the empty payload rather than just being skipped and a default value being returned. We also altered the existing default deserializers to still return a default value when the payload is empty to maintain backward compatibility. However, any custom deserializer implementations might break with this update. Applications can place a check for empty payloads or test their deserialization behavior for empty payloads:

```csharp
public class MyDeserialize<T> : INatsDeserialize<T>
    public T? Deserialize(in ReadOnlySequence<byte> buffer)
    {
        if (buffer.Length == 0)
        {
            return default;
        }
        // deserializer implementation
    }
}
```

* resolves #449

### DIFF
--- a/src/NATS.Client.Core/INatsSerialize.cs
+++ b/src/NATS.Client.Core/INatsSerialize.cs
@@ -368,13 +368,11 @@ public class NatsUtf8PrimitivesSerializer<T> : INatsSerializer<T>
     /// <inheritdoc />
     public T? Deserialize(in ReadOnlySequence<byte> buffer)
     {
-        if (buffer.Length == 0)
-        {
-            return default;
-        }
 
         if (typeof(T) == typeof(string))
         {
+            if (buffer.Length == 0)
+                return default;
             return (T)(object)Encoding.UTF8.GetString(buffer);
         }
 
@@ -382,6 +380,9 @@ public class NatsUtf8PrimitivesSerializer<T> : INatsSerializer<T>
 
         if (typeof(T) == typeof(DateTime) || typeof(T) == typeof(DateTime?))
         {
+            if (buffer.Length == 0)
+                return default;
+
             if (Utf8Parser.TryParse(span, out DateTime value, out _))
             {
                 return (T)(object)value;
@@ -392,6 +393,9 @@ public class NatsUtf8PrimitivesSerializer<T> : INatsSerializer<T>
 
         if (typeof(T) == typeof(DateTimeOffset) || typeof(T) == typeof(DateTimeOffset?))
         {
+            if (buffer.Length == 0)
+                return default;
+
             if (Utf8Parser.TryParse(span, out DateTimeOffset value, out _))
             {
                 return (T)(object)value;
@@ -402,6 +406,9 @@ public class NatsUtf8PrimitivesSerializer<T> : INatsSerializer<T>
 
         if (typeof(T) == typeof(Guid) || typeof(T) == typeof(Guid?))
         {
+            if (buffer.Length == 0)
+                return default;
+
             if (Utf8Parser.TryParse(span, out Guid value, out _))
             {
                 return (T)(object)value;
@@ -412,6 +419,9 @@ public class NatsUtf8PrimitivesSerializer<T> : INatsSerializer<T>
 
         if (typeof(T) == typeof(TimeSpan) || typeof(T) == typeof(TimeSpan?))
         {
+            if (buffer.Length == 0)
+                return default;
+
             if (Utf8Parser.TryParse(span, out TimeSpan value, out _))
             {
                 return (T)(object)value;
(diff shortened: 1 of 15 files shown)


caleblloyd:
Should we handle this at the sterilizer level instead?  Make a chain-able `ignoreEmptyPayload` sterilizer and make that the default?

I feel like the setting is tied to the serailizer, not the subscription

mtmk:
> Should we handle this at the sterilizer level instead? Make a chain-able `ignoreEmptyPayload` serailizer and add that as a default

good idea but would still be a breaking change. it may break custom serializers application is using.

thinkbeforecoding:
Is there a way to keep existing serializers as is, and be able to define new deserializers that parse empty as non null ? Like using an interface that inherit from IDeserialize...

mtmk:
> Is there a way to keep existing serializers as is, and be able to define new deserializers that parse empty as non null ? Like using an interface that inherit from IDeserialize...
> That would not be a breaking change

genius! 🥇 ... created [INatsDeserializeWithEmpty marker interface](https://github.com/nats-io/nats.net.v2/pull/455/commits/1e68c46c9bfde0b4c43c8aaa96a37cab865b2745)

thinkbeforecoding:
That seems perfect.

caleblloyd:
For what it's worth, the [Interface Design Guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/interface?redirectedfrom=MSDN) say:

> ❌ AVOID using marker interfaces (interfaces with no members).
> If you need to mark a class as having a specific characteristic (marker), in general, use a custom attribute rather than an interface.

Should we use a custom attribute here instead?

caleblloyd:
Another issue I see is the way that we have our serializes chain-able would mean that this marker interface (or attribute, if changed to an attribute) would mean that the marker interface or attribute would have to be on the first serailizer in the chain.  What if my chain is

```
NatsRawSerializer (does not serialize empty) -> CustomSerailizer (does serialize empty)
```

Even though my `CustomSerailizer` is marked, it still does not work

If we want to preserve the chain-ability, we should probably re-think the approach to either

1. Add another property to `INatsDeserialize` such as `HandlesEmptyPayload`
2. Always try to Deserailize, but catch certain kinds of exceptions like `System.NotSupportedException` and return `null` when the payload is empty instead of throwing.  Biggest issue with this approach is different serailizers may use different exceptoin types for this

mtmk:
> AVOID using marker interfaces

didn't know this. I see it is a workaround and somewhat misusing a feature of the language etc. at the same time I don't like to be dogmatic. if it works, it's OK! but never mind...

> 1) Add another property to INatsDeserialize such as HandlesEmptyPayload
> 2) Always try to Deserailize, but catch certain kinds of exceptions like System.NotSupportedException

If we're going to introduce a breaking change, maybe we should fix the issue and just remove the 'payload > 0' check and be done with it. We already are going for a breaking release, maybe just fix the issue instead of working around it. as I think @caleblloyd already suggested this, we can fix the default serializers to handle empty payloads.

WDYT?

src/NATS.Client.Core/INatsSerialize.cs
```diff
@@ -368,6 +368,11 @@ public void Serialize(IBufferWriter<byte> bufferWriter, T value)
     /// <inheritdoc />
     public T? Deserialize(in ReadOnlySequence<byte> buffer)
     {
+        if (buffer.Length == 0)
+        {
+            return default;
+        }
```
caleblloyd, line 374: Shouldn't this only be run if `T` is one of those that is normally handled by this deserailizer?  Otherwise it is not possible to chain a deserailzer that handles an empty payload for another `T`.

src/NATS.Client.Core/INatsSerialize.cs
```diff
@@ -613,6 +618,11 @@ public void Serialize(IBufferWriter<byte> bufferWriter, T value)
...
```
caleblloyd, line 624: Shouldn't this only be run if `T` is one of those that is normally handled by this deserailizer?  Otherwise it is not possible to chain a deserailzer that handles an empty payload for another `T`.

caleblloyd (approved):
Needs a format and a few tests failing then LGTM

PR #456, opened 2024-03-28
Release 2.2.0

mtmk:
* #451 Dispose fixes
* #433 Update NATS.Client.Hosting package as NATS.Extensions.Microsoft.DependencyInjection
* (staged) #454 Add client certificate chain support and SSL context generation
   * #463 add NatsTlsOpts.LoadClientCertContext 
* #459 Object store metadata fix
* #455 Deserialize empty payloads

[nats:update-docs]


caleblloyd:
I think since #454 is breaking that this should be at least a Feature release - 2.2.0.  It's not necessarily that common of a break since most folks are probably not using a custom callback to load the cert, but nonetheless it is breaking

mtmk:
> I think since #454 is breaking that this should be at least a Feature release - 2.2.0. It's not necessarily that common of a break since most folks are probably not using a custom callback to load the cert, but nonetheless it is breaking

I like that. I'll leave the branch since don't want to create a new PR.

caleblloyd (approved):
LGTM once #463 is merged

PR #463, opened 2024-04-02
add NatsTlsOpts.ConfigureClientAuthentication

### LINKED ISSUE #285: TLS Certificate Reloading
### Proposed change

#283 introduces a couple callbacks to load TLS Certificates on top of the PEM File paths that already exist

For PEM Files-
On reconnect, reload cert from disk if it has changed.

For Callbacks-
On reconnect, invoke the callback every time.

Document the behavior, so that the end-user knows when to expect that certificate reloading might occur.

### Use case

Automatically rotating certificates

### Contribution

Yes

caleblloyd:
Makes following changes to `NatsTlsOpts`
- Adds: `Func<SslClientAuthenticationOptions, ValueTask>? ConfigureClientAuthentication`
- Obsoletes:
  - `LoadClientCert`
  - `LoadCaCerts`
  - `CertificateRevocationCheckMode`
  
Adds a `NatsTlsOptsExtensions` class with extension methods for loading CA Certs and Client Certs

* Resolves #285


caleblloyd:
> On Windows, ephemeral keys/certificates do not work with schannel. e.g. unless stored in certificate store.

Need to confirm if setting `SslClientAuthenticationOptions.ClientCertificateContext` adds the ephemeral keys/certificates on windows.  [SslStreamCertificateContext.Windows.cs](https://github.com/dotnet/runtime/blob/4bbde33ac01496375ee8902c886c9f0c3c7c709c/src/libraries/System.Net.Security/src/System/Net/Security/SslStreamCertificateContext.Windows.cs#L43-L101) seems to add intermediates to the store but I'm not sure about leafs.

Will just running the tests on windows confirm or deny whether it's all working?  Or is this only on certain versions of windows?

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -156,32 +145,181 @@ internal bool TryTls(NatsUri uri)
...
+        }
+
+        return options;
+#pragma warning restore CS0618 // Type or member is obsolete
+    }
+}
+
+public static class NatsTlsOptsExtensions
+{
+    public static SslClientAuthenticationOptions LoadClientCertFromPem(this SslClientAuthenticationOptions options, string certPem, string keyPem, bool offline = false, SslCertificateTrust? trust = null)
```
caleblloyd, line 214: @mtmk what is your opinion of putting these as extension methods on `SslClientAuthenticationOptions`?  Is that a good pattern or is there a better pattern to use here?

mtmk: agreed. would be much cleaner.

tests/NATS.Client.Core.Tests/TlsOptsTest.cs
```diff
@@ -0,0 +1,195 @@
...
+            Encoding.UTF8.GetString(decryptValue).Should().Be("test123");
+#else
+            Assert.NotNull(clientOpts.ClientCertificates);
+            Assert.Single(clientOpts.ClientCertificates);
+#endif
+        }
+    }
+
+    [Fact]
+    public async Task Load_client_cert_chain_and_key()
```
mtmk, line 91: This one is failing on Windows with .NET 8.0
```
[xUnit.net 00:00:03.67]     NATS.Client.Core.Tests.TlsOptsTest.Load_client_cert_chain_and_key [FAIL]
[xUnit.net 00:00:03.67]       Assert.Equal() Failure
[xUnit.net 00:00:03.67]       Expected: 2
[xUnit.net 00:00:03.67]       Actual:   3
[xUnit.net 00:00:03.67]       Stack Trace:
[xUnit.net 00:00:03.67]         C:\Users\mtmk\src\nats.net.v2-1\tests\NATS.Client.Core.Tests\TlsOptsTest.cs(125,0): at NATS.Client.Core.Tests.TlsOptsTest.<Load_client_cert_chain_and_key>g__ValidateAsync|2_0(NatsTlsOpts opts)
[xUnit.net 00:00:03.68]         C:\Users\mtmk\src\nats.net.v2-1\tests\NATS.Client.Core.Tests\TlsOptsTest.cs(96,0): at NATS.Client.Core.Tests.TlsOptsTest.Load_client_cert_chain_and_key()
[xUnit.net 00:00:03.68]         --- End of stack trace from previous location ---
[xUnit.net 00:00:03.68]   Finished:    NATS.Client.Core.Tests
  Failed NATS.Client.Core.Tests.TlsOptsTest.Load_client_cert_chain_and_key [103 ms]
```

tests/NATS.Client.Core.Tests/TlsOptsTest.cs
```diff
@@ -0,0 +1,195 @@
...
+            var ctx = clientOpts.ClientCertificateContext;
+            Assert.NotNull(ctx);
+
+            var leafCert = ctx.TargetCertificate;
+            leafCert.Subject.Should().Be("CN=leafclient");
+            var encryptValue = leafCert.GetRSAPublicKey()!.Encrypt(Encoding.UTF8.GetBytes("test123"), RSAEncryptionPadding.OaepSHA1);
+            var decryptValue = leafCert.GetRSAPrivateKey()!.Decrypt(encryptValue, RSAEncryptionPadding.OaepSHA1);
+            Encoding.UTF8.GetString(decryptValue).Should().Be("test123");
+
+            Assert.Equal(2, ctx.IntermediateCertificates.Count);
```
mtmk, line 125: On Windows fails here with 3 intermediate certs, last one being CA.

mtmk: It looks like when creating the context from intermediates (I don't know how) the root CA is populated and then removed on Linux but not on Windows.

https://github.com/dotnet/runtime/blob/223e9e71954433544106d5c90e06b5c8113e51d9/src/libraries/System.Net.Security/src/System/Net/Security/SslStreamCertificateContext.cs#L77

https://github.com/dotnet/runtime/blob/223e9e71954433544106d5c90e06b5c8113e51d9/src/libraries/System.Net.Security/src/System/Net/Security/SslStreamCertificateContext.Windows.cs#L12

https://github.com/dotnet/runtime/blob/223e9e71954433544106d5c90e06b5c8113e51d9/src/libraries/System.Net.Security/src/System/Net/Security/SslStreamCertificateContext.Linux.cs#L24

tests/NATS.Client.Core.Tests/TlsOptsTest.cs
mtmk: it's only happening on my PC.

tests/NATS.Client.Core.Tests/TlsOptsTest.cs
mtmk: it's only happening on my PC.

mtmk (approved):
LGTM

PR #465, opened 2024-04-04
[EXPERIMENTAL] NATS protocol parser

mtmk:
This is an experimental NATS Protocol Parser with zero allocations. Somewhat inspired by the .NET Runtime System.Text.Json library's Utf8JsonReader.

| Method | Mean     | Error    | StdDev   | Allocated |
|------- |---------:|---------:|---------:|----------:|
| Parse  | 862.5 ns | 320.3 ns | 17.56 ns |         - |


mtmk:
too old. need a fresh look at the solution in a clean way.

PR #473, opened 2024-04-13
Make Build() on NatsBuilder Internal

rickdotnet:
When swapping my code over to use the new hosting extensions I found what appears to be blunder of mine on the visibility of the `Build()` method. In order to prevent confusion for consumers thinking they need to call this method, I think hiding it would be best.

```cs
services.AddNatsClient(
    nats =>
    {
        nats.ConfigureOptions(ops => ops with { Url = config.Url });
        nats.Build(); // oops, this could cause confusion
    });
```


mtmk:
looks good. just checking the issue with ci

mtmk (approved):
LGTM thanks @rickdotnet

PR #482, opened 2024-04-16
Always use seed URLs when reconnecting

mtmk:


src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -530,12 +530,17 @@ private async void ReconnectLoop()
...
+                                     ?? Array.Empty<NatsUri>();
+
+            // Always keep the original seed URLs in the list of URLs to connect to
+            var connectUrls = serverReportedUrls.Concat(Opts.GetSeedUris());
+
+            var urls = Opts.NoRandomize
+                ? connectUrls.Distinct().ToArray()
+                : connectUrls.OrderBy(_ => Guid.NewGuid()).Distinct().ToArray();
 
             // add last.
```
caleblloyd, line 545: Nit but is it a little cleaner if we drop the ternary?

```suggestion
            var urls = serverReportedUrls.Concat(Opts.GetSeedUris()).Distinct();
            if (!Opts.NoRandomize)
            {
                urls = urls.OrderBy(_ => Guid.NewGuid());
            }

            // Ensure the current URL is last in the list
```

mtmk: nice!

caleblloyd (approved):
LGTM

PR #485, opened 2024-04-19
Resolves #476 - watch multiple kv filters

### LINKED ISSUE #476: KV watch multiple filters
* https://github.com/nats-io/nats-architecture-and-design/issues/277

rickdotnet:
Resolves #476 

> KV watch supports only single filter, which has limitations similar to those of single filter consumers.
> 
> Clients should support kv::watch watching multiple filters, leveraging consumers with multiple filters.
> 
> That can be a separate method (for example watch_many), or extension to the current one, depending on the language and it's ability to introduce multiple KV watch filters without breaking changes.

Would love for someone to do a more comprehensive safety check. I'm venturing into bits of the client that I haven't used yet. I looked at the v1 implementation and [this comment](https://github.com/nats-io/nats.net/pull/887/files/9ad54302c8a442a54eb8fac5e65138dbcca3e18a#r1559883388) had me scratching my head. It didn't seem to affect the tests I ran. So I assume this is handled differently in v2.

### DIFF
--- a/src/NATS.Client.KeyValueStore/INatsKVStore.cs
+++ b/src/NATS.Client.KeyValueStore/INatsKVStore.cs
@@ -72,7 +72,7 @@ public interface INatsKVStore
     ValueTask<NatsKVEntry<T>> GetEntryAsync<T>(string key, ulong revision = default, INatsDeserialize<T>? serializer = default, CancellationToken cancellationToken = default);
 
     /// <summary>
-    /// Start a watcher for specific keys
+    /// Start a watcher for a specific key
     /// </summary>
     /// <param name="key">Key to watch (subject-based wildcards may be used)</param>
     /// <param name="serializer">Serializer to use for the message type.</param>
@@ -81,6 +81,17 @@ public interface INatsKVStore
     /// <typeparam name="T">Serialized value type</typeparam>
     /// <returns>An asynchronous enumerable which can be used in <c>await foreach</c> loops</returns>
     IAsyncEnumerable<NatsKVEntry<T>> WatchAsync<T>(string key, INatsDeserialize<T>? serializer = default, NatsKVWatchOpts? opts = default, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Start a watcher for specific keys
+    /// </summary>
+    /// <param name="keys">Keys to watch (subject-based wildcards may be used)</param>
+    /// <param name="serializer">Serializer to use for the message type.</param>
+    /// <param name="opts">Watch options</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
+    /// <typeparam name="T">Serialized value type</typeparam>
+    /// <returns>An asynchronous enumerable which can be used in <c>await foreach</c> loops</returns>
+    IAsyncEnumerable<NatsKVEntry<T>> WatchAsync<T>(IList<string> keys, INatsDeserialize<T>? serializer = default, NatsKVWatchOpts? opts = default, CancellationToken cancellationToken = default);
 
     /// <summary>
     /// Start a watcher for all the keys in the bucket
--- a/src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
+++ b/src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
@@ -35,7 +35,7 @@ internal class NatsKVWatcher<T> : IAsyncDisposable
     private readonly NatsSubOpts? _subOpts;
     private readonly CancellationToken _cancellationToken;
     private readonly string _keyBase;
-    private readonly string _filter;
+    private readonly IList<string> _filters;
     private readonly NatsConnection _nats;
     private readonly Channel<NatsKVWatchCommandMsg<T>> _commandChannel;
     private readonly Channel<NatsKVEntry<T>> _entryChannel;
@@ -55,7 +55,7 @@ internal class NatsKVWatcher<T> : IAsyncDisposable
     public NatsKVWatcher(
         NatsJSContext context,
         string bucket,
-        string key,
+        IList<string> keys,
         INatsDeserialize<T> serializer,
         NatsKVWatchOpts opts,
         NatsSubOpts? subOpts,
@@ -69,7 +69,7 @@ internal class NatsKVWatcher<T> : IAsyncDisposable
         _opts = opts;
         _subOpts = subOpts;
         _keyBase = $"$KV.{_bucket}.";
-        _filter = $"{_keyBase}{key}";
+        _filters = keys.Select(key => $"{_keyBase}{key}").ToList();
         _cancellationToken = cancellationToken;
         _nats = context.Connection;
         _stream = $"KV_{_bucket}";
@@ -358,7 +358,7 @@ internal class NatsKVWatcher<T> : IAsyncDisposable
             DeliverPolicy = ConsumerConfigDeliverPolicy.All,
             AckPolicy = ConsumerConfigAckPolicy.None,
             DeliverSubject = _sub.Subject,
-            FilterSubject = _filter,
+            FilterSubjects = _filters,
             FlowControl = true,
             IdleHeartbeat = _opts.IdleHeartbeat,
             AckWait = TimeSpan.FromHours(22),
--- a/tests/NATS.Client.KeyValueStore.Tests/NatsKVWatcherTest.cs
+++ b/tests/NATS.Client.KeyValueStore.Tests/NatsKVWatcherTest.cs
@@ -30,7 +30,7 @@ public class NatsKVWatcherTest
         var js2 = new NatsJSContext(nats2);
         var kv2 = new NatsKVContext(js2);
         var store2 = (NatsKVStore)await kv2.CreateStoreAsync(config, cancellationToken: cancellationToken);
-        var watcher = await store2.WatchInternalAsync<NatsMemoryOwner<byte>>("k1.*", cancellationToken: cancellationToken);
+        var watcher = await store2.WatchInternalAsync<NatsMemoryOwner<byte>>(["k1.*"], cancellationToken: cancellationToken);
 
         await store1.PutAsync("k1.p1", 1, cancellationToken: cancellationToken);
         await store1.PutAsync("k1.p1", 2, cancellationToken: cancellationToken);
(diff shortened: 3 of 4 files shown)


mtmk:
Thanks @rickdotnet 💯 

> Would love for someone to do a more comprehensive safety check. I'm venturing into bits of the client that I haven't used yet. I looked at the v1 implementation and [this comment](https://github.com/nats-io/nats.net/pull/887/files/9ad54302c8a442a54eb8fac5e65138dbcca3e18a#r1559883388) had me scratching my head. It didn't seem to affect the tests I ran. So I assume this is handled differently in v2.

We don't follow the v1 client (style and implementation details wise) since it's pretty much a direct port from the Java client. Of course functionality should be the same.

mtmk:
Looks like nats-server 2.9.x isn't happy:

```
NATS.Client.KeyValueStore.Tests.GetKeysTest.Get_keys_should_not_hang_when_there_are_deleted_keys [FAIL]
[xUnit.net 00:00:01.14]       NATS.Client.JetStream.NatsJSApiException : consumer delivery policy is deliver last per subject, but optional filter subject is not set
```

https://github.com/nats-io/nats.net.v2/actions/runs/8747098190/job/24005035527?pr=485#step:9:21

src/NATS.Client.KeyValueStore/INatsKVStore.cs
```diff
@@ -72,7 +72,7 @@ public interface INatsKVStore
     ValueTask<NatsKVEntry<T>> GetEntryAsync<T>(string key, ulong revision = default, INatsDeserialize<T>? serializer = default, CancellationToken cancellationToken = default);
 
     /// <summary>
-    /// Start a watcher for specific keys
+    /// Start a watcher for a specific key
```
mtmk, line 75: I think plural is still correct here since a wildcard may be used.

src/NATS.Client.KeyValueStore/INatsKVStore.cs
```diff
@@ -82,6 +82,17 @@ public interface INatsKVStore
...
+    /// <summary>
+    /// Start a watcher for specific keys
+    /// </summary>
+    /// <param name="keys">Keys to watch (subject-based wildcards may be used)</param>
+    /// <param name="serializer">Serializer to use for the message type.</param>
+    /// <param name="opts">Watch options</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
+    /// <typeparam name="T">Serialized value type</typeparam>
+    /// <returns>An asynchronous enumerable which can be used in <c>await foreach</c> loops</returns>
+    IAsyncEnumerable<NatsKVEntry<T>> WatchAsync<T>(IList<string> keys, INatsDeserialize<T>? serializer = default, NatsKVWatchOpts? opts = default, CancellationToken cancellationToken = default);
```
mtmk, line 94: IList smells like Java :)

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
```diff
@@ -358,7 +358,7 @@ private async ValueTask<INatsJSConsumer> CreatePushConsumer(string origin)
             DeliverPolicy = ConsumerConfigDeliverPolicy.All,
             AckPolicy = ConsumerConfigAckPolicy.None,
             DeliverSubject = _sub.Subject,
-            FilterSubject = _filter,
+            FilterSubjects = _filters,
```
mtmk, line 361: filter-subjects (plural) was introduced in server 2.10. to keep 2.9 compat we should still use filter-subject (singular) if there is only one filter subject. (which should fix the ci failure)

tests/NATS.Client.KeyValueStore.Tests/NatsKVWatcherTest.cs
```diff
@@ -140,6 +140,64 @@ await foreach (var entry in store.WatchAsync<string>("*", cancellationToken: can
...
+        var signal = new WaitSignal(timeout);
+        var watchTask = Task.Run(
+            async () =>
+            {
+                await foreach (var entry in store.WatchAsync<string>(["k1", "k2"], cancellationToken: cancellationToken))
+                {
+                    signal.Pulse();
+                    _output.WriteLine($"WATCH: {entry.Key} ({entry.Revision}): {entry.Value}");
+
+                    Assert.True(entry.Key is "k1" or "k2");
```
mtmk, line 172: we're not checking a couple of cases here:
1) we should see k1
2) we should not see k3

note we can also break on `entry.Delta == 0` to end the watch if that helps.

rickdotnet: I originally had it as `! "k3"`, but thought `"k1" or "k2"` read better. I did end up adding "k1" after a commit, but it looks like it's being checked here.

Will look at this again though.

src/NATS.Client.KeyValueStore/INatsKVStore.cs
rickdotnet: I adjusted this back. Although, it reads goofy to me. The param is called "key" singular, but the comment says "keys" to represent a wildcard. I have no preference, but my OCD is triggered.

tests/NATS.Client.KeyValueStore.Tests/NatsKVWatcherTest.cs
rickdotnet: Looks like I'll need to do more than just copy/paste and rework that watcher test. Seems to be a new exception for 2.9 now.

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
rickdotnet: https://github.com/rickdotnet/nats.net.v2/blob/7548b7481ff2cb4767cdf577563a83a4438f7237/src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs#L370

tests/NATS.Client.KeyValueStore.Tests/NatsKVWatcherTest.cs
mtmk: I think we can ignore this test for 2.9:
```csharp
[SkipIfNatsServer(versionEarlierThan: "2.10")]
public async Task Watch_subset()
```

mtmk: just tweaked the test a little to make sure we received both k1 and k2.

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
```diff
@@ -368,6 +367,18 @@ private async ValueTask<INatsJSConsumer> CreatePushConsumer(string origin)
             ReplayPolicy = ConsumerConfigReplayPolicy.Instant,
         };
 
+        // nat2.9.x
+        if (_filters.Length == 1)
```
mtmk, line 371: if this is not the case and server version is <2.10 should we error?

src/NATS.Client.KeyValueStore/INatsKVStore.cs
mtmk: `IEnumerable` is a good idea.

mtmk:
@rickdotnet, @caleblloyd, I wanted to have a quick discussion about the new interface where we accept multiple keys (see above). wdyt?

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
```diff
@@ -368,6 +367,18 @@ private async ValueTask<INatsJSConsumer> CreatePushConsumer(string origin)
...
+        {
+            config.FilterSubject = _filters[0];
+        }
+
+        // nat2.10.x
+        if (_filters.Length > 1)
+        {
+            config.FilterSubjects = _filters;
+        }
+
```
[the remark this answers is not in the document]
rickdotnet: Since it's internal, I thought about throwing an exception in the [Watch()](https://github.com/rickdotnet/nats.net.v2/blob/19da2112b0190454f52e76d59ec011ef75b33204/src/NATS.Client.KeyValueStore/NatsKVStore.cs#L324) method in `NatsKVStore` since that will be closer to the caller. That'd be my vote.

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
rickdotnet: Are there any NATS policies on how many versions you want to support? I ask, because if 2.11 is around the corner, maybe it's a moot point. I think for convenience it'd be nice to at least log an error, but since there is a chance that the functionality will differ from expectations, I think an exception makes sense.

If one of y'all end up adding the additional exception handling, please feel free to fix my two typos (`// nat2.9.x`). My OCD is triggered.

caleblloyd: Technically the NATS Server hosting JS must be 2.10, not necessarily the NATS Server it is connected to. I think if multiple filter subjects are passed and the server doesn't support it, it will see that as no filter subjects and throw an error, which could be confusing though.

caleblloyd (commented):
API looks good to me!

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
mtmk: > Are there any NATS policies on how many versions you want to support? I ask, because if 2.11 is around the corner, maybe it's a moot point.

I think we should stick to >=2.9 for the time being but you're right we will have to think about it as versions go up.

> Technically the NATS Server hosting JS must be 2.10, not necessarily the NATS Server it is connected to

good point. so there is not use trying to check server version.

src/NATS.Client.KeyValueStore/INatsKVStore.cs
mtmk: maybe we keep your edit and add something in parenthesis or a remark that wildcards can be used?

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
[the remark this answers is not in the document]
mtmk: I put a check deep down in the watcher class because that's where the enumeration materializes.

caleblloyd (approved):
LGTM

mtmk (approved):
LGTM thank you @rickdotnet 💯

PR #488, opened 2024-04-23
Fix NUID sequence

mtmk:
Fixed nuid sequence so the last few digits change similar to Go reference implementation.

there is also a report that Service ids aren't being generated uniquely which relies on NUIDs. I was not able to reproduce this but reminded me an old test fix I made (` DifferentThreads_DifferentPrefixes()` which is now reverted) which was failing intermittently due to id clashes.

cc @jasper-d

### DIFF
--- a/src/NATS.Client.Core/Internal/NuidWriter.cs
+++ b/src/NATS.Client.Core/Internal/NuidWriter.cs
@@ -63,8 +63,10 @@ internal sealed class NuidWriter
         // NOTE: We must never write to digitsPtr!
         ref var digitsPtr = ref MemoryMarshal.GetReference(Digits);
 
-        for (nuint i = PrefixLength; i < NuidLength; i++)
+        // write backwards so the last two characters change the fastest
+        for (nuint i = NuidLength; i > PrefixLength;)
         {
+            i--;
             var digitIndex = (nuint)(sequential % Base);
             Unsafe.Add(ref buffer[0], i) = Unsafe.Add(ref digitsPtr, digitIndex);
             sequential /= Base;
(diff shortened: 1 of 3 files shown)


aachinfiev:
@mtmk  as per your request, I was getting this issue using Windows 11, .Net 8, and C# client 2.2.1, NATS server 2.10.14. Had issues in earlier versions too. This is the latest one I have tested and still had a problem.

mtmk:
thanks @aachinfiev I also have the same setup but unable to reproduce it. Is this `[Service_ids_unique()](https://github.com/nats-io/nats.net.v2/pull/488/files#diff-0f31823345f059a336df462dbc653d55bc58db73d44487cb46393d52e15c579eR355)` test close to what your code where you're seeing the issue?

aachinfiev:
I don't dispose the service in the same await. Mine is more like this:

In ExecuteAsync of a BackgroundService I do 

```csharp
var ctx = new NatsSvcContext(_nats);

var svc = await ctx.AddServiceAsync("svc1", "1.0.0", "svc1");
_svcSubs.Add(svc);
// add endpoints for svc1
svc = await ctx.AddServiceAsync("svc2", "1.0.0", "svc2");
_svcSubs.Add(svc);
// add endpoints for svc2
```

And in StopAsync I try to dipose all svcSubs

aachinfiev:
It's working for me correctly now as well. I think problem was that the middle single character was changing and I didn't spot that Ids were different because front/back were the same. With this PR making last character different it's more noticeable. Thank you.

caleblloyd (approved):
LGTM

PR #489, opened 2024-04-24
FIX : PurgeDeletesAsync hangs if KV Bucket is already empty

darkwatchuk:
NatsKVStore.PurgeDeletesAsync hangs if KV Bucket is already empty.

Also updated Purge_deletes test


mtmk (approved):
LGTM thanks @darkwatchuk

caleblloyd (approved):

PR #490, opened 2024-04-24
Change ConsumerConfig AckPolicy default to Explicit

hanlong-chen-1047:
Currently the no-args constructor for `ConsumerConfig` does not set the `AckPolicy` for the consumer, which defaults it to `None`. This is confusing given that the NATS docs state that the default `AckPolicy` is `Explicit`:
https://docs.nats.io/nats-concepts/jetstream/consumers#ackpolicy

This has led to unexpected behaviors since we had expected this to be set to `Explicit` by default for all consumers.

This PR changes the no-args `ConsumerConfig` constructor to set the `AckPolicy` to `Explicit`, to be the same as the constructor with a durable name.

### DIFF
--- a/sandbox/Example.JetStream.PullConsumer/Program.cs
+++ b/sandbox/Example.JetStream.PullConsumer/Program.cs
@@ -19,7 +19,7 @@ await using var nats = new NatsConnection(options);
 
 var js = new NatsJSContext(nats);
 
-var consumer = await js.CreateOrUpdateConsumerAsync("s1", new ConsumerConfig { Name = "c1", DurableName = "c1", AckPolicy = ConsumerConfigAckPolicy.Explicit });
+var consumer = await js.CreateOrUpdateConsumerAsync("s1", new ConsumerConfig { Name = "c1", DurableName = "c1" });
 
 var idle = TimeSpan.FromSeconds(5);
 var expires = TimeSpan.FromSeconds(10);
--- a/src/NATS.Client.JetStream/Models/ConsumerConfig.cs
+++ b/src/NATS.Client.JetStream/Models/ConsumerConfig.cs
@@ -29,7 +29,6 @@ public record ConsumerConfig
     {
         Name = name;
         DurableName = name;
-        AckPolicy = ConsumerConfigAckPolicy.Explicit;
     }
 
     [System.Text.Json.Serialization.JsonPropertyName("deliver_policy")]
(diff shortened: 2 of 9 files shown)


mtmk:
This sounds like a sensible thing to do actually ([potentially related discussions](https://github.com/nats-io/nats.net.v2/discussions?discussions_q=consumerconfig))

cc @robertmircea

caleblloyd:
Should we consider changing the default value of the enum `ConsumerConfigAckPolicy` to `Explicit` instead of `None`?

mtmk:
> Should we consider changing the default value of the enum `ConsumerConfigAckPolicy` to `Explicit` instead of `None`?

it might make more sense for other scenarios as well e.g.
```csharp
    AckPolicy = default,
```

robertmircea:
Fine with me. I was setting it to Explicit every time

mtmk:
looking at this closer enum:
```csharp
public enum ConsumerConfigAckPolicy
{
    Explicit = 0, // <--- hurts my eyes
    All = 1,
    None = 2,
}
```
...having `None = 0` is more conventional and it's aligned with how the server is working on the wire.

hanlong-chen-1047:
> > Should we consider changing the default value of the enum `ConsumerConfigAckPolicy` to `Explicit` instead of `None`?
> 
> it might make more sense for other scenarios as well e.g.
> 
> ```cs
>     AckPolicy = default,
> ```
> 
> **Edit:** @hanlong-chen-1047 are you also happy with this suggestion? (I can make the change if you don't have time, not to worry)

yeah that'd be great

caleblloyd:
There are some usages / examples now that no longer need to explicitly set the Ack policy to Explicit

https://github.com/nats-io/nats.net.v2/blob/f6c6419f34741dc9dabfbe400c4861d25901a579/tests/NATS.Client.JetStream.Tests/JetStreamTest.cs#L81

https://github.com/nats-io/nats.net.v2/blob/f6c6419f34741dc9dabfbe400c4861d25901a579/tests/NATS.Client.CheckNativeAot/Program.cs#L91

https://github.com/nats-io/nats.net.v2/blob/f6c6419f34741dc9dabfbe400c4861d25901a579/tests/NATS.Client.Core.MemoryTests/NatsConsumeTests.cs#L32

https://github.com/nats-io/nats.net.v2/blob/f6c6419f34741dc9dabfbe400c4861d25901a579/tests/NATS.Net.DocsExamples/JetStream/ManagingPage.cs#L73

https://github.com/nats-io/nats.net.v2/blob/f6c6419f34741dc9dabfbe400c4861d25901a579/src/NATS.Client.JetStream/Models/ConsumerConfig.cs#L32

caleblloyd (approved):
LGTM

PR #464, opened 2024-04-04
Run core tests on Windows and test run stability

mtmk:
Fixing some of the core test reliability issues and improving test coverage on Windows. Because test `NatsServer` seems to be a little heavy, chopped up the `Core` tests into `Core2` and `CoreUnit` tests. `Core2` is now using a lighter test NATS server process and `CoreUnit` doesn't have any need for running the server.

Also, fixed a few build warnings mainly around package security warnings for utility projects (not the public libraries); organized solution files.

* fixes #579

PR #491, opened 2024-04-24
Add ResumeAtRevision support to KV Watcher

### LINKED ISSUE #478: KV ResumeAtRevision option for watches
* https://github.com/nats-io/nats-architecture-and-design/issues/261

niklasfp:
Resolves #478

### DIFF
--- a/src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
+++ b/src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
@@ -383,10 +383,14 @@ internal class NatsKVWatcher<T> : IAsyncDisposable
             config.HeadersOnly = true;
         }
 
-        if (sequence > 0)
+        // Resume from a specific revision ?
+        if (sequence > 0 || _opts.ResumeAtRevision > 0)
         {
             config.DeliverPolicy = ConsumerConfigDeliverPolicy.ByStartSequence;
-            config.OptStartSeq = sequence + 1;
+
+            // If Sequence is set, it means that the consumer is being recreated, and we should start
+            // from the next sequence, otherwise we should use the revision specified in the options.
+            config.OptStartSeq = sequence > 0 ? sequence + 1 : _opts.ResumeAtRevision;
         }
 
         var consumer = await _context.CreateOrUpdateConsumerAsync(
--- a/src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
+++ b/src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
@@ -384,13 +384,15 @@ internal class NatsKVWatcher<T> : IAsyncDisposable
         }
 
         // Resume from a specific revision ?
-        if (sequence > 0 || _opts.ResumeAtRevision > 0)
+        if (sequence > 0)
         {
             config.DeliverPolicy = ConsumerConfigDeliverPolicy.ByStartSequence;
-
-            // If Sequence is set, it means that the consumer is being recreated, and we should start
-            // from the next sequence, otherwise we should use the revision specified in the options.
-            config.OptStartSeq = sequence > 0 ? sequence + 1 : _opts.ResumeAtRevision;
+            config.OptStartSeq = sequence + 1;
+        }
+        else if (_opts.ResumeAtRevision > 0)
+        {
+            config.DeliverPolicy = ConsumerConfigDeliverPolicy.ByStartSequence;
+            config.OptStartSeq = _opts.ResumeAtRevision;
         }
 
         var consumer = await _context.CreateOrUpdateConsumerAsync(
--- a/src/NATS.Client.KeyValueStore/NatsKVOpts.cs
+++ b/src/NATS.Client.KeyValueStore/NatsKVOpts.cs
@@ -48,7 +48,7 @@ public record NatsKVWatchOpts
     /// Setting this to a non-zero value will cause the watcher to ignore the values for <see cref="IncludeHistory"/> and <see cref="UpdatesOnly"/>.
     /// </remarks>
     /// </summary>
-    public ulong ResumeAtRevision { get; set; }
+    public ulong ResumeAtRevision { get; init; }
 }
 
 public record NatsKVDeleteOpts
--- a/src/NATS.Client.KeyValueStore/NatsKVOpts.cs
+++ b/src/NATS.Client.KeyValueStore/NatsKVOpts.cs
@@ -45,7 +45,7 @@ public record NatsKVWatchOpts
     /// <summary>
     /// The revision to start from, if set to 0 (default) this will be ignored.
     /// <remarks>
-    /// Setting this to a non-zero value will cause the watcher to ignore the values for <see cref="IncludeHistory"/> and <see cref="IncludeHistory"/>.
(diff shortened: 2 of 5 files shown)


mtmk:
> @mtmk I'm seeing some test flappers in Watcher_reconnect_with_history and Watcher_timeout_reconnect locally (also without my changes)
> 
> Machine: Mac M2 Pro arm64

I'm not surprised. they're most stable on Linux at the moment. We also have #464 to deal with Windows flappers. we might open one for Mac too. I'm guessing it's mostly around managing the server process and sockets, but I might be wrong.

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
```diff
@@ -383,10 +383,14 @@ private async ValueTask<INatsJSConsumer> CreatePushConsumer(string origin)
             config.HeadersOnly = true;
         }
 
-        if (sequence > 0)
+        // Resume from a specific revision ?
+        if (sequence > 0 || _opts.ResumeAtRevision > 0)
```
mtmk, line 387: do we still need to do these if we assigned `_sequenceStream = _opts.ResumeAtRevision` in the constructor?

src/NATS.Client.KeyValueStore/NatsKVOpts.cs
```diff
@@ -41,6 +41,14 @@ public record NatsKVWatchOpts
...
     /// </summary>
     public Func<CancellationToken, ValueTask<bool>>? OnNoData { get; init; }
+
+    /// <summary>
+    /// The revision to start from, if set to 0 (default) this will be ignored.
+    /// <remarks>
+    /// Setting this to a non-zero value will cause the watcher to ignore the values for <see cref="IncludeHistory"/> and <see cref="UpdatesOnly"/>.
+    /// </remarks>
+    /// </summary>
+    public ulong ResumeAtRevision { get; set; }
```
mtmk, line 51: Do we need opts to be mutable? Can we use `init` property setter?

src/NATS.Client.KeyValueStore/Internal/NatsKVWatcher.cs
mtmk: Ah! you're spot on, apologies. can we simplify the logic so sequence only checked once? e.g.:
```csharp
if (sequence > 0)
{
    config.DeliverPolicy = ConsumerConfigDeliverPolicy.ByStartSequence;
    config.OptStartSeq = sequence + 1;
}
else if (_opts.ResumeAtRevision > 0)
{
    config.DeliverPolicy = ConsumerConfigDeliverPolicy.ByStartSequence;
    config.OptStartSeq = _opts.ResumeAtRevision;
}
```

mtmk (approved):
LGTM thank you @niklasfp 👍

src/NATS.Client.KeyValueStore/NatsKVOpts.cs
```diff
@@ -41,6 +41,14 @@ public record NatsKVWatchOpts
     /// Async function called when the enumerator reaches the end of data. Return True to break the async enumeration, False to allow the enumeration to continue.
     /// </summary>
     public Func<CancellationToken, ValueTask<bool>>? OnNoData { get; init; }
+
+    /// <summary>
+    /// The revision to start from, if set to 0 (default) this will be ignored.
+    /// <remarks>
+    /// Setting this to a non-zero value will cause the watcher to ignore the values for <see cref="IncludeHistory"/> and <see cref="UpdatesOnly"/>.
+    /// </remarks>
```
caleblloyd, line 49: Is it more consistent/correct to annotate that this option silently overrides other options?  Or should we fail with an Exception that states that two options that are mutually exclusive have been set?

caleblloyd: Probably whichever pattern we follow in Consumer Creation is the same pattern we would want to follow here

caleblloyd: On second thought it might be a little abrasive to throw here, since `IncludeHistory` will default to `false` but when setting `ResumeAtRevision` it _must_ be true.  I think it makes sense to ignore `IncludeHistory` when `ResumeAtRevision` is set.

`UpdatesOnly = true` and `ResumeAtRevision > 0` are 2 non-defaults though so I think it makes sense to throw when those are both set.

[Regular Consumer Crerate](https://github.com/nats-io/nats.net.v2/blob/8952bc8bd1a0e60c1848eabd4e4f52326c1de4ef/src/NATS.Client.JetStream/NatsJSContext.Consumers.cs#L203-L225) isn't checking any pre-conditions it looks like, it's taking the Deliver Policy and Start Seq / Start Time separately.  So if the Deliver Policy doesn't match the Server will fail with an error, and an exception will be thrown.

caleblloyd: `ArgumentException` is consistent with the library

mtmk: Calling GetKeys with the resume parameter might be a valid use case. I agree that including history is overridden due to the method overload. However, considering the following points, I suggest keeping it in the options:

1) Cross-cutting functionality: resume seems to be applicable to multiple scenarios and not specific to a single use case. Including it in the options would make it more reusable and accessible across different parts of the codebase.

2) Consistency with other NATS clients: It's important to maintain a certain degree of consistency with other client implementations, unless there are .NET specific differences.

3) Impact on other clients: Altering the method signature or introducing overloads would require updating other client implementations as well. This can be a significant effort, and there should be a strong justification for making such changes unless we can say this is a .NET specific difference.

mtmk: I was checking Go and JS implementations https://github.com/nats-io/nats.go/pull/1489/files https://github.com/nats-io/nats.deno/pull/651/files .. so we're not that consistent after all😄

I guess we can add overloads if you think it's going to be a better developer experience. it just feels like a less generic solution to me. wdyt @niklasfp @caleblloyd

caleblloyd: I am also in favor of leaving it in Options and adding Validation.

If the Validation needs to run from multiple callers, we could add a `internal void ApplyOptsToConsumerConfig(ConsumerConfig config)` method on the Options, which performs validation, then mutates a ConsumerConfig?

We do something similar [in NatsTlsOpts](https://github.com/nats-io/nats.net.v2/blob/df478c27425a12c786f604fd9da8b4f9c83e71a3/src/NATS.Client.Core/NatsTlsOpts.cs#L163-L172) - where validation is performed at the time of use

mtmk:
@niklasfp, @caleblloyd just made the last bit of validation changes. Please feel free to make suggestion, especially for docs. Let me know if y'all happy with it then I can merge it ;)

mtmk:
> > @niklasfp, @caleblloyd just made the last bit of validation changes. Please feel free to make suggestion, especially for docs. Let me know if y'all happy with it then I can merge it ;)
> 
> Omg thank u, and so sorry for not following up on this, life happened 🥲 I'll take a look today.

thanks @niklasfp very kind of you 🙏 honestly, we really appreciate your contribution, you've done the hard work and I'm always happy to tie the loose ends when you don't have time 😅

caleblloyd (approved):
LGTM

Issue #475, opened 2024-04-16
Consumer Action API


mtmk:
### Proposed change

https://github.com/nats-io/nats-architecture-and-design/issues/276

### Use case

x

Ivandemidov00:
Hi, @mtmk can I make this change?

mtmk:
thanks @Ivandemidov00 that'd be awesome. go for it!

PR #497, opened 2024-05-11
Add support for Consumer Create Action

Ivandemidov00:
Resolves [#475](https://github.com/nats-io/nats.net.v2/issues/475)


Ivandemidov00:
Fixed problem with serialize .net8 and test on create publisher action

src/NATS.Client.JetStream/Models/ConsumerCreateRequest.cs
```diff
@@ -26,4 +28,13 @@ internal record ConsumerCreateRequest
...
     public ConsumerConfig? Config { get; set; }
+
+    [System.Text.Json.Serialization.JsonPropertyName("action")]
+    [System.Text.Json.Serialization.JsonIgnore(Condition = System.Text.Json.Serialization.JsonIgnoreCondition.Never)]
+#if NET6_0
+    [System.Text.Json.Serialization.JsonConverter(typeof(NatsJSJsonStringEnumConverter<ConsumerCreateRequestAction>))]
+#else
+    [System.Text.Json.Serialization.JsonConverter(typeof(NatsJSJsonStringEnumConverter<ConsumerCreateRequestAction>))]
+#endif
+    public ConsumerCreateRequestAction Action { get; set; }
```
mtmk, line 39: Could we make this optional so that we don't have to deal with net8 serialization exception for an empty value?

src/NATS.Client.JetStream/Models/ConsumerCreateRequestAction.cs
```diff
@@ -0,0 +1,11 @@
+using System.Runtime.Serialization;
+
+namespace NATS.Client.JetStream.Models;
+
+public enum ConsumerCreateRequestAction
+{
+    Create = 0,
+    Update = 1,
+    [EnumMember(Value = "")]
+    CreateOrUpdate = 2,
```
mtmk, line 10: we can remove this option if action is optional.

src/NATS.Client.JetStream/Internal/NatsJSJsonSerializer.cs
```diff
@@ -314,6 +327,71 @@ public override void Write(Utf8JsonWriter writer, TEnum value, JsonSerializerOpt
...
+                writer.WriteStringValue(string.Empty);
+                return;
+            }
+        }
+
+        throw new InvalidOperationException($"Writing unknown enum value {value.GetType().Name}.{value}");
+    }
+}
+#else
+internal class NatsJSJsonStringEnumConverter<TEnum> : JsonConverter<TEnum>
```
mtmk, line 350: is this for net8? I guess we can leave it as camelCase and not need this class?

src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
```diff
@@ -33,31 +33,25 @@ public partial class NatsJSContext : INatsJSContext
         CancellationToken cancellationToken = default)
     {
         ThrowIfInvalidStreamName(stream);
+        return await CreateOrUpdateConsumerInternalAsync(stream, config, ConsumerCreateRequestAction.CreateOrUpdate, cancellationToken);
```
mtmk, line 36: we could make action nullable here.

src/NATS.Client.JetStream/Models/ConsumerCreateRequestAction.cs
```diff
@@ -0,0 +1,11 @@
+using System.Runtime.Serialization;
+
+namespace NATS.Client.JetStream.Models;
+
+public enum ConsumerCreateRequestAction
```
mtmk, line 5: what do you think of `ConsumerCreateAction` as the name? more succinct?

mtmk (commented):
looks good over all, just a suggestion about removing CreateOrUpdate from the enum since absence of the action field would mean the same thing as an empty string.

src/NATS.Client.JetStream/Models/ConsumerCreateRequestAction.cs
Ivandemidov00: Yes, think it's good too. Renamed

src/NATS.Client.JetStream/Models/ConsumerCreateRequest.cs
Ivandemidov00: Yes, that was one of the ideas. In this case, the property will sample be ignored and `NatsJSJsonStringEnumConverter` will be less strict

mtmk:
@Ivandemidov00 thanks for the updates. Just made a change to enum default and removed the nullable marking. I think it's easier to reason with this way.

mtmk (approved):
LGTM thanks @Ivandemidov00 
(will wait for another review because of my changes)

src/NATS.Client.JetStream/Models/ConsumerCreateAction.cs
```diff
@@ -0,0 +1,8 @@
+namespace NATS.Client.JetStream.Models;
+
+public enum ConsumerCreateAction
+{
+    Default = 0,
```
caleblloyd, line 5: Would it make sense to make the default descriptive?  I.e. `CreateOrUpdate = 0` instead of `Default = 0`?

Or is it best to leave it vague as the NATS Server can decide the default?

caleblloyd (approved):
LGTM

PR #498, opened 2024-05-13
Update NuidWriter test to check the last four digits

mtmk:
Saw a change of three digits failing test. There maybe a few more than two changing depending on the increment.


caleblloyd (approved):
LGTM

PR #502, opened 2024-05-29
[PROPOSAL] Add overload for empty requests

mtmk:
Proposing a convenience method where request payload isn't required in request-reply calls. It is fairly common to request data without sending any payload. With this method application code can avoid boilerplate null data parameter and request type parameter making the call much more succinct.

```csharp
// current way of sending an empty request payload
var reply1 = await nats.RequestAsync<object, int>(subject, data: null);

// proposed API. calls the method above
var reply2 = await nats.RequestAsync<int>(subject);
```


caleblloyd (approved):
Seems valuable!  LGTM

PR #508, opened 2024-06-05
Update consume backpressure handling

### LINKED ISSUE #608: MaxMsg=1 not respected. second message ALWAYS gets redelivered as if its ackwait started before yielded
MaxMsg=1 not respected. second message ALWAYS gets redelivered when ack'ed within time, it is  as if its ackwait started before yielded (as if MaxMsg 1 wasn't originally respected and this is was in the buffer before the loop back to foreach). Theory only. Reproducible consistently.  I get the point that one should handle  redeliveries via idempotency onwards or some KV lookup to check, due to anything from network partitions, server restarts, client crashes, internal buffers being too strained, and so on, but this is consistently reproducible locally. 

### Observed behavior (local setup VM and client app next to each other, no network chat)

```using NATS.Client.Core;
using NATS.Client.JetStream;

Console.WriteLine("Start");

var c = new NatsConnection(new NatsOpts()
{
    Url = "nats://127.0.0.1:4222"
});

var s = new NatsJSContext(c);
var consumer = await s.GetConsumerAsync("email-queue", "email-queue-consumer");

await foreach(var i in consumer.ConsumeAsync<string>(opts: new NatsJSConsumeOpts() {
                  MaxMsgs = 1
})) {
    Console.WriteLine("TS" + DateTime.UtcNow + "   -> " + i.Data);
    await Task.Delay(32500); // ack wait configured to 36second on consumer config
    await i.AckAsync();
}
Console.WriteLine("Done");

```

consumer
```
Configuration:

                    Name: email-queue-consumer
               Pull Mode: true
          Deliver Policy: All
              Ack Policy: Explicit
                Ack Wait: 36.00s
           Replay Policy: Instant
      Maximum Deliveries: 2
         Max Ack Pending: 1,000
       Max Waiting Pulls: 512

State:

  Last Delivered Message: Consumer sequence: 20 Stream sequence: 14 Last delivery: 1m49s ago
    Acknowledgment Floor: Consumer sequence: 20 Stream sequence: 14 Last Ack: 40.49s ago
        Outstanding Acks: 0 out of maximum 1,000
    Redelivered Messages: 0
    Unprocessed Messages: 0
           Waiting Pulls: 1 of maximum 512
```

```
➜  ~ nats stream info email-queue
Information for Stream email-queue created 2024-07-18 12:13:40

              Subjects: email.queue
              Replicas: 1
               Storage: File

Options:

             Retention: WorkQueue
       Acknowledgments: true
        Discard Policy: Old
      Duplicate Window: 2m0s
     Allows Msg Delete: true
          Allows Purge: true
        Allows Rollups: false

Limits:

      Maximum Messages: unlimited
   Maximum Per Subject: unlimited
         Maximum Bytes: unlimited
           Maximum Age: 1d0h0m0s
  Maximum Message Size: unlimited
     Maximum Consumers: unlimited

State:

              Messages: 0
                 Bytes: 0 B
        First Sequence: 17
         Last Sequence: 16 @ 2024-08-22 12:09:39 UTC
      Active Consumers: 1

```

Once setup run this:

```
nats publish email.queue 'ahhh1'
nats publish email.queue 'ahhh2'
```
### Expected behavior

shouldnt see second message 2 times. as acked within 32.5 seconds when ackwait is 36s

### Server and client version

v2.10.17 and    "NATS.Net" Version="2.3.3". All local Fedora 40.  Dotnet 8. Plenty of resources.  No docker.

### Host environment

linux, localhost, all same machine

### Steps to reproduce

setup stream and consumer via cli
build and start dotnet simple app
nats publish email.queue 'ahhh1'
nats publish email.queue 'ahhh2'

mtmk:
Moved pull request calculations to when the messages are yielded to calculate if we need to pull more messages accurately to prevent pulling excessive messages from the server. It also has been bounded to a maximum of 1024 to avoid large object heap allocations.

* resolves #608

### DIFF
--- a/tests/NATS.Client.JetStream.Tests/ConsumerConsumeTest.cs
+++ b/tests/NATS.Client.JetStream.Tests/ConsumerConsumeTest.cs
@@ -443,4 +443,61 @@ public class ConsumerConsumeTest
             break;
         }
     }
+
+    [Fact]
+    public async Task Consume_right_amount_of_messages()
+    {
+        await using var server = NatsServer.StartJS();
+        await using var nats = server.CreateClientConnection();
+
+        var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
+
+        var js = new NatsJSContext(nats);
+        await js.CreateStreamAsync("s1", ["s1.*"], cts.Token);
+
+        var payload = new byte[1024];
+        for (var i = 0; i < 50; i++)
+        {
+            var ack = await js.PublishAsync("s1.foo", payload, cancellationToken: cts.Token);
+            ack.EnsureSuccess();
+        }
+
+        // Max messages
+        {
+            var consumer = await js.CreateOrUpdateConsumerAsync("s1", "c1", cancellationToken: cts.Token);
+            var opts = new NatsJSConsumeOpts { MaxMsgs = 10, };
+            var count = 0;
+            await foreach (var msg in consumer.ConsumeAsync<byte[]>(opts: opts, cancellationToken: cts.Token))
+            {
+                await msg.AckAsync(cancellationToken: cts.Token);
+                if (++count == 4)
+                    break;
+            }
+
+            await Retry.Until("consumer stats updated", async () =>
+            {
+                var info = (await js.GetConsumerAsync("s1", "c1", cts.Token)).Info;
+                return info is { NumAckPending: 6, NumPending: 40 };
+            });
+        }
+
+        // Max bytes
+        {
+            var consumer = await js.CreateOrUpdateConsumerAsync("s1", "c2", cancellationToken: cts.Token);
+            var opts = new NatsJSConsumeOpts { MaxBytes = 10 * (1024 + 50), };
+            var count = 0;
+            await foreach (var msg in consumer.ConsumeAsync<byte[]>(opts: opts, cancellationToken: cts.Token))
+            {
+                await msg.AckAsync(cancellationToken: cts.Token);
+                if (++count == 4)
+                    break;
+            }
+
+            await Retry.Until("consumer stats updated", async () =>
+            {
+                var info = (await js.GetConsumerAsync("s1", "c2", cts.Token)).Info;
+                return info is { NumAckPending: 6, NumPending: 40 };
+            });
+        }
+    }
 }


src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
```diff
@@ -128,11 +128,12 @@ internal class NatsJSConsume<TMsg> : NatsSubBase
...
-        // to the user from the subscription channel (which should be set to a
-        // sufficiently large value to avoid blocking socket reads in the
-        // NATS connection).
-        _userMsgs = Channel.CreateBounded<NatsJSMsg<TMsg>>(1000);
+        // This channel is used to pass messages to the user from the subscription channel.
+        // Capacity is set to maxMsgs to avoid pulling too many messages from the server
+        // resulting in a large number of pending messages (which can be a problem if the
+        // application is slow to process messages). The capacity is bounded to a maximum of 1024
+        // to avoid LOH allocations.
+        _userMsgs = Channel.CreateBounded<NatsJSMsg<TMsg>>(maxMsgs > 1024 ? 1024 : (int)maxMsgs);
```
caleblloyd, line 136: Is there ever a case where max msgs is 0(or negative), such as if they only specify max bytes?

mtmk: Good point. should we set a minimum?

caleblloyd: Maybe we should just make sure it’s at least 1? If <=0 default it to 1024, because that probably means they are setting max bytes and not max msgs

mtmk: makes sense

caleblloyd:
What happens if they use MaxBytes instead?  Won't the channel be 1024 again and it will pull through all of the messages until it is full again?

Should we introduce an intermediate channel that consumes the full pull, then calls for the next pull after it is half way consumed?

mtmk:
(revisiting this because of #608)

So the idea is to properly handle the backpressure i.e. do not issue pull requests unless application code has received the message. Before this PR that check was done after the message was buffered through the channel and we were pulling 1000 messages (size of the bounded channel) regardless of application code actually picking up the message.

Issuing a `Delivered()` call after yielding the message to application, we can then make the calculation and decide if we should issue another Pull Request to the server or not.

mtmk:
I need to fix the tests...

src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
mtmk: changing the implementation to use 'Delivered()' function. this is set to a fixed 1024 since the size of the channel will not affect the pull requests anymore.

mtmk:
> What happens if they use MaxBytes instead? Won't the channel be 1024 again and it will pull through all of the messages until it is full again?
> 
> Should we introduce an intermediate channel that consumes the full pull, then calls for the next pull after it is half way consumed?

we don't have to worry about this any more since calculation is done after message is yielded through the `Delivered()` method.

mtmk:
moved to #626

PR #509, opened 2024-06-06
Build and test updates

mtmk:
* Added `-p:ContinuousIntegrationBuild=true` to packing for Deterministic builds: https://nuget.info/packages/NATS.Client.Core/2.2.2
image
* Build memory tests in release mode to get GC behave realistically
* Updated nunit for memory tests
* Don't pack by default to avoid accidental publishing

Note that we have an issue running dotMemory unit with net8.0


caleblloyd (approved):
Lgtm

PR #522, opened 2024-06-20
Release 2.3.0-preview.1

mtmk:
* Support for .NET Standard 2.0 and 2.1 (#513)
* Add TLS options for netstandard (#524)


caleblloyd:
I think we need to go through the TLS Options, and Throw on netstandard if they are using a TLS Option that is not supported instead of silently skipping those options

mtmk:
> I think we need to go through the TLS Options, and Throw on netstandard if they are using a TLS Option that is not supported instead of silently skipping those options

done in #524

Jarema (approved):
LGTM

PR #524, opened 2024-06-20
Add TLS options for netstandard

mtmk:


tests/NATS.Client.Platform.Windows.Tests/TlsTests.cs
```diff
@@ -0,0 +1,150 @@
...
+    private static bool ValidateCert(SslPolicyErrors errors, X509Certificate certificate, List<X509Certificate2> trustedCaCerts)
+    {
+        if (errors == SslPolicyErrors.None)
+            return true;
+
+        var cert2 = new X509Certificate2(certificate);
+
+        foreach (var caCert in trustedCaCerts)
+        {
+            var chain = new X509Chain();
```
to11mtm, line 70: I know they're tests but X509 is _weird_ so I get paranoid about files/etc leaking from tests...

`X509Chain`  implements `IDisposable`, should this be in a using statement for safety?

tests/NATS.Client.Platform.Windows.Tests/TlsTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        return caCertificates;
+    }
+
+    private static X509Certificate2Collection LoadClientCerts(string certPemPath, string keyPemPath)
+    {
+        var collection = new X509Certificate2Collection();
+
+        var certPem = File.ReadAllText(certPemPath);
+        var certParser = new X509CertificateParser();
+        var cert = certParser.ReadCertificate(new MemoryStream(Encoding.ASCII.GetBytes(certPem)));
```
to11mtm, line 108: ... On one hand, I would -think- that `ReadAllBytes` should work here as PEM files themselves -should- be ASCII (or perhaps 'UTF8 but in ASCII space so who notices'.

OTOH, it's a test so 🤷‍♂️ 

And TBH this approach looks pretty slick overall, I need to double-check but I _think_ it's way less 'hazardous' than loading the PEM directly (if you do that, at least on windows the risk of getting random zombie cert bytes in a local store go way up if the lifetime isn't properly managed... [interesting read](https://snede.net/the-most-dangerous-constructor-in-net/).)

Might need to borrow some of this elsewhere regardless,  😉 .

to11mtm (commented):
I will try to look into a few things but IMO the 'risks' are primarily on tests (i.e. cert file leakage on windows platforms, but this looks as safe as it can short of managing lifetimes of the certs themselves.)

That last thing said, I do feel there may be some benefit to mentioning some of the 'hazards' of `X509Certificate2` lifetimes... somewhere, as mismanagement can result in adverse conditions in some cases, giving users some sort of warning may alleviate or ease the explanation of issues filed around related issues on the repo. 

Or, I'm totally overthinking it. 😅 .

mtmk:
No that's fair. I guess you're referring to the callback for loading certs? I tried to keep everything else mapping directly to sslstream so any warnings using that would apply I guess.

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -235,6 +240,18 @@ internal async ValueTask<SslClientAuthenticationOptions> AuthenticateAsClientOpt
         return options;
 #pragma warning restore CS0618 // Type or member is obsolete
     }
+
 #endif
 
+    internal TlsMode EffectiveMode(NatsUri uri) => Mode switch
+    {
+        TlsMode.Auto => HasTlsCerts || uri.Uri.Scheme.ToLower() == "tls" ? TlsMode.Require : TlsMode.Prefer,
```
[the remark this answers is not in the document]
mtmk: fixed

src/NATS.Client.Core/NatsTlsOpts.cs
```diff
@@ -45,6 +45,55 @@ public sealed record NatsTlsOpts
...
+    public LocalCertificateSelectionCallback? LocalCertificateSelectionCallback { get; init; }
+
+    /// <summary>
+    /// Function called to load client certificates for TLS authentication.
+    /// </summary>
+    /// <remarks>
+    /// The function should return a <see cref="X509Certificate2Collection"/> containing the client certificates to be used for TLS authentication.
+    /// If this function is not set, the client certificates will not be loaded and <c>null</c> will be passed to <see cref="SslStream"/>.
+    /// Returned value of this function will be passed to <see cref="SslStream"/> and it will be called when connecting and reconnecting
+    /// so that the client certificates can be reloaded if necessary.
```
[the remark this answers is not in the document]
mtmk: good point. done.

tests/NATS.Client.Platform.Windows.Tests/TlsTests.cs
mtmk: done

tests/NATS.Client.Platform.Windows.Tests/TlsTests.cs
mtmk: done

to11mtm (approved):
I pulled the branch down and could not get errant files to show up where I know to look,

LGTM, gr8 job navigating the maze of X509!

PR #527, opened 2024-06-26
NETStandard TLS Parity

caleblloyd:
- ~~Imports BouncyCastle when target is NETStandard for PEM Reading~~
- Removes obsolete `NatsTlsOpts` methods
- Adds ability to load a Client Cert/Key Bundle from PKCS#12 .pfx file
- API Parity for `NatsTlsOpts` between NETStandard  and NET6/8, except Client Cert/Key PEM functions are excluded from NETStandard

### DIFF
--- a/src/NATS.Client.Core/Internal/SslClientAuthenticationOptionsExtensions.cs
+++ b/src/NATS.Client.Core/Internal/SslClientAuthenticationOptionsExtensions.cs
@@ -44,7 +44,10 @@ internal static class SslClientAuthenticationOptionsExtensions
         var leafCert = new X509Certificate2(certBundleFile);
         var intermediateCerts = new X509Certificate2Collection();
         intermediateCerts.Import(certBundleFile);
-        if (intermediateCerts.Count > 0)
+
+        // Linux does not include the leaf by default, but Windows does
+        // compare leaf to first intermediate just to be sure to catch all platform differences
+        if (intermediateCerts.Count > 0 && intermediateCerts[0].RawData.SequenceEqual(leafCert.RawData))
         {
             intermediateCerts.RemoveAt(0);
         }


mtmk:
I'm wondering Bouncy dependency is going to be an issue for some people. Be nice if we can bring in a simple implementation for reading PEM. it doesn't need to be super efficient at all.

caleblloyd:
We could potentially require a different dependency for Bouncy Castle instead of bundling into `Nats.Client.Core`?  Then throw `NotSuppportedException` with a message that you have to install another dependency to use the pem-related fields on NETStandard

caleblloyd:
After further discussion on Slack, @mtmk and I thought that it would be best to

- Remove Bouncy Castle dependency
- Add support for importing PEM-encoded CA Certs on NETStandard
- Remove support for importing PEM-encoded Client Cert/Key on NETStandard.  Too tricky to build an X509Certificate2 without any of the PEM methods to load/associate PEM-encoded Private Keys
- Add support for using PKCS#12 .pem files for client certificates on all targets

src/NATS.Client.Core/Internal/SslClientAuthenticationOptionsExtensions.cs
```diff
@@ -18,7 +18,41 @@ public static SslClientAuthenticationOptions LoadClientCertFromPem(this SslClien
...
+
+        // On Windows, ephemeral keys/certificates do not work with schannel. e.g. unless stored in certificate store.
+        // https://github.com/dotnet/runtime/issues/66283#issuecomment-1061014225
+        // https://github.com/dotnet/runtime/blob/380a4723ea98067c28d54f30e1a652483a6a257a/src/libraries/System.Net.Security/tests/FunctionalTests/TestHelper.cs#L192-L197
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            var ephemeral = leafCert;
+            leafCert = new X509Certificate2(leafCert.Export(X509ContentType.Pfx));
+            ephemeral.Dispose();
+        }
```
mtmk, line 36: why is this ephemeral trick moved up here?

src/NATS.Client.Core/Internal/SslClientAuthenticationOptionsExtensions.cs
```diff
@@ -38,16 +72,6 @@ public static SslClientAuthenticationOptions LoadCaCertsFromPem(this SslClientAu
...
     {
-        // On Windows, ephemeral keys/certificates do not work with schannel. e.g. unless stored in certificate store.
-        // https://github.com/dotnet/runtime/issues/66283#issuecomment-1061014225
-        // https://github.com/dotnet/runtime/blob/380a4723ea98067c28d54f30e1a652483a6a257a/src/libraries/System.Net.Security/tests/FunctionalTests/TestHelper.cs#L192-L197
-        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-        {
-            var ephemeral = leafCert;
-            leafCert = new X509Certificate2(leafCert.Export(X509ContentType.Pfx));
-            ephemeral.Dispose();
-        }
```
mtmk, line 49: same comment. do we think application code calling this extension should make their own choice about the ephemeral thing?

mtmk (approved):
LGTM

Really good work. api looks quite nice, all the obsoletes gone, and as consistent as it can be between netstnadard and net'normal', thank you @caleblloyd 

(minor question about ephemeral fix for windows, just curious)

src/NATS.Client.Core/Internal/SslClientAuthenticationOptionsExtensions.cs
caleblloyd: The ephemeral thing didn’t work with a PFX loaded with default options, it threw an exception that it couldn’t be exported

src/NATS.Client.Core/Internal/SslClientAuthenticationOptionsExtensions.cs
caleblloyd: The extensions are internal currently. I was thinking if there is interest we could make them public eventually

PR #530, opened 2024-06-30
[PROPOSAL] NATS Client POC

mtmk:
Expand public API of NATS client library

* Introduce NATS.Client NuGet package (to be published on top on v1 client)
* Simplified NatsClient with more beginner-friendly options
* Exposed more internals of NatsConnection as low level APIs
* Use INatsConnection interface as JetStream dependency


mtmk:
implemented in #607

PR #535, opened 2024-07-02
Full coverage kv management methods

### LINKED ISSUE #510: Ensure full coverage of KV management methods
* https://github.com/nats-io/nats-architecture-and-design/issues/284

Ivandemidov00:
* resolves #510

### DIFF
--- a/src/NATS.Client.KeyValueStore/NatsKVContext.cs
+++ b/src/NATS.Client.KeyValueStore/NatsKVContext.cs
@@ -55,7 +55,7 @@ public class NatsKVContext : INatsKVContext
     {
         ValidateBucketName(config.Bucket);
 
-        var streamConfig = CreateStreamConfig(config);
+        var streamConfig = NatsKVContext.CreateStreamConfig(config);
 
         var stream = await _context.CreateStreamAsync(streamConfig, cancellationToken);
 
@@ -99,7 +99,7 @@ public class NatsKVContext : INatsKVContext
     {
         ValidateBucketName(config.Bucket);
 
-        var streamConfig = CreateStreamConfig(config);
+        var streamConfig = NatsKVContext.CreateStreamConfig(config);
 
         var stream = await _context.UpdateStreamAsync(streamConfig, cancellationToken);
 
@@ -166,12 +166,12 @@ public class NatsKVContext : INatsKVContext
     }
 
     private static string BucketToStream(string bucket)
-        => KvStreamNamePrefix + bucket;
+        => $"{KvStreamNamePrefix}{bucket}";
 
     private static string ExtractBucketName(string streamName)
         => streamName.Substring(KvStreamNamePrefixLen);
 
-    private StreamConfig CreateStreamConfig(NatsKVConfig config)
+    private static StreamConfig CreateStreamConfig(NatsKVConfig config)
     {
         // TODO: KV Mirrors
         var subjects = new[] { $"$KV.{config.Bucket}.>" };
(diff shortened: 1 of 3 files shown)


Ivandemidov00:
@mtmk I couldn't find the CreateOrUpdateStream api. Do I correctly understand that it is impossible to implement the CreateOrUpdateStore method without calling the server twice? Is CreateOrUpdateStore necessary under such conditions?

mtmk:
> @mtmk I couldn't find the CreateOrUpdateStream api. Do I correctly understand that it is impossible to implement the CreateOrUpdateStore method without calling the server twice? Is CreateOrUpdateStore necessary under such conditions?

Well spotted. let's leave that out. afaik there is no create-or-update stream call. cc @Jarema

tests/NATS.Client.KeyValueStore.Tests/KeyValueContextTest.cs
```diff
@@ -0,0 +1,122 @@
...
+        var cts = new CancellationTokenSource(TimeSpan.FromSeconds(10));
+        var cancellationToken = cts.Token;
+
+        await using var server = NatsServer.StartJS();
+        await using var nats = server.CreateClientConnection();
+
+        var js = new NatsJSContext(nats);
+        var kv = new NatsKVContext(js);
+
+        var natsKVConfig = new NatsKVConfig(bucketName) { Compression = true };
```
mtmk, line 113: Compression isn't supported in server 2.9.x which we're still testing against so it's failing the build. maybe use some other property e.g. max_msgs here.

Ivandemidov00: Fixed. Strange, now fallen another `Reconnect_doesnt_drop_partially_sent_msgs `

mtmk: don't worry. that's a flapper. I'll re-run it.

src/NATS.Client.KeyValueStore/NatsKVContext.cs
```diff
@@ -19,6 +19,8 @@ public enum NatsKVStorageType
 /// </summary>
 public class NatsKVContext : INatsKVContext
 {
+    private const string KvStreamPrefix = "KV_";
```
rickdotnet, line 22: Generally, wouldn't be a stickler on this but using `KvStreamNamePrefix` would separate it from the subject prefix `$KV` used below.

src/NATS.Client.KeyValueStore/NatsKVContext.cs
```diff
@@ -45,10 +49,130 @@ public ValueTask<INatsKVStore> CreateStoreAsync(string bucket, CancellationToken
...
+    internal static void ValidateBucketName(string bucket)
+    {
+        if (!ValidBucketRegex.IsMatch(bucket))
+        {
+            throw new NatsKVException("Invalid bucket name");
+        }
+    }
+
+    private static string BucketToStream(string bucket)
+        => $"KV_{bucket}";
```
rickdotnet, line 169: I'd suggest using `KvStreamNamePrefix` here.

rickdotnet (changes requested):
Aside from two minor suggestions, this looks great. I can make those updates if you'd like.

mtmk:
@Ivandemidov00 can you sign your commits next time please? (if you're comfortable with git rebase you can push -f that would be good but don't worry if not) sorry keep forgetting to remind you and you've contributed quite a bit already, we're very grateful for that 💯

mtmk (approved):

PR #541, opened 2024-07-04
Fixed version number sent to server and changed lang to .NET from C#.…

darkwatchuk:
Version number of the assembly wasn't previously picked up and was always sent to the server as 1.0.0

This PR sends through the version, and also changes the lang to .NET from C#.  This aligns with the V1 version.

### DIFF
--- a/src/NATS.Client.Core/Internal/ClientOpts.cs
+++ b/src/NATS.Client.Core/Internal/ClientOpts.cs
@@ -68,7 +68,7 @@ internal sealed class ClientOpts
 
     /// <summary>The implementation language of the client.</summary>
     [JsonPropertyName("lang")]
-    public string ClientLang { get; init; } = "C#";
+    public string ClientLang { get; init; } = ".NET";
 
     /// <summary>The version of the client.</summary>
     [JsonPropertyName("version")]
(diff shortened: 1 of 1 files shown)


src/NATS.Client.Core/Internal/ClientOpts.cs
```diff
@@ -97,23 +97,12 @@ public static ClientOpts Create(NatsOpts opts)
...
-            }
-        }
-
-        return version;
+        var asm = typeof(ClientOpts).Assembly;
+
+        return asm?.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version
+            ?? asm?.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion
+            ?? asm?.GetCustomAttribute<AssemblyVersionAttribute>()?.Version
+            ?? "?.?.?";
```
mtmk, line 105: let's keep this as 1.0.0 in case there is something out there parsing it as numbers

darkwatchuk: Have done that, but thinking about it, maybe 2.0.0 might be better as V2 of the library?

mtmk: ah yes! good point

src/NATS.Client.Core/Internal/ClientOpts.cs
```diff
@@ -97,23 +97,12 @@ public static ClientOpts Create(NatsOpts opts)
...
-            {
-                version = asmVersion.Version;
-            }
-        }
-
-        return version;
+        var asm = typeof(ClientOpts).Assembly;
+
+        return asm?.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version
+            ?? asm?.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion
```
mtmk, line 103: Question about which attribute to use:

Just playing with it changing version.txt here are the differences between file and informational versions:

| version.txt     | file    | informational                                                   |
|-----------------|---------|-----------------------------------------------------------------|
| 2.3.0           | 2.3.0.0 | 2.3.0\n\u002B040efe2cffc07cbbcb4267afdc779275aa12d4a7           |
| 2.3.0-preview.1 | 2.3.0.0 | 2.3.0-preview.1\n\u002B040efe2cffc07cbbcb4267afdc779275aa12d4a7 |

do you think we should chop informational so that we can send preview suffixes as well?

mtmk:
Also we should probably fix the name:

https://github.com/nats-io/nats.net.v2/blob/d80fa385c3dc0656d17368b1c94bcbf415634436/src/NATS.Client.Core/NatsOpts.cs#L18

`.Net` should be all caps: `"NATS .NET Client"`

src/NATS.Client.Core/Internal/ClientOpts.cs
darkwatchuk: In that case, let's just use the first part of the informational attribute, and forget trying the other ones.  I don't see the point in trying for the other attributes....

So just a simplied version;

```
   var asm = typeof(ClientOpts).Assembly;
   var version = "2.0.0";

   var attrInformationalVersion = asm!.GetCustomAttribute<AssemblyInformationalVersionAttribute>();
   if (attrInformationalVersion != null)
   {
       version = attrInformationalVersion.InformationalVersion.Split('\n')[0];
   }

   return version;
```

PR #542, opened 2024-07-04
netstandard Nullable ref fix

mtmk:
Made a mistake with Nullable reference. It's only required compile time.

image

image

see also https://github.com/nats-io/nats.net.v2/issues/306#issuecomment-2209404686

cc @galvesribeiro


galvesribeiro:
LGTM

stebet (approved):
LGTM

PR #543, opened 2024-07-04
Unity il2cpp build fix

mtmk:
Fixing unity il2cpp / AOT issue https://github.com/nats-io/nats.net.v2/issues/306#issuecomment-2209483081

cc @jasper-d (sorry for the tag, it'd be great if you can comment or the inner function move. I assumed it was used to save `sequential` local allocation)

cc @galvesribeiro

### DIFF
--- a/src/NATS.Client.Core/Internal/NuidWriter.cs
+++ b/src/NATS.Client.Core/Internal/NuidWriter.cs
@@ -144,13 +144,13 @@ internal sealed class NuidWriter
         }
 
         return RefreshAndWrite(nuidBuffer);
+    }
 
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        bool RefreshAndWrite(Span<char> buffer)
-        {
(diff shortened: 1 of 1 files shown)


to11mtm (approved):
LGTM!

jasper-d:
@mtmk LGTM

Sorry, I am currently a bit slow to respond.

I believe the only reason to use a local function was that it has only one call site.

NB: This function exists only to be able to outline it, since it will be called infrequently.

mtmk:
> @mtmk LGTM
> 
> Sorry, I am currently a bit slow to respond.
> 
> I believe the only reason to use a local function was that it has only one call site.
> 
> NB: This function exists only to be able to outline it, since it will be called infrequently.

Thank you for the response @jasper-d good to hear from you. unfortunately there's been a little confusion on my part. this PR actually did not fix the ip2cpp issue. The fix is in a new PR #548 if you have a few more minutes please feel free to have a look at that one.

PR #549, opened 2024-07-06
Base64 Encoding simplification + optimization

### LINKED ISSUE #518: StoredMessage.Data as System.Buffers.Text
System.Buffers.Text [has To/From for utf8 byte sequences](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.text.base64.decodefromutf8?view=net-8.0) and it looks like they work with netstandard20, adapting to such an approach longer term may help consolidate this code as well as lower the deserialization cost of KV and other stream reads in NET8 (as we can hold the base64 in UTF8 bytes rather than UTF16 string, maybe pool in future too? Not sure how easy it would be to do that part though...)

see also https://github.com/nats-io/nats.net.v2/pull/513#discussion_r1642041779

to11mtm:
Resolves #518 

tl;dr - we are optimizing a lot of our base64 encode parsing by using `System.Text.Buffers` as it has NetStandard2.0 compatibility, which will also allow us to have less code to maintain long term. :)

This is a semi-breaking change however, as it does the following:

1. `System.Text.Json` is updated to the newest version, to support `ReadOnlyMemory<byte>`
   1. this version is compatible with NetStandard2.0 and Net6, so, win-win?
   2. Other win-win, by doing this, we instead rely on the implementations of how `System.Text.Json` handles `ReadOnlyMemory<byte>`, i.e. it does the Base64 decoding internally and should be 'good' for all versions (if there is issue we can always PR them!)
2. We use `ReadOnlyMemory<Byte>` for our model here now. If someone is somehow relying on that type for some test or other behavior it will be a breaking change
  1. FWIW this should be a -compile- breaking change, but if someone has a dis-proof of that I'm all ears.

### DIFF
--- a/src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
+++ b/src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
@@ -36,7 +36,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Text.Json" Version="8.0.3" />
+    <PackageReference Include="System.Text.Json" Version="8.0.4" />
   </ItemGroup>
 
 </Project>
--- a/src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
+++ b/src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
@@ -35,8 +35,4 @@
     <InternalsVisibleTo Include="NATS.Client.Core.MemoryTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100db7da1f2f89089327b47d26d69666fad20861f24e9acdb13965fb6c64dfee8da589b495df37a75e934ddbacb0752a42c40f3dbc79614eec9bb2a0b6741f9e2ad2876f95e74d54c23eef0063eb4efb1e7d824ee8a695b647c113c92834f04a3a83fb60f435814ddf5c4e5f66a168139c4c1b1a50a3e60c164d180e265b1f000cd"/>
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="System.Text.Json" Version="8.0.4" />
-  </ItemGroup>
-
 </Project>
--- a/src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
+++ b/src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
@@ -35,4 +35,8 @@
     <InternalsVisibleTo Include="NATS.Client.Core.MemoryTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100db7da1f2f89089327b47d26d69666fad20861f24e9acdb13965fb6c64dfee8da589b495df37a75e934ddbacb0752a42c40f3dbc79614eec9bb2a0b6741f9e2ad2876f95e74d54c23eef0063eb4efb1e7d824ee8a695b647c113c92834f04a3a83fb60f435814ddf5c4e5f66a168139c4c1b1a50a3e60c164d180e265b1f000cd"/>
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Text.Json" Version="8.0.3" />
+  </ItemGroup>
+
 </Project>
--- a/src/NATS.Client.ObjectStore/NatsObjStore.cs
+++ b/src/NATS.Client.ObjectStore/NatsObjStore.cs
@@ -1,6 +1,7 @@
 using System.Buffers;
 using System.Runtime.CompilerServices;
 using System.Security.Cryptography;
+using System.Text;
 using System.Text.Json;
 using NATS.Client.Core;
 using NATS.Client.Core.Internal;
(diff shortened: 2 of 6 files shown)


src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
```diff
@@ -35,4 +35,8 @@
     <InternalsVisibleTo Include="NATS.Client.Core.MemoryTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100db7da1f2f89089327b47d26d69666fad20861f24e9acdb13965fb6c64dfee8da589b495df37a75e934ddbacb0752a42c40f3dbc79614eec9bb2a0b6741f9e2ad2876f95e74d54c23eef0063eb4efb1e7d824ee8a695b647c113c92834f04a3a83fb60f435814ddf5c4e5f66a168139c4c1b1a50a3e60c164d180e265b1f000cd"/>
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Text.Json" Version="8.0.3" />
+  </ItemGroup>
+
```
to11mtm, line 41: Admittedly, IDK if I should be putting this elsewhere, but this was the 'lowest need' for the PR.

As noted, this works gr8 with netstandard2.0 and net6, so the biggest 'hazard' is consumers may have to deal with other changes.

src/NATS.Client.ObjectStore/NatsObjStore.cs
```diff
@@ -530,17 +531,33 @@ public async ValueTask<ObjectMetadata> GetInfoAsync(string key, bool showDeleted
...
+                }
+                else
+                {
+                    throw new NatsObjException("Can't decode data message value");
+                }
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(bytes);
+            }
```
to11mtm, line 560: Ironically, this optimizes NET8 too 😁. Not sure how much of a hot path this can be but why not?

That said, If there is one regret, IDK how to make this easily 'composable where this doesn't feel like a copypaste of the parse with buffer'. OTOH I had had a colleague spend some time teaching me go so the overton window for pragmatism has changed. 🤷

to11mtm (commented):
Added some thoughts/reasoning to the changes given.

to11mtm:
Thinking these errors -may- be related to some whitespace/etc present in the payload.

`Convert.FromBase64String` tends to be more tolerant of that... FWIW we should still be able to 'safely' trim things if that's the case (and I don't forsee a perf penalty since `FromBaase64String` is doing it too...)

Just gotta get my local test flow working again. 😅

mtmk:
> `Convert.FromBase64String` tends to be more tolerant of that... FWIW we should still be able to 'safely' trim things if that's the case (and I don't forsee a perf penalty since `FromBaase64String` is doing it too...)

it's not only us we should be compatible with other clients as well, but I'd say that probably won't be a problem once we verified we're able to read and write proper base64. we should double check once local tests are passing (I can do that as well)

src/NATS.Client.ObjectStore/NatsObjStore.cs
```diff
@@ -530,17 +531,18 @@ public async ValueTask<ObjectMetadata> GetInfoAsync(string key, bool showDeleted
...
+            if (response.Message.Data.Length == 0)
             {
                 throw new NatsObjException("Can't decode data message value");
             }
 
-            var base64String = Convert.FromBase64String(response.Message.Data);
-            var data = NatsObjJsonSerializer<ObjectMetadata>.Default.Deserialize(new ReadOnlySequence<byte>(base64String)) ?? throw new NatsObjException("Can't deserialize object metadata");
+            ObjectMetadata data;
+                    var buffer = new ReadOnlySequence<byte>(response.Message.Data);
+                    data = NatsObjJsonSerializer<ObjectMetadata>.Default.Deserialize(buffer) ?? throw new NatsObjException("Can't deserialize object metadata");
```
to11mtm, line 545: oops, should still clean this up.

to11mtm:
> > `Convert.FromBase64String` tends to be more tolerant of that... FWIW we should still be able to 'safely' trim things if that's the case (and I don't forsee a perf penalty since `FromBaase64String` is doing it too...)
> 
> it's not only us we should be compatible with other clients as well, but I'd say that probably won't be a problem once we verified we're able to read and write proper base64. we should double check once local tests are passing (I can do that as well)

🤯 

I am amazed.

Tests were -actually- failing because `ReadOnlyMemory<byte>` in JSON Schema, will handle Base64 all on it's own! 🤯 

Once I was able to get tests (kinda) running locally it was pretty obvious to see what was happening. (will discuss to see if we should file something for the sake of doc or easing for others in future).

On one hand, we are semi 'at mercy' of however `System.Text.Json` is doing the Base64 decode, OTOH it's a lot less code for us to maintain and I'm guessing that we are still winning overall on alloc/perf since we are avoiding a UTF8->String conversion here.

mtmk:
> Tests were -actually- failing because `ReadOnlyMemory<byte>` in JSON Schema, will handle Base64 all on it's own! 🤯

woow! very cool

galvesribeiro:
> Tests were -actually- failing because ReadOnlyMemory<byte> in JSON Schema, will handle Base64 all on it's own! 

Yeah, there is a converter for it. Similar to `byte[]`.

> On one hand, we are semi 'at mercy' of however System.Text.Json is doing the Base64 decode, OTOH it's a lot less code for us to maintain and I'm guessing that we are still winning overall on alloc/perf since we are avoiding a UTF8->String conversion here.

Yeah, I would leave that to STJ. There is a huge work over the past couple of major releases to optimize the UTF8 handling in it. I don't see why we should need to roll our own in this case. 

What we should do is to make sure all types that come from Nats are using STJ Source Generators. That would avoid the reflection costs at runtime. Besides that, I won't see why would we go lower than that. I mean, yeah, we can always optimize something but it gets to a point of diminish return.

to11mtm:
> Yeah, there is a converter for it. Similar to `byte[]`.

> Yeah, I would leave that to STJ. There is a huge work over the past couple of major releases to optimize the UTF8 handling in it. I don't see why we should need to roll our own in this case.

Agreed.

> What we should do is to make sure all types that come from Nats are using STJ Source Generators. That would avoid the reflection costs at runtime. Besides that, I won't see why would we go lower than that. I mean, yeah, we can always optimize something but it gets to a point of diminish return.

Agreed, if we do see an opportunity we can always contrib to upstream if needed. :)

I believe there's already sourcegen in play for STJ (I remember seeing some errors when I had bad composition), however your comment did lead me to look and I think we can also obviate a decent amount of code in ObjectStore with this change...

Although, not sure if we should keep that in same PR or separate, just for sake of sizing... open to opinions here.

galvesribeiro:
> Although, not sure if we should keep that in same PR or separate, just for sake of sizing... open to opinions here.

Yeah, lets do on a separated round.

to11mtm:
> > Although, not sure if we should keep that in same PR or separate, just for sake of sizing... open to opinions here.
> 
> Yeah, lets do on a separated round.

Created an issue to handle that (#557)

... TBH I think I'm good to pull this out of draft, my only concern is the weird flicker on unit tests. This latest push failed on test/dotnet (2.9) pull_request, the previous push failed on test/donet(latest).

I am no stranger to flickery CI but will defer to the opinions of core team here.

galvesribeiro:
>  my only concern is the weird flicker on unit tests. 

When I was discussing the netstandard2.0 support, the tests flakiness were mentioned. I think it is essentially the fixtures as your test results are looking good. 

I suggested using TestContainers on it to get the fixtures working properly. I'll take a look on how to improve it and will create an issue.

mtmk (approved):
LGTM thanks @to11mtm

PR #554, opened 2024-07-09
Remove timeout from readerTask in CommandWriter

### LINKED ISSUE #553: Retry loop stopped and connection state is invalid
Reconnect loop is terminating because of a timeout exception in `CommanWriter.CancelReaderLoopAsync()`

https://github.com/nats-io/nats.net.v2/blob/7d72a38bf1d69507a0c8f411019f90623dbccce3/src/NATS.Client.Core/Commands/CommandWriter.cs#L124

The only place I'm guessing stopping readerLoop to not exit is the socket send:

https://github.com/nats-io/nats.net.v2/blob/7d72a38bf1d69507a0c8f411019f90623dbccce3/src/NATS.Client.Core/Commands/CommandWriter.cs#L476

...which doesn't take a cancellation token.

### Possible solutions

* Remove timeout on awaiting readerLoop since we don't actually have a good way to recover from a timeout since the Pipe isn't recreated and pipe is tied to that task. So we can't do anything but wait for the readerLoop to complete (possibly with a socket exception
* ~We need to handle that exception and ignore if it's a socket exception~ Actually any exception is just logged and ignored within the reader loop method.

* Another solution is to extend the timeout or make it configurable?
* suggestions?

### Discussed in https://github.com/nats-io/nats.net.v2/discussions/552

<sup>Originally posted by **robertmircea** July  9, 2024</sup>
I have a .net core 8 server which runs as a service permanently. I've noticed that starting at some point in time, the following message appeared in the log from NatsConnection and from that moment on, the server could not publish Nats messages anymore and never recovered without restart which is very bad for a server which supposes to run continuously. Even if the nats server was not accessible for a while (which I don't think is the case), wouldn't NatsConnection suppose to restore the connection to Nats servers when appropriate?

Nats nuget versions:
```
        <PackageReference Include="NATS.Extensions.Microsoft.DependencyInjection" Version="2.3.0" />
        <PackageReference Include="NATS.Net" Version="2.3.0" />
```
Nats servers are all three running in cluster mode hosted on-prem, version:  2.10.11

Stack trace of first exception observed in the app logs (nothing is missing) and from which issue started to occur:
```
[23:18:49.861 ERR] Retry loop stopped and connection state is invalid  } <s:NATS.Client.Core.NatsConnection>
System.TimeoutException: The operation has timed out.
   at NATS.Client.Core.Commands.CommandWriter.CancelReaderLoopAsync()
   at NATS.Client.Core.NatsConnection.ReconnectLoop()
```

The next lines of logs when attempting jetstream.PublishAsync were like these:

```
[23:18:58.461 ERR] NatsAuditEventProducer could not publish NATS Jetstream message  <s:AuditEvents.NatsAuditEventProducer>
NATS.Client.JetStream.NatsJSPublishNoResponseException: No response received from the server
   at NATS.Client.JetStream.NatsJSContext.PublishAsync[T](String subject, T data, INatsSerialize`1 serializer, NatsJSPubOpts opts, NatsHeaders headers, CancellationToken cancellationToken)
   at MyApp.AuditEvents.NatsAuditEventProducer.PublishAsync(AbstractAuditEvent auditEvent, CancellationToken cancellationToken)

[23:20:08.848 ERR] NatsAuditEventProducer could not publish NATS Jetstream message <s:MyApp.AuditEvents.NatsAuditEventProducer>
System.OperationCanceledException: The operation was canceled.
   at NATS.Client.Core.Commands.CommandWriter.PublishStateMachineAsync(Boolean lockHeld, String subject, String replyTo, NatsPooledBufferWriter`1 headersBuffer, NatsPooledBufferWriter`1 payloadBuffer, CancellationToken cancellationToken)
   at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
   at NATS.Client.Core.NatsConnection.RequestSubAsync[TRequest,TReply](String subject, TRequest data, NatsHeaders headers, INatsSerialize`1 requestSerializer, INatsDeserialize`1 replySerializer, NatsPubOpts requestOpts, NatsSubOpts replyOpts, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.PublishAsync[T](String subject, T data, INatsSerialize`1 serializer, NatsJSPubOpts opts, NatsHeaders headers, CancellationToken cancellationToken)
   at MyApp.AuditEvents.NatsAuditEventProducer.PublishAsync(AbstractAuditEvent auditEvent, CancellationToken cancellationToken)
```

Unfortunately I don't know how to reproduce this, but I was wondering if any recent nuget library code changes might have triggered this.</div>

mtmk:
The previous version of the code had a potential issue where a fixed timeout was set for the readerTask in the CommandWriter class. This timeout has been removed since it's not practical to recover from a timeout situation here. The reader loop should finish quickly as it's governed by the cancellation token and won't be blocked by the socket connection.

* resolves #553

### DIFF
--- a/src/NATS.Client.Core/Commands/CommandWriter.cs
+++ b/src/NATS.Client.Core/Commands/CommandWriter.cs
@@ -121,7 +121,14 @@ internal sealed class CommandWriter : IAsyncDisposable
         }
 
         if (readerTask != null)
-            await readerTask.WaitAsync(TimeSpan.FromSeconds(3), _cts.Token).ConfigureAwait(false);
+        {
+            // We have to wait for the reader loop to finish before we can reuse the pipe.
+            // Setting a timeout here is not practical, as we won't be able to recover from a timeout.
+            // The reader loop should finish quickly, as it will be canceled by the reader cancellation token.
+            // If it doesn't, the only potential blocker would be the socket connection, which should be
+            // closed by the time we get here or soon after.
+            await readerTask.WaitAsync(_cts.Token).ConfigureAwait(false);
+        }
     }
 
     public async ValueTask DisposeAsync()


caleblloyd (approved):
LGTM

PR #567, opened 2024-07-17
Resolved issue of ObjStore GetInfo MTime returning 0

### LINKED ISSUE #566: Modified Timestamp is missing from ObjectStore.GetInfoAsync
### Observed behavior

The MTime field on calling GetInfoAsync is coming back as 0, I think it might actually be coming back from the server as 0?

### Expected behavior

MTime should match the MTime for the same object that comes back as part of the ListAsync call

### Server and client version

Server 2.10.17
.NET client 2.3.2

### Host environment

tested locally on windows, unix, and nats.demo.io

### Steps to reproduce

This is a test against the selfies bucket on demo.nats.io
```

  [Fact]
  public async Task TestModifiedTimeStamps()
  {
      var cts = new CancellationTokenSource(TimeSpan.FromSeconds(10));
      var cancellationToken = cts.Token;

      await using var nats = new NatsConnection(new NatsOpts {
           Url = "nats://demo.nats.io",
      });

      var ob = new NatsObjContext(js);

      var objStore = await ob.GetObjectStoreAsync("selfies", cancellationToken);


      const string FILE_NAME = "selfie.png";
      DateTimeOffset? dtFromList = null;
      await foreach (var entry in objStore.ListAsync(cancellationToken: cancellationToken))
      {
          if (entry.Name == FILE_NAME)
          {
              dtFromList = entry.MTime;
              break;
          }
      }

      var f = await objStore.GetInfoAsync(FILE_NAME);
      DateTimeOffset? dtFromInfo = f.MTime;

      Assert.False(dtFromInfo == DateTimeOffset.MinValue, "GetInfo Timestamp is Zero");
      Assert.True(dtFromInfo == dtFromList, "Modified timestamps should match");
  }
```

darkwatchuk:
Resolved #566 issue of ObjStore GetInfo MTime returning 0


mtmk:
remind me where do we see this issue? is it related to the latest issue? also is it possible to have a test case?

mtmk:
build failures: looks like nunit tool update has an issue. I'll wait a bit to see if they fix it otherwise we might have to pin it to a specific version.

src/NATS.Client.ObjectStore/NatsObjStore.cs
```diff
@@ -542,6 +542,10 @@ public async ValueTask<ObjectMetadata> GetInfoAsync(string key, bool showDeleted
             var base64String = Convert.FromBase64String(response.Message.Data);
             var data = NatsObjJsonSerializer<ObjectMetadata>.Default.Deserialize(new ReadOnlySequence<byte>(base64String)) ?? throw new NatsObjException("Can't deserialize object metadata");
 
+            if (!DateTimeOffset.TryParse(response.Message.Time, out var created))
+                throw new NatsObjException("Can't parse timestamp message value");
+            data.MTime = created;
+
```
mtmk, line 548: This is good. matches the Go implementation: [implementation](https://github.com/nats-io/nats.go/blob/52a2d33b791b6e819aacdaeef2c0b79c2cfbf81b/object.go#L975), [type definition](https://github.com/nats-io/nats.go/blob/833574b050e80aa552e3f78e9767251159bb3452/jsm.go#L1205)

...except rather than parsing here we should change the `Time`'s type to `DateTimeOffset` instead here:

https://github.com/nats-io/nats.net/blob/b09b9ae09aa0751677690091fd066029e28b98e0/src/NATS.Client.JetStream/Models/StoredMessage.cs#L34-L46

mtmk (changes requested):
This fixes a bug and client compatibility issue thanks @darkwatchuk 

Only change i suggest is the type for time

PR #569, opened 2024-07-17
Update repository renaming information

mtmk:
The renaming info of the repository has been updated from the future tense to the past tense in both README.md and REPO_RENAME.md files. An additional link has been added in README.md pointing to the old NATS .NET v1 repository.


scottf (approved):
LGTM

PR #572, opened 2024-07-20
ObjStore empty list fix

### LINKED ISSUE #563: NatsObjStore.ListAsync hangs if bucket is empty
### Observed behavior

As per title, if an object store bucket has no entries, then calling ListAsync hangs and doesn't return.

### Expected behavior

Enumerator should return

### Server and client version

Client : .Net 2.3.3
Server : Various

### Host environment

_No response_

### Steps to reproduce

_No response_

Ivandemidov00:
Resolves #563

### DIFF
--- a/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
@@ -141,6 +141,9 @@ internal class NatsJSOrderedPushConsumer<T>
     {
         _nats.ConnectionDisconnected -= OnDisconnected;
 
+        // For correctly Dispose,
+        // first stop the consumer Creation operations and then the command execution operations.
+        // It is necessary that all consumerCreation operations have time to complete before command CommandLoop stop
         _consumerCreateChannel.Writer.TryComplete();
         await _consumerCreateTask;
 
--- a/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
@@ -142,11 +142,12 @@ internal class NatsJSOrderedPushConsumer<T>
         _nats.ConnectionDisconnected -= OnDisconnected;
 
         _consumerCreateChannel.Writer.TryComplete();
-        _commandChannel.Writer.TryComplete();
-        _msgChannel.Writer.TryComplete();
-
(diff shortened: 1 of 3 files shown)


src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
```diff
@@ -142,12 +142,13 @@ public async ValueTask DisposeAsync()
...
-        _commandChannel.Writer.TryComplete();
-        _msgChannel.Writer.TryComplete();
-
         await _consumerCreateTask;
+
+        _commandChannel.Writer.TryComplete();
         await _commandTask;
 
+        _msgChannel.Writer.TryComplete();
+
```
mtmk, line 151: why did we change these? did you encounter an issue? (you're probably right to make these changes, I just can't remember my reasoning here so just curious)

mtmk (commented):
Looking good 💯 just a quick question below.

src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
Ivandemidov00: Yes, I encountered some problem.  
Before exiting WatchAsync, if Enumerable is empty, call `pushConsumer.Init();`
The consumer is initialized and the same consumer is immediately deleted in the dispose method.  
The creation operation `_consumerCreateTask` has not yet had time to complete and it requires an active `_commandChannel`, but it is already closed. Based on this, I decided to change the order of stopping channel reading.

Ivandemidov00: Otherwise, when Dispose we get `NatsJSApiException: consumer not found`.

mtmk: awesome, thanks fixing that 💯 should we stick this as comment on top of the relevant lines do you think?

mtmk (approved):
LGTM thanks @Ivandemidov00

ps @Ivandemidov00 are you on slack.nats.io dotnet channel?

mtmk:
ouch sorry @Ivandemidov00 I keep forgetting this finally we put a check in github. could you please sign your commits? cheers
image

Ivandemidov00:
I signed the commits and moved them to the next branch https://github.com/nats-io/nats.net/pull/578.
I think we can close this request

PR #577, opened 2024-07-24
KV never set duplicate window

scottf:


mtmk (approved):
LGTM

src/NATS.Client.KeyValueStore/NatsKVContext.cs
```diff
@@ -226,13 +226,8 @@ private static StreamConfig CreateStreamConfig(NatsKVConfig config)
             // MirrorDirect =
             // Mirror =
             Retention = StreamConfigRetention.Limits, // from ADR-8
-            DuplicateWindow = config.MaxAge != default ? config.MaxAge : TimeSpan.FromMinutes(2), // 120_000_000_000ns, from ADR-8
```
mtmk, line 229: just to note bug was that when MaxAge was set to something like 30 minutes, DumplicateWindow was always set to the same value and that was creating a wrong config on the server. by letting the server decide we're avoiding any checks here. if we wanted to duplicate what Go client is doing, we could set DuplicateWindow to 2min here (avoiding the MaxAge check) and leave the check below. but I think this is a better solution. less complexity.

* see also #362

PR #578, opened 2024-07-24
Obj store empty list fixed

Ivandemidov00:
Resolves https://github.com/nats-io/nats.net/issues/563
initially from https://github.com/nats-io/nats.net/pull/572

PR #580, opened 2024-07-24
Release 2.3.3-preview.2

mtmk:
* Obj store empty list fixed (#578)
* KV never set duplicate window (#577)
* Resolved issue of ObjStore GetInfo MTime returning 0 (#567)


scottf (approved):
LGTM

PR #584, opened 2024-07-26
Handle various protocol errors and socket exceptions

mtmk:
Added code to handle socket exceptions during command writing, signaling disconnection to trigger reconnections as needed.

Breaking: log level changed for send loop socket exceptions


scottf (approved):
LGTM

PR #589, opened 2024-08-01
Add NATS client implementation

mtmk:
Added the initial implementation of the NATS client, including publish, subscribe, and request/reply functionalities. NATS.Client package is aimed at basic applications and samples or POCs, where especially serialization setup and other params are set to a sensible defaults.


galvesribeiro (approved):
LGTM

PR #582, opened 2024-07-25
Fetch idle timeout default fix

mtmk:
Allow heartbeat timeout to be set as TimeSpan.Zero so it's not sent to server e.g.
```
PUB $JS.API.CONSUMER.MSG.NEXT.s1.c1 _INBOX.hHyTxVZaGlxkX2lhPMFlho 35
{"expires":30000000000,"batch":100}
```

PR #570, opened 2024-07-18
Add logging enhancements and improve error handling

mtmk:
The commit refactors the error handling and logging within the NatsSvcServer and NatsSvcEndPoint classes. Updates have been made to the SvcListener, enhancing message logging and ensuring clean unsubscription on object disposal. The changes also facilitate better debugging and analysis by adding more logging statements for subscription handling within the SubscriptionManager class. Additional updates are made to handle possible errors in the handler loop within the NatsSvcEndPoint class.

PR #591, opened 2024-08-02
Release 2.3.3

mtmk:
* Handle various protocol errors and socket exceptions (#584)
* Fetch idle timeout default fix (#582)
* Obj store empty list fixed (#578)
* KV never set duplicate window (#577)
* Resolved issue of ObjStore GetInfo MTime returning 0 (#567)
* Add logging enhancements and improve error handling (#570)
* Base64 Encoding simplification + optimization (#549)
* Use string.Create when building a new inbox string (#551)


aricart (approved):
LGTM

PR #594, opened 2024-08-05
Add default timeout to initial commands

mtmk:
By removing infinite timeout sets the command writing to use default timeouts. There is a chance, while reconnecting, socket errors might be causing the re-subscriptions (performed by the PriorityCommandWriter) to stall. By having a timeout, we should recover from socket errors happening during re-subscriptions.


levb (approved):
LGTM

PR #607, opened 2024-08-20
Simplified NATS client

mtmk:
We moved functionality from the `NATS.Client` package into `NATS.Client.Simplified` because the `NATS.Client` package belongs to v1 and cannot be used. While we could have used the `NATS.Net` package, doing so would create dependencies on all other packages. To avoid this, NATS.Net remains a meta package that pulls everything together, allowing developers the flexibility to pull in packages gradually if they choose to use the simplified package with minimal dependencies.

There is one breaking change: the `NATS.Net` package will now reference the ad-hoc JSON serialization package, which may affect AOT (Ahead-of-Time) publishing for certain use cases, though this is likely uncommon. It's crucial that we avoid using non-AOT-friendly APIs, such as ad-hoc JSON serialization, in the core library.

### DIFF
--- a/src/NATS.Client.Simplified/NatsClient.cs
+++ b/src/NATS.Client.Simplified/NatsClient.cs
@@ -1,4 +1,4 @@
-﻿using System.Threading.Channels;
+using System.Threading.Channels;
 using NATS.Client.Core;
 
 namespace NATS.Net;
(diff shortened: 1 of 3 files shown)


to11mtm:
Comment without any direction:

As awkward as the convention is, should we consider cases where `.Core` is used when specific delineation is required?

(I'm fine either way just DAing here)

mtmk:
Untitled Diagram drawio

rickdotnet (approved):
Looks ok to me. 

One thing I've been meaning to dive into is looking for gaps in documentation. PRs like this could include documentation that addresses cases such as folks wanting to remain AOT-friendly. Perhaps an article or some section in the docs.

Could also include documentation like you've mentioned about where the delineation is for projects that should be AOT-friendly. Either in [CONTRIBUTING.md](https://github.com/nats-io/nats.net/blob/main/CONTRIBUTING.md) or the docs.

PR #618, opened 2024-08-25
Make NuidWriter public

### LINKED ISSUE #617: Make the NuidWriter class public
### Proposed change

Move `NuidWriter` from `NATS.Client.Core.Internal` to `NATS.Client.Core` and make the class `public` instead of `internal`.

### Use case

I use Nuids for identifiers throughout my application. The Nuid generator is public in v1 of the client, so it makes sense for it to be so in v2.

### Contribution

I'm creating this issue for tracking purposes from the [original discussion post](https://github.com/nats-io/nats.net/discussions/616), PR incoming shortly.

Madgvox:
This PR makes the `NuidWriter` class public, and moves it into the `Core` namespace out of the `Internal` namespace. 

Resolves #617.

### DIFF
--- a/src/NATS.Client.Core/Nuid.cs
+++ b/src/NATS.Client.Core/Nuid.cs
@@ -8,8 +8,15 @@ using Random = NATS.Client.Core.Internal.NetStandardExtensions.Random;
 
 namespace NATS.Client.Core;
 
+/// <summary>
+/// Represents a unique identifier generator.
+/// </summary>
+/// <remarks>
+/// The <c>Nuid</c> class generates unique identifiers that can be used
+/// to ensure uniqueness in distributed systems.
+/// </remarks>
 [SkipLocalsInit]
-public sealed class NuidWriter
+public sealed class Nuid
 {
     // NuidLength, PrefixLength, SequentialLength were nuint (System.UIntPtr) in the original code
     // however, they were changed to uint to fix the compilation error for IL2CPP Unity projects.
@@ -25,13 +32,13 @@ public sealed class NuidWriter
     private const int MaxIncrement = 333;
 
     [ThreadStatic]
-    private static NuidWriter? _writer;
+    private static Nuid? _writer;
 
     private char[] _prefix;
     private ulong _increment;
     private ulong _sequential;
 
-    private NuidWriter()
+    private Nuid()
     {
         Refresh(out _);
     }
@@ -42,16 +49,11 @@ public sealed class NuidWriter
     private static ReadOnlySpan<char> Digits => "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
 #endif
 
-    public static bool TryWriteNuid(Span<char> nuidBuffer)
-    {
-        if (_writer is not null)
-        {
-            return _writer.TryWriteNuidCore(nuidBuffer);
-        }
-
-        return InitAndWrite(nuidBuffer);
-    }
-
+    /// <summary>
+    /// Generates a new NATS unique identifier (NUID).
+    /// </summary>
+    /// <returns>A new NUID as a string.</returns>
+    /// <exception cref="InvalidOperationException">Thrown when unable to generate the NUID.</exception>
     public static string NewNuid()
     {
         Span<char> buffer = stackalloc char[22];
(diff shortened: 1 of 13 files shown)


mtmk:
Thanks @Madgvox. Can I ask you to sign your commit please? It's a CNCF requirement, and we won't be able to merge because of GitHub rules setup for this repo. You will have to force push, which should be fine. Thank you 💯

Madgvox:
> Thanks @Madgvox. Can I ask you to sign your commit please? It's a CNCF requirement, and we won't be able to merge because of GitHub rules setup for this repo. You will have to force push, which should be fine. Thank you 💯

Should be corrected now.

jasper-d:
I think it would be better change the naming as follows:

```diff
- public static class NuidWriter {
+ public static class Nuid { // or NuidGenerator if you prefer
-   public static bool TryWriteNuid(Span<char> nuidBuffer);
+   public static int Write(Span<char> destination); // Throw for inappropriate sized destination, return number of chars written
}
```

Reasoning:

1. Having `GetString()` on a type that is call "writer" seems a bit odd, calling the type `Nuid` or `NuidGenerator` avoids that issue.
2. `TryWriteNuid` does not provide information how many chars are written, this is internal knowledge of the implementation.
3. `NuidWriter.TryWriteNuid` duplicates notions of "nuid" and "write" unnecessarily
4. I would expect callers to usually know how large their buffer is, so the try-pattern seems a bit unecessary here.
5. If you want to keep the try-pattern (may possibly provide some perf-benefits with enough inlining and a suitable call-site), I'd propose `bool TryWrite(Span<char> destination, out int charsWritten)` to align better with other try-APIs in the BCL.
6. Additionally, you may want to expose `NuidLength` as `public static readonly int Length` or something, so callers can use it when allocating buffers instead of relying on undocumented implementation details.

mtmk:
> I think it would be better change the naming as follows:
> 
> ```diff
> - public static class NuidWriter {
> + public static class Nuid { // or NuidGenerator if you prefer
> -   public static bool TryWriteNuid(Span<char> nuidBuffer);
> +   public static int Write(Span<char> destination); // Throw for inappropriate sized destination, return number of chars written
> }
> ```
> 
> Reasoning:
> 
> 1. Having `GetString()` on a type that is call "writer" seems a bit odd, calling the type `Nuid` or `NuidGenerator` avoids that issue.

I'd be happy with either name

> 2. `TryWriteNuid` does not provide information how many chars are written, this is internal knowledge of the implementation.
> 3. `NuidWriter.TryWriteNuid` duplicates notions of "nuid" and "write" unnecessarily
> 4. I would expect callers to usually know how large their buffer is, so the try-pattern seems a bit unecessary here.
> 5. If you want to keep the try-pattern (may possibly provide some perf-benefits with enough inlining and a suitable call-site), I'd propose `bool TryWrite(Span<char> destination, out int charsWritten)` to align better with other try-APIs in the BCL.

This makes a lot of sense if we want to expose it. @Madgvox would you be using `TryWrite`? 

> 6. Additionally, you may want to expose `NuidLength` as `public static readonly int Length` or something, so callers can use it when allocating buffers instead of relying on undocumented implementation details.

If we want `TryWrite` it'd be good to expose `Length`

Madgvox:
I'd be using the GetString() variant most likely. We use Nuids for dynamic
subject IDs.

mtmk:
Renamed the class to `Nuid` and made TryWriteNuid internal again. we can open it up if there is need. plus xml docs. @Madgvox @jasper-d let me know if you're happy then I shall merge. thank you 💯

src/NATS.Client.Core/Nuid.cs
```diff
@@ -42,25 +49,30 @@ private NuidWriter()
...
-    public static bool TryWriteNuid(Span<char> nuidBuffer)
+    /// <summary>
+    /// Generates a new NATS unique identifier (NUID).
+    /// </summary>
+    /// <returns>A new NUID as a string.</returns>
+    /// <exception cref="InvalidOperationException">Thrown when unable to generate the NUID.</exception>
+    public static string NewNuid()
     {
-        if (_writer is not null)
+        Span<char> buffer = stackalloc char[22];
```
jasper-d, line 59: nit:

```suggestion
        Span<char> buffer = stackalloc char[NuidLength];
```

jasper-d (approved):
Code changes LGTM, did not review workflow changes.

Madgvox:
LGTM 👍

mtmk (approved):
LGTM thanks @Madgvox and @jasper-d

PR #631, opened 2024-09-16
Added Domain support for stream mirroring and sourcing and KV full support for the same.

darkwatchuk:
Added Sources and Mirror to KV Store Creation

fixes https://github.com/nats-io/nats.net/issues/494

PS https://github.com/nats-io/nats.go/pull/1112


src/NATS.Client.JetStream/Models/StreamSource.cs
```diff
@@ -54,4 +54,7 @@ public record StreamSource
     [System.Text.Json.Serialization.JsonPropertyName("external")]
     [System.Text.Json.Serialization.JsonIgnore(Condition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingDefault)]
     public ExternalStreamSource? External { get; set; }
+
+    [System.Text.Json.Serialization.JsonIgnore]
+    public string? Domain { get; set; }
```
mtmk, line 59: why do we need this? sorry not too familiar with mirroring

darkwatchuk: I took it from the go implementation here https://github.com/nats-io/nats.go/blob/a06b6a92ee7d6c70e7348bcfdb719691ba1b4b63/jsm.go#L276

I believe it's used in Jetstream domains to source data from a different domain.  E.g. creating a stream in a leaf node that's sourced from say the main cluster e.g. NGS.  

Was surprised it wasn't serialized though....

mtmk: thanks for checking other implementations. my preference would be not to pollute the json model if possible. btw I checked [js](https://github.com/nats-io/nats.deno/blob/72cd3e52626b1a239d3b45c2e159cec513c7a477/jetstream/jsapi_types.ts#L324-L330) and [rust](https://github.com/nats-io/nats.rs/blob/ea4c372ba71a5528b9c24bb122a8806cf26d9187/async-nats/src/jetstream/stream.rs#L1551) they seem to have it too 😕 cc @aricart @Jarema

darkwatchuk:
Due to the difficulty of automating leaf node testing below are examples of tests/usage of mirroring and stream sourcing.

```
   [Fact]
   public async Task Test_LeafStreamMirrorDomain()
   {
       await using var natsLeaf = new NatsConnection(new NatsOpts { Url = "nats://leaf:4223/", AuthOpts = new NatsAuthOpts { Username = "leaf", Password = "leaf" } });
       await natsLeaf.ConnectAsync();

       var jsContext = new NatsJSContext(natsLeaf);

       var mirroredStream = await jsContext.CreateStreamAsync(new StreamConfig()
       {
           Name = "mirroredStream",
           Mirror = new StreamSource
           {
               Name = "sourceStream1",
               Domain = "hub",
           },
       });
   }

   [Fact]
   public async Task Test_LeafStreamMultiSourceDomain()
   {
       await using var natsLeaf = new NatsConnection(new NatsOpts { Url = "nats://leaf:4223/", AuthOpts = new NatsAuthOpts { Username = "leaf", Password = "leaf" } });
       await natsLeaf.ConnectAsync();

       var jsContext = new NatsJSContext(natsLeaf);

       var mirroredStream = await jsContext.CreateStreamAsync(new StreamConfig()
       {
           Name = "mirroredStreamMulti",
           Sources = [
                   new StreamSource { Name = "sourceStream1", Domain = "hub" },
                   new StreamSource { Name = "sourceStream2", Domain = "hub" }
               ],
       });
   }

   [Fact]
   public async Task Test_LeafKVMirrorDomain()
   {
       await using var natsLeaf = new NatsConnection(new NatsOpts { Url = "nats://leaf:4223/", AuthOpts = new NatsAuthOpts { Username = "leaf", Password = "leaf" } });
       await natsLeaf.ConnectAsync();

       var jsContext = new NatsJSContext(natsLeaf);
       var kvContext = new NatsKVContext(jsContext);

       var mirroredStore = await kvContext.CreateStoreAsync(new NatsKVConfig("mirroredKvTestMirror")
       {
           Mirror = new StreamSource
           {
               Name = "kvTest",
               Domain = "hub",
           },
       });
   }

   [Fact]
   public async Task Test_LeafKVMirrorSourcesDomain()
   {
       await using var natsLeaf = new NatsConnection(new NatsOpts { Url = "nats://leaf:4223/", AuthOpts = new NatsAuthOpts { Username = "leaf", Password = "leaf" } });
       await natsLeaf.ConnectAsync();

       var jsContext = new NatsJSContext(natsLeaf);
       var kvContext = new NatsKVContext(jsContext);

       var mirroredStore = await kvContext.CreateStoreAsync(new NatsKVConfig("mirroredKvTestSources")
       {
           Sources = [
                   new StreamSource { Name = "kvTest", Domain = "hub" },
           ],
       });
   }
```

darkwatchuk:
Object stores don't seem to support mirroring so this has not been included as previously discussed.

src/NATS.Client.JetStream/Models/StreamSource.cs
mtmk: we can keep it same as other clients for consistency.

rickdotnet: I've just started chartering some of these waters. Tossing some tidbits out at an 80% confidence level.

If the server + node are using domains, then the typical `$JS.API` isn't used when sending traffic to the leaf. Thus, in order to mirror the stream, you need the prefix in order to receive the the `$JS` traffic.

It shouldn't need to be sent to the server (or leaf in my example) since it is already configured to use the prefix in it's configuration.

rickdotnet: I've just started chartering some of these waters. Tossing some tidbits out at an 80% confidence level.

If the server + node are using domains, then the typical `$JS.API` isn't used when sending traffic to the leaf. Thus, in order to mirror the stream, you need the prefix in order to receive the the `$JS` traffic.

It shouldn't need to be sent to the server (or leaf in my example) since it is already configured to use the prefix in it's configuration.

mtmk: correct. 'domain' field is not needed or used by the server.

mtmk:
> Due to the difficulty of automating leaf node testing below are examples of tests/usage of mirroring and stream sourcing.

* see #634

mtmk:
@rickdotnet @Jarema I wouldn't mind an extra LGTM here please since I did make a few changes myself as well 😅

rickdotnet:
Your changes LGTM. The `CreateStreamConfig()` method is a a beast. :laughing: 

Could also get a shallow copy from records using `with { }`, but that's new age trickery. :wink:

mtmk:
> The CreateStreamConfig() method is a a beast. 😆

hehe true. we can write more tests for it maybe?

> Could also get a shallow copy from records using `with { }`, but that's new age trickery. 😉

ah that's a good point. do you think we should use that to keep models clean? @darkwatchuk and I had a quick discussion about keeping them clean as well.

rickdotnet:
> do you think we should use that to keep models clean?

That'd be a by-product, but essentially. If we had a mixed back of classes/records and wanted a consistent experience, I'd opt to keep the `ShallowCopy()` method.

```cs
mirror = config.Mirror.ShallowCopy();
```
vs
```cs
mirror = config.Mirror with { };
```

Both feel good to me.

mtmk:
> > do you think we should use that to keep models clean?
> 
> That'd be a by-product, but essentially. If we had a mixed bag of classes/records and wanted a consistent experience, I'd opt to keep the `ShallowCopy()` method.
> 
> ```cs
> mirror = config.Mirror.ShallowCopy();
> ```
> 
> vs
> 
> ```cs
> mirror = config.Mirror with { };
> ```
> 
> Both feel good to me.

I think `with` syntax is good especially when there are properties updated. Updated!

rickdotnet:
Nice, looks good to me!

mtmk:
Looks like we're missing some of reference implementation details here:

https://github.com/nats-io/nats.go/blob/94fa0cb99411c202a4c9c2ae84f444a10f2e5ad8/kv.go#L1182-L1193

mtmk:
Double checked the implementation against https://github.com/nats-io/nats.go/pull/1112 (with latest changes in main) Looks good. Only, some of the mirror bucket mapping seems to be missing which is captured in #642 and can be handled in a follow up PR

PR #641, opened 2024-09-30
Adds type-forwarders NET 5.0+

### LINKED ISSUE #640: Missing Type-Forwarders
### Observed behavior

If NATS.NET is used within a .netstandard library, which is then used within a .net 8 project, using any property that has an `init` setter fails with a `MissingMethodException`

`StackExchange.Redis` had a similar issue: https://github.com/StackExchange/StackExchange.Redis/pull/2621
and the issue was discussed within the dotnet runtime project: https://github.com/dotnet/runtime/issues/96197

I have replicated the fix from `StackExchange.Redis` and tested locally. Will create a PR against this issue shortly.

### Expected behavior

Setting properties should work without throwing a `MissingMethodException`

### Server and client version

NATS.NET 2.4.0

### Host environment

_No response_

### Steps to reproduce

1. Create a .netstandard 2.0 (or 2.1) library, reference NATS.NET and write some code to set up a client
2. Reference this new library from a .NET 8.0 app and call the method that will set up the NATS client
3. Start the app

VaticanUK:
Adds missing type forwarders to fix `MissingMethodExceptions` when this library is consumed by a .NET standard library, which is then consumed by a > .NET 5 app

Resolves #640

### DIFF
--- a/src/NATS.Client.Core/Internal/netstandard.cs
+++ b/src/NATS.Client.Core/Internal/netstandard.cs
@@ -10,9 +10,11 @@
 #pragma warning disable SA1204
 #pragma warning disable SA1405
 
-#if NETSTANDARD
-
 // Enable init only setters
+#if NET5_0_OR_GREATER
+[assembly: System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Runtime.CompilerServices.IsExternalInit))]
+#elif NETSTANDARD
+
 namespace System.Runtime.CompilerServices
 {
     internal static class IsExternalInit
@@ -33,7 +35,9 @@ namespace System.Runtime.CompilerServices
     {
     }
 }
+#endif
 
+#if NETSTANDARD
 namespace System.Diagnostics
 {
     internal sealed class StackTraceHiddenAttribute : Attribute


src/NATS.Client.Core/Internal/netstandard.cs
```diff
@@ -10,9 +10,13 @@
 #pragma warning disable SA1204
 #pragma warning disable SA1405
 
-#if NETSTANDARD
-
 // Enable init only setters
+#if NET5_0_OR_GREATER
+[assembly: System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Runtime.CompilerServices.IsExternalInit))]
+[assembly: System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Runtime.CompilerServices.CallerArgumentExpressionAttribute))]
+[assembly: System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Runtime.CompilerServices.SkipLocalsInitAttribute))]
```
mtmk, line 17: Looks like `CallerArgumentExpression` and `SkipLocalsInit` are only required by the compiler and don't need to be forwarded. I can reproduce the issue when not having `IsExternalInit` but other two doesn't make any difference. Also redis library only has`IsExternalInit` forwarded. Do you think we need the other two?

VaticanUK: Redis only has an implementation for IsExternalInit, which is why it only has a forwarder for that type.

Here we have implementations for:
`IsExternalInit`
`CallerArgumentExpressionAttribute`
`SkipLocalsInitAttribute`

I'm unsure where those are used in the codebase, but if used in any codepath similar to the issue that I noticed with `IsExternalInit`, (if I understand the root cause of this correctly...) they won't exist.

Happy to remove those if preferred though since I don't know a way to break things with those forwarders missing

VaticanUK: New commit incoming

mtmk (approved):
LGTM thanks @VaticanUK

mtmk:
ouch! sorry @VaticanUK forgot to mention. Could you rebase and sign your commit please? (You will need to force push and don't worry about my commit, that can be absorbed as well) Otherwise I can't merge since it's a CNCF requirement and we have a rule for it in repo settings.

VaticanUK:
Ok, got there in the end :)

PR #644, opened 2024-10-01
Release 2.5.0-preview.1

mtmk:
* Added Domain support for stream mirroring and sourcing and KV full support for the same. (#631)
* Adds type-forwarders NET 5.0+ (#641)
* Run core tests on Windows and test run stability (#464)
* Add JetStream NATS Client Extensions (#598)


Jarema (approved):
LGTM

PR #646, opened 2024-10-07
[EXPERIMENT] Channel lock experiment

mtmk:


src/NATS.Client.Core/Commands/CommandWriter.cs
```diff
@@ -704,10 +704,7 @@ private async ValueTask ConnectStateMachineAsync(bool lockHeld, ClientOpts conne
     {
         if (!lockHeld)
         {
-            if (!await _semLock.WaitAsync(_defaultCommandTimeout, cancellationToken).ConfigureAwait(false))
-            {
-                throw new OperationCanceledException();
-            }
+            await _semLock.Writer.WriteAsync(0, cancellationToken).AsTask().WaitAsync(_defaultCommandTimeout, cancellationToken).ConfigureAwait(false);
```
to11mtm, line 707: I think a problem you may run into here is that due to the semantics of WaitAsync(), the operation may have cancelled due to timeout -but- the operation could still go through.

Work-around should be to use a CancellationTokenSource.LinkedTokenSource() and then `CancelAfter()` on that before passing that token into the `WriteAsync()`...

mtmk:
stale

PR #652, opened 2024-10-12
Add flags to NatsMsg and JetStream Request

mtmk:
This is part of preparing to reduce exception usage as stated in #581.

Note that a similar attempt was made in #275 but due to NatsMsg struct size getting larger it wasn't merged. With this PR we're using the message size fields left most two bits to set the flags relying on the fact a 30-bit field would still be large enough to hold message size int.

JetStream request changed to handle API error responses without throwing exceptions.


src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -119,20 +121,121 @@ public interface INatsMsg<T>
...
+        string? replyTo,
+        int size,
+        NatsHeaders? headers,
+        T? data,
+        INatsConnection? connection,
+        NatsMsgFlags flags = default)
+    {
+        Subject = subject;
+        ReplyTo = replyTo;
+        Size = size;
```
to11mtm, line 172: Is it worth setting `_flagsAndSize` directly here? (Sharplab is not cooperating so I can't check whether it gets optimized as is or not)

src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -207,6 +321,16 @@ internal static NatsMsg<T> Build(
         INatsDeserialize<T> serializer)
     {
         NatsHeaders? headers = null;
+        var flags = NatsMsgFlags.None;
+
+        if (payloadBuffer.Length == 0)
+        {
+            flags |= NatsMsgFlags.Empty;
+            if (NatsSubBase.IsHeader503(headersBuffer))
```
to11mtm, line 329: Should this be unnested?

to11mtm (commented):
Just some comments/etc.

src/NATS.Client.JetStream/Internal/NatsJSForcedJsonDocumentSerializer.cs
```diff
@@ -0,0 +1,21 @@
+using System.Buffers;
+using System.Text.Json;
+using NATS.Client.Core;
+
+namespace NATS.Client.JetStream.Internal;
+
+internal sealed class NatsJSForcedJsonDocumentSerializer<T> : INatsDeserialize<T>
```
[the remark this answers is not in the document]
mtmk: yea much better

src/NATS.Client.Core/NatsMsg.cs
mtmk: it's pretty specific [where server sets the 503 header for no responders](https://github.com/nats-io/nats-server/blob/290f4095a21392175a5a158167efd31cfc81f595/server/client.go#L4081) and [Go client seems to check for data being empty](https://github.com/nats-io/nats.go/blob/94fa0cb99411c202a4c9c2ae84f444a10f2e5ad8/nats.go#L4074) too. tbh I think no-responders is the only usage of 503 so i suppose it would be ok to move it out idk.

src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -119,20 +121,121 @@ public interface INatsMsg<T>
...
+
+    /// <summary>Serializable data object.</summary>
+    public T? Data { get; init; }
+
+    /// <summary>NATS connection this message is associated to.</summary>
+    public INatsConnection? Connection { get; init; }
+
+    public bool IsEmpty => (Flags & NatsMsgFlags.Empty) == NatsMsgFlags.Empty;
+
+    public bool HasNoResponders => (Flags & NatsMsgFlags.NoResponders) == NatsMsgFlags.NoResponders;
```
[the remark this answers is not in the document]
mtmk: interesting. if I changed getter to:
```csharp
public bool HasNoResponders => (_flagsAndSize & 0x80000000) != 0;
```

I get the same code:
```asm
NATS.Client.Core.dll!NATS.Client.Core.NatsMsg<T>.HasNoResponders.get():
00007FF8841C5B80  xor         eax,eax  
00007FF8841C5B82  test        dword ptr [rcx+20h],80000000h  
00007FF8841C5B89  setne       al
```

my guess is after inlined
```csharp
public bool IsEmpty => ((_flagsAndSize >> 30) & 1) == 1;
public bool HasNoResponders => ((_flagsAndSize >> 30) & 2) == 2;
```
there must be an optimization for `x & 1 == 1` only.

src/NATS.Client.JetStream/NatsJSContext.cs
```diff
@@ -285,49 +286,90 @@ internal async ValueTask<NatsJSResponse<TResponse>> JSRequestAsync<TRequest, TRe
...
-                throw new NatsJSException("No response data received");
+                return new NatsJSException("No response data received");
             }
 
-            return new NatsJSResponse<TResponse>(msg.Data, default);
+            // We need to determine what type we're deserializing into
+            // .NET 6 new APIs to the rescue: we can read the buffer once
+            // by deserializing into a document, inspect and using the new
+            // API deserialize to the final type from the document.
+            var jsonDocument = msg.Data;
```
to11mtm, line 338: There may be a concern here, due to the fact that `JsonDocument` is `IDisposable` and based on how the [Parse function for ReadOnlySequence<byte>](https://github.com/dotnet/runtime/blob/3001129ea7b0061c68176c32ba2e3520ccbb23a8/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonDocument.Parse.cs#L77-L101) looks it is a case where we are likely pulling in a buffer.

src/NATS.Client.Core/NatsMsg.cs
to11mtm: Oh nvm I guess I didn't realize they noresponders only came with empty 😅

src/NATS.Client.JetStream/NatsJSContext.cs
to11mtm: TBH I think we are fine as long as we dispose our jsondocument after getting our part, but will check if I have time this weekend.

mtmk: it's now in `using`

scottf (approved):
LGTM

PR #658, opened 2024-10-17
Add isAotCompatible build flag

mtmk:
Just to make sure we are AOT friendly in core libraries. I think we were good anyway especially with our native AOT check runs this is to be compliant with general advice: https://learn.microsoft.com/dotnet/core/deploying/native-aot#aot-compatibility-analyzers


scottf (approved):
LGTM

PR #664, opened 2024-10-22
Fix date time serialization to use ISO8601

mtmk:
Doesn't make sense to serialize datetime types with the default culture dependent formatting since it might differ between different platforms. with this PR we use ISO8601 formatting as indicated by `O` to maximize the compatibility among different platforms and languages.

### DIFF
--- a/src/NATS.Client.Core/INatsSerialize.cs
+++ b/src/NATS.Client.Core/INatsSerialize.cs
@@ -116,7 +116,7 @@ public class NatsUtf8PrimitivesSerializer<T> : INatsSerializer<T>
         {
             if (value is DateTime input)
             {
-                if (Utf8Formatter.TryFormat(input, span, out var written))
+                if (Utf8Formatter.TryFormat(input, span, out var written, new StandardFormat('O')))
                 {
                     bufferWriter.Advance(written);
                 }
@@ -133,7 +133,19 @@ public class NatsUtf8PrimitivesSerializer<T> : INatsSerializer<T>
         {
             if (value is DateTimeOffset input)
             {
-                if (Utf8Formatter.TryFormat(input, span, out var written))
+                bool result;
+                int written;
+                if (input.Offset == TimeSpan.Zero)
+                {
+                    // This will make it place `Z` instead of `+00:00` at the end
+                    result = Utf8Formatter.TryFormat(input.UtcDateTime, span, out written, new StandardFormat('O'));
+                }
+                else
+                {
+                    result = Utf8Formatter.TryFormat(input, span, out written, new StandardFormat('O'));
+                }
+
+                if (result)
                 {
                     bufferWriter.Advance(written);
                 }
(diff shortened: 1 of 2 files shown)


src/NATS.Client.Core/INatsSerialize.cs
```diff
@@ -133,7 +133,19 @@ public void Serialize(IBufferWriter<byte> bufferWriter, T value)
         {
             if (value is DateTimeOffset input)
             {
-                if (Utf8Formatter.TryFormat(input, span, out var written))
+                bool result;
+                int written;
+                if (input.Offset == TimeSpan.Zero)
+                {
+                    // This will make it place `Z` instead of `+00:00` at the end
+                    result = Utf8Formatter.TryFormat(input.UtcDateTime, span, out written, new StandardFormat('O'));
```
to11mtm, line 141: Serialization is a hot path, so it's worth noting one thing here...

[SharpLab seems to show better ASM for a `readonly` prop being referenced vs `new` even for simple case](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCEBXAMy5ilwDcNek1acefXCwAqMBBiG0AzE1IMAwgwDeNBnqYqAFMAgQANmgYBLAHYYAlAwCyhgMoAHbDYA8wAJ4YMAB8DAAmMLgYlgAi2IHSVvgwAPI8uDAYDABu2GbkjjrU+sXZ2FAMuAwAvAwAqhhcABwAYtD4cYFQMlB+rVDtGIY5eWjhkWgQHJm2mcAA7pY2MHMMrhheoWWhfQOGAOTJe/b2iiX6xADsDIa4aPMnuvoAvo96rwzuUFY5gQyw2KEIDYzH5VusbJsoNs2nEGAB9CDVBhLFZrDZbHZxfaHB5Fc5GEzmSwzRxOUhuTw+fyBEJjKIMWLxRIpNIZUp5ArvYo5cqVGr1JqYjCdbq9GGDYbkUYRKKTaZ2BjzSwI3FnPSXa63e6nZ40J5AA=)

[Looks like Ctor can throw which may prevent some inlining](https://github.com/dotnet/runtime/blob/edaa25e6942f3e8d9e2f078d99f34ed015658fbe/src/libraries/System.Private.CoreLib/src/System/Buffers/StandardFormat.cs#L56-L65) **However** for all of this pedanticism, it may or may not get handled by PGO when it's actually run. Might be worth trying to run a microbench between the two, while also noting 'real' implementation of a `readonly` based on gist could be `static` (alas, Sharplab  can't do statics easily) and likely could provide additional optimizations/benefits.

to11mtm (approved):
Nice and clean. Should we be having any sort of 'fallback' for backward compat or is it not needed?

mtmk:
> Should we be having any sort of 'fallback' for backward compat or is it not needed?

I'm guessing not many if anyone is using this feature i.e. passing `DateTime/Offset` as data. I'd consider the old implementation as a bug since it would only work both sender and receiver of the data are using the same culture. e.g. if one using US and other using UK many dates will be muddled up e.g. 10/12/2024 might mean October or December.

so I'm seeing this as a bug fix and although it might break some installations I'm assuming nobody is really using this hence it's not worth the effort and confusion trying to create compatibility. But I might be wrong - I am at least 4-5 times a day 😅

caleblloyd:
Looks good, it appears that the ISO8601 format used also [overlaps rfc3339](https://ijmacd.github.io/rfc3339-iso8601/) which is beneficial interop with other languages too

PR #670, opened 2024-11-02
Add support for auth in URL

Mrxx99:
@mtmk As discussed, here is the PR for support for Auth in URL, I also made it to support token in URL, since if it works with username + password, I think users would also expect for it to work with a token.

Aspire PR that will uses this: https://github.com/dotnet/aspire/pull/6259

Note: some Serializer unit test were already failing before I made changes.

### DIFF
--- a/src/NATS.Client.Core/Internal/NatsUri.cs
+++ b/src/NATS.Client.Core/Internal/NatsUri.cs
@@ -4,6 +4,8 @@ internal sealed class NatsUri : IEquatable<NatsUri>
 {
     public const string DefaultScheme = "nats";
 
+    private readonly Uri _redactedUri;
+
     public NatsUri(string urlString, bool isSeed, string defaultScheme = DefaultScheme)
     {
         IsSeed = isSeed;
@@ -38,6 +40,20 @@ internal sealed class NatsUri : IEquatable<NatsUri>
         }
 
         Uri = uriBuilder.Uri;
+
+        if (uriBuilder.UserName is { Length: > 0 })
+        {
+            if (uriBuilder.Password is { Length: > 0 })
+            {
+                uriBuilder.Password = "***";
+            }
+            else
+            {
+                uriBuilder.UserName = "***";
+            }
+        }
+
+        _redactedUri = uriBuilder.Uri;
     }
 
     public Uri Uri { get; }
@@ -65,7 +81,7 @@ internal sealed class NatsUri : IEquatable<NatsUri>
 
     public override string ToString()
     {
-        return IsWebSocket && Uri.AbsolutePath != "/" ? Uri.ToString() : Uri.ToString().Trim('/');
+        return IsWebSocket && Uri.AbsolutePath != "/" ? _redactedUri.ToString() : _redactedUri.ToString().Trim('/');
     }
 
     public override int GetHashCode() => Uri.GetHashCode();
--- a/src/NATS.Client.Core/Internal/NatsUri.cs
+++ b/src/NATS.Client.Core/Internal/NatsUri.cs
@@ -4,7 +4,7 @@ internal sealed class NatsUri : IEquatable<NatsUri>
 {
     public const string DefaultScheme = "nats";
 
-    private readonly Uri _redactedUri;
+    private readonly string _redacted;
 
     public NatsUri(string urlString, bool isSeed, string defaultScheme = DefaultScheme)
     {
@@ -41,6 +41,7 @@ internal sealed class NatsUri : IEquatable<NatsUri>
 
         Uri = uriBuilder.Uri;
 
+        // Redact user/password or token from the URI string for logging
         if (uriBuilder.UserName is { Length: > 0 })
         {
             if (uriBuilder.Password is { Length: > 0 })
@@ -53,7 +54,7 @@ internal sealed class NatsUri : IEquatable<NatsUri>
             }
         }
 
-        _redactedUri = uriBuilder.Uri;
+        _redacted = IsWebSocket && Uri.AbsolutePath != "/" ? uriBuilder.Uri.ToString() : uriBuilder.Uri.ToString().Trim('/');
     }
 
     public Uri Uri { get; }
@@ -79,10 +80,7 @@ internal sealed class NatsUri : IEquatable<NatsUri>
         return new NatsUri(newUri, IsSeed);
     }
 
-    public override string ToString()
-    {
-        return IsWebSocket && Uri.AbsolutePath != "/" ? _redactedUri.ToString() : _redactedUri.ToString().Trim('/');
-    }
+    public override string ToString() => _redacted;
 
     public override int GetHashCode() => Uri.GetHashCode();
 
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -290,74 +290,41 @@ public partial class NatsConnection : INatsConnection
 
     private static NatsOpts ReadUserInfoFromConnectionString(NatsOpts opts)
     {
-        var first = true;
-
-        var natsUris = opts.GetSeedUris(suppressRandomization: true);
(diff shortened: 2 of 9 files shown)


mtmk:
thanks @Mrxx99 could you please force push signed commits? it#s CNCF requirement we have the commits signed. sorry about the inconvenience :-(

mtmk:
> Note: some Serializer unit test were already failing before I made changes.

we have a few flaky tests especially on Windows but I can check if you can let me know which tests aren't passing for you

tests/NATS.Client.TestUtilities/Properties/launchSettings.json
```diff
@@ -0,0 +1,12 @@
+{
```
mtmk, line 1: do we really need this file?

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -289,6 +289,62 @@ internal ValueTask UnsubscribeAsync(int sid)
...
+
+                        uriBuilder.Password = "***"; // to redact the password from logs
+                    }
+                    else
+                    {
+                        opts = opts with
+                        {
+                            AuthOpts = opts.AuthOpts with
+                            {
+                               Token = uriBuilder.UserName,
```
mtmk, line 329: glad you handle the token as well 👍

tests/NATS.Client.TestUtilities/Properties/launchSettings.json
Mrxx99: no, this was somehow always auto generated, I had removed it and it suddenly appeared again, don't know how this is happening, but I can exclude it.

mtmk: is it Visual studio generating it?

Mrxx99: could be, I don't know, I will just exclude it from the commit

Mrxx99:
@mtmk commit is signed now

Mrxx99:
> > Note: some Serializer unit test were already failing before I made changes.
> 
> we have a few flaky tests especially on Windows but I can check if you can let me know which tests aren't passing for you

The test ist `NATS.Client.Core.Tests.SerializerTest.Utf8_serializer` and btw the ConsoleApp sandbox project also fails during serialization.
And the memory tests failed, it seems they need some external program to run.

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -289,6 +289,60 @@ internal ValueTask UnsubscribeAsync(int sid)
         return default;
     }
 
+    private static NatsOpts ReadUserInfoFromConnectionString(NatsOpts opts)
+    {
+        var first = true;
+
+        var natsUris = opts.GetSeedUris();
```
mtmk, line 296: GetSeedUris could return a random one from the list. we should check the first one provided. edit: maybe add an optional parameter to always get the original order

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -289,6 +289,60 @@ internal ValueTask UnsubscribeAsync(int sid)
...
+        var maskedUris = new List<string>(natsUris.Length);
+
+        foreach (var natsUri in natsUris)
+        {
+            var uriBuilder = new UriBuilder(natsUri.Uri);
+
+            if (uriBuilder.UserName is { Length: > 0 } username)
+            {
+                if (first)
```
mtmk, line 305: we should alter all the uris same as the first one in case passwords on multiple uris are entered.

mtmk (commented):
couple of minor comments.

mtmk:
> > > Note: some Serializer unit test were already failing before I made changes.
> > 
> > 
> > we have a few flaky tests especially on Windows but I can check if you can let me know which tests aren't passing for you
> 
> The test ist `NATS.Client.Core.Tests.SerializerTest.Utf8_serializer` and btw the ConsoleApp sandbox project also fails during serialization. And the memory tests failed, it seems they need some external program to run.

ah, you need nats-server.exe in your PATH

Mrxx99:
> > > > Note: some Serializer unit test were already failing before I made changes.
> > > 
> > > 
> > > we have a few flaky tests especially on Windows but I can check if you can let me know which tests aren't passing for you
> > 
> > 
> > The test ist `NATS.Client.Core.Tests.SerializerTest.Utf8_serializer` and btw the ConsoleApp sandbox project also fails during serialization. And the memory tests failed, it seems they need some external program to run.
> 
> ah, you need nats-server.exe in your PATH

I have, I figured this out, otherwise most tests would fail.

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -289,6 +288,65 @@ internal ValueTask UnsubscribeAsync(int sid)
...
+                        {
+                            AuthOpts = opts.AuthOpts with
+                            {
+                                Username = uriBuilder.UserName,
+                                Password = uriBuilder.Password,
+                            },
+                        };
+                    }
+
+                    uriBuilder.Password = "***"; // to redact the password from logs
```
mtmk, line 320: username needs to be set as well in case they are different.

Mrxx99: Sorry, I don't understand what you mean? Username should also be redacted when using password?

mtmk: no in case a different username is provided e.g. on the second url logs still should show the first username which is actually being used to connect.

mtmk:
fyi failing tests might be flappers. I'll rerun them.

src/NATS.Client.Core/NatsConnection.cs
Mrxx99: I hope I have fixed it how you meant it

mtmk:
@Mrxx99 just added a few tests and did a minor fix in URL user-password rewrite.

src/NATS.Client.Core/Internal/NatsUri.cs
```diff
@@ -65,7 +81,7 @@ public NatsUri CloneWith(string host, int? port = default)
 
     public override string ToString()
     {
-        return IsWebSocket && Uri.AbsolutePath != "/" ? Uri.ToString() : Uri.ToString().Trim('/');
+        return IsWebSocket && Uri.AbsolutePath != "/" ? _redactedUri.ToString() : _redactedUri.ToString().Trim('/');
```
to11mtm, line 84: Suggestion: If the redacted `Uri` is only being used here, IMO we would be best of just making the right string for it based on other args for Ctor vs keeping a separate URI instance and `.ToString()`ing it every call.

mtmk:
@Mrxx99 I moved redaction to NatsUri class to keep the original URLs since I suspect some people might be using this to go through WebSocket proxies.

mtmk (approved):
LGTM thanks @Mrxx99

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -76,7 +75,7 @@ public NatsConnection()
     public NatsConnection(NatsOpts opts)
     {
         _logger = opts.LoggerFactory.CreateLogger<NatsConnection>();
-        Opts = opts;
+        Opts = ReadUserInfoFromConnectionString(opts);
```
scottf, line 78: I'm not a fan of this form. You are both modifying the value and returning/assigning it. I'd prefer a "procedure" versus a function. Maybe a private setter?

scottf (changes requested):

src/NATS.Client.Core/NatsConnection.cs
mtmk: since we're not using a builder it'd be quite a big change trying to create `NatsOpts` once. so the pattern here is using the immutable `opts` object and a pure function `ReadUserInfoFromConnectionString()` to return a new `opts` object with the desired properties set to new values.

PR #674, opened 2024-11-09
open Header.Writer class and create GetBytesLength method

### LINKED ISSUE #638: Expose a public API for getting NatsHeaders in the byte array (serialization)
### Proposed change

Such kind of methods :
 - ```public byte[] NatsHeaders.SerializeToArray()``` or ```public byte[] NatsHeaders.GetSerialized()``` - Returns the byte array in a serialized version (ASCII or any other encoding).
-  ```public int SerializedLength()``` - Returns the number of bytes in a serialized version (ASCII or any other encoding).

### Use case

Need to calculate/take into account NatsHeaders size (length in bytes) up front to call ```Publish``` for large Nats message which exceeds payload size against server's maximum payload size. 

### Contribution

_No response_

SiuTung08:
Open Header.Writer class and create GetBytesLength method for calculating NatsHeaders size (length in bytes) up front to call Publish for large Nats message which may exceeds payload size against server's maximum payload size.

* fixes #638

### DIFF
--- a/src/NATS.Client.Core/Internal/HeaderWriter.cs
+++ b/src/NATS.Client.Core/Internal/HeaderWriter.cs
@@ -1,11 +1,10 @@
 using System.Buffers;
 using System.Text;
 using NATS.Client.Core.Commands;
-using NATS.Client.Core.Internal;
 
-namespace NATS.Client.Core;
+namespace NATS.Client.Core.Internal;
 
-public class HeaderWriter
+internal class HeaderWriter
 {
     private const byte ByteCr = (byte)'\r';
     private const byte ByteLf = (byte)'\n';
@@ -21,7 +20,7 @@ public class HeaderWriter
 
     private static ReadOnlySpan<byte> ColonSpace => new[] { ByteColon, ByteSpace };
 
-    public long GetBytesLength(NatsHeaders headers)
+    internal static long GetBytesLength(NatsHeaders headers, Encoding encoding)
     {
         var len = CommandConstants.NatsHeaders10NewLine.Length;
         foreach (var kv in headers)
@@ -31,14 +30,14 @@ public class HeaderWriter
                 if (value != null)
                 {
                     // key length
-                    var keyLength = _encoding.GetByteCount(kv.Key);
+                    var keyLength = encoding.GetByteCount(kv.Key);
                     len += keyLength;
 
                     // colon space length
                     len += ColonSpace.Length;
 
                     // value length
-                    var valueLength = _encoding.GetByteCount(value);
+                    var valueLength = encoding.GetByteCount(value);
                     len += valueLength;
 
                     // CrLf length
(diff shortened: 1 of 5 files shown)


SiuTung08:
> @SiuTung08 should we provide a convenience method on the header class so you can call it like `headers.GetBytesLength()`?

Make sense!  Just pushed the commit for this.

SiuTung08:
@mtmk looks like the test jobs for Linux (latest) failed.  Looking at the actual results though it seems that all tests have passed.  Can we retrigger it?

src/NATS.Client.Core/NatsHeaders.cs
```diff
@@ -294,6 +294,16 @@ public void CopyTo(KeyValuePair<string, StringValues>[] array, int arrayIndex)
         }
     }
 
+    /// <summary>
+    /// Returns the bytes length of the header
+    /// </summary>
+    /// <param name="headerWriter">an instance of headerWriter</param>
+    /// <returns>bytes length of th header</returns>
+    public long GetBytesLength(HeaderWriter headerWriter)
```
mtmk, line 302: just thinking about it from the application developer standpoint: is there a way we can make this taking an optional encoding rather than the writer? maybe in that case we won't even need to expose a writer? e.g. have an internal static GrtBytesLength(Encoding) method on the writer where we can reuse here?
```csharp
public long GetBytesLength(Encoding? encoding = null)
{
    // e.g. if null set to utf-8
```
what do you think?

SiuTung08: @mtmk I agree, I will make some changes.

mtmk (approved):
LGTM thanks @SiuTung08

src/NATS.Client.Core/Internal/HeaderWriter.cs
```diff
@@ -37,10 +67,7 @@ internal long Write(IBufferWriter<byte> bufferWriter, NatsHeaders headers)
                     var keySpan = bufferWriter.GetSpan(keyLength);
                     _encoding.GetBytes(kv.Key, keySpan);
                     if (!ValidateKey(keySpan.Slice(0, keyLength)))
-                    {
-                        throw new NatsException(
-                            $"Invalid header key '{kv.Key}': contains colon, space, or other non-printable ASCII characters");
-                    }
+                        throw new NatsException($"Invalid header key '{kv.Key}': contains colon, space, or other non-printable ASCII characters");
```
to11mtm, line 70: 1. @mtmk correct me if I'm off here, do we prefer including braces?
2.  If possible, this would be a good opportunity to wrap this `throw` into a helper method, e.x.

```
if (!ValidateKey(keySpan.Slice(0, keyLength)))
{
    NatsException.Throw($"Invalid header key '{kv.Key}': contains colon, space, or other non-printable ASCII characters")
}

public class NatsException : Exception
{
   //etc
   public static void Throw(string message)
   {
      throw new NatsException(message);
   }
   //etc
}
```

(There may be a throwhelper already floating around, if so prefer that)

Main advantage of such is we can get some better inlining/opts and avoid certain callstack checks on happy path....

mtmk: ouch just merged it. I shall open a new PR

PR #675, opened 2024-11-11
Fixed an exception which happens when PutAsync is used more than once and activity logging is enabled in main project

uhfath:
Fixes #672

PR #678, opened 2024-11-13
Release 2.5.4

mtmk:
* Add OnSocketAvailableAsync Hook (#647)
* Refactor exception handling in NatsException (#677)
* Fixed an exception which happens when PutAsync is used more than once and activity logging is enabled in main project (#675)
* open Header.Writer class and create GetBytesLength method (#674)
* Fix date time serialization to use ISO8601 (#664)


scottf (approved):
LGTM

PR #688, opened 2024-12-04
Add KV try get

mtmk:
`KVBench.cs`:

| Method      | Mean     | Error   | StdDev  | Ratio | RatioSD | Gen0   | Gen1   | Allocated | Alloc Ratio |
|------------ |---------:|--------:|--------:|------:|--------:|-------:|-------:|----------:|------------:|
| TryGetAsync | 150.3 us | 2.87 us | 2.95 us |  0.89 |    0.02 | 0.2441 | 0.2441 |    4.5 KB |        0.82 |
| GetAsyncNew | 170.5 us | 3.33 us | 3.56 us |  1.01 |    0.03 | 0.2441 |      - |   5.78 KB |        1.05 |
| GetAsync    | 168.3 us | 2.82 us | 2.50 us |  1.00 |    0.02 | 0.2441 | 0.2441 |   5.49 KB |        1.00 |

`TryGet` is performing much better.

the new `Get` (which is using `TryGet`) is slightly worst because of the additional async/await state machine I'm guessing.

cc @srs-adamr @to11mtm

* see also #581


srs-adamr:
You're still throwing an `Exception` in your benchmark for TryGet's, get rid of that and then test it! 🤔

srs-adamr:
I am surprised the performance is not better than that 🤔, hmmm.

srs-adamr (approved):
Looks good!

to11mtm:
I'll give this a look tonight @mtmk

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -190,44 +219,44 @@ public async ValueTask<NatsKVEntry<T>> GetEntryAsync<T>(string key, ulong revisi
             if (direct is { Headers: { } headers } msg)
             {
                 if (headers.Code == 404)
```
to11mtm, line 221: TBH stuffing this synchronous validation that fetches a value or two, probably should get moved into static method.  (AFAIK this can be moved to static although the return may change due to the sequence parsing).

That will improve overall state machine quality and likely improve performance as a result. I peeked at the IL for something close to current main and there's lots of locals and large state machines involved. Sometimes moving certain parts of logic to a static or even separate method can help with that.

Additionally, it's probably worth at least filing an issue to see whether splitting these 'if/else' branches into sub-calls gives a performance or memory boost due to likelyhood State machine will allocate for things only in one path...

to11mtm:
> the new Get (which is using TryGet) is slightly worst because of the additional async/await state machine I'm guessing.

Async await and Delegate state captures are _very_ interesting.  AFAIR for example, if you've got multiple anonymous delegates (e.x. `myList.Where(a=> a.Foo == this.CaptureOne).Where(a=>a.Bar== captureBar).Where(a.FooBar == captureFooBar)`) the compiler may prefer to re-use the same 'capture' type between instances (i.e. it's not re-using instances, but it keeps number of compiler generated types down), so your delegates can allocate more memory. In _extreme_ cases, splitting synchronous logic out into separate calling methods was sometimes useful, both for this and to save on locals.

to11mtm (approved):
I'm OK with it, My biggest concern would be my comments about `EnsureSuccess` and how to use, since that method would be exposed API.

The other stuff can be handled via future PR if needed, but hopefully is good food for thought for possible perf/mem improvements

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -166,13 +174,34 @@ public ValueTask PurgeAsync(string key, NatsKVDeleteOpts? opts = default, Cancel
...
+#if NET8_0_OR_GREATER
+        var keySubject = string.Create(key.Length + Bucket.Length + 5, (Bucket, key), static (span, state) =>
+        {
+            "$KV.".CopyTo(span);
+            state.Bucket.CopyTo(span[4..]);
+            span[state.Bucket.Length + 4] = '.';
+            state.key.CopyTo(span[(state.Bucket.Length + 5)..]);
+        });
+#else
         var keySubject = $"$KV.{Bucket}.{key}";
```
to11mtm, line 201: Another opportunity to use `_keySubjectRequestBeginning` to lower copies even if we keep string.Create pattern, (BTW name is trash feel free to pick a better one :))

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -166,13 +174,34 @@ public ValueTask PurgeAsync(string key, NatsKVDeleteOpts? opts = default, Cancel
...
         ValidateKey(key);
         serializer ??= JetStreamContext.Connection.Opts.SerializerRegistry.GetDeserializer<T>();
 
-        var request = new StreamMsgGetRequest();
+#if NET8_0_OR_GREATER
+        var keySubject = string.Create(key.Length + Bucket.Length + 5, (Bucket, key), static (span, state) =>
+        {
+            "$KV.".CopyTo(span);
+            state.Bucket.CopyTo(span[4..]);
+            span[state.Bucket.Length + 4] = '.';
```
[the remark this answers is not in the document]
to11mtm: Actually, thinking about it more, I'd ask we double-check the perf benefit of this ifdef vs pre-net8 code with `_keySubjecttRequestBeginning` in the pattern, i.e. maybe once we have something like that the interpolated builder with normal buffer will work well enough? IDK but if they both stay close, might be worth not having the ifdefs.

to11mtm (commented):
More comments

src/NATS.Client.KeyValueStore/NatsKVStore.cs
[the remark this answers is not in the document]
mtmk: very good points as usual! simple concat seems to be the winner. preparing the prefix in .ctor defo making a huge difference here. checkout my 'Bench string' commit for the benchmarks

| Method       | Mean      | Error     | StdDev    | Ratio | Gen0   | Gen1   | Allocated | Alloc Ratio |
|------------- |----------:|----------:|----------:|------:|-------:|-------:|----------:|------------:|
| StringOrig   | 18.749 ns | 0.0381 ns | 0.0356 ns |  1.00 | 0.0058 | 0.0000 |      80 B |        1.00 |
| StringInter  |  8.906 ns | 0.0900 ns | 0.0798 ns |  0.48 | 0.0058 | 0.0000 |      80 B |        1.00 |
| StringConcat |  8.829 ns | 0.0459 ns | 0.0429 ns |  0.47 | 0.0058 | 0.0000 |      80 B |        1.00 |
| StringCreate | 14.292 ns | 0.0723 ns | 0.0641 ns |  0.76 | 0.0058 | 0.0000 |      80 B |        1.00 |

src/NATS.Client.Core/NatsResult.cs
```diff
@@ -29,6 +29,13 @@ public NatsResult(Exception error)
 
     public static implicit operator NatsResult<T>(Exception error) => new(error);
 
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public void EnsureSuccess()
+    {
+        if (_error != null)
+            throw _error;
```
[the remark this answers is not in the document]
mtmk: removed the public method and replaced with helper

src/NATS.Client.KeyValueStore/NatsKVStore.cs
mtmk: moved to #691

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -166,13 +176,27 @@ public ValueTask PurgeAsync(string key, NatsKVDeleteOpts? opts = default, Cancel
 
     /// <inheritdoc />
     public async ValueTask<NatsKVEntry<T>> GetEntryAsync<T>(string key, ulong revision = default, INatsDeserialize<T>? serializer = default, CancellationToken cancellationToken = default)
+    {
+        var result = await TryGetEntryAsync(key, revision, serializer, cancellationToken);
+        if (!result.Success)
+        {
+            ThrowException(result.Error);
+        }
```
to11mtm, line 184: Nice and concise, love it!

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -44,13 +44,23 @@ public class NatsKVStore : INatsKVStore
     private const string NatsSequence = "Nats-Sequence";
     private const string NatsTimeStamp = "Nats-Time-Stamp";
     private static readonly Regex ValidKeyRegex = new(pattern: @"\A[-/_=\.a-zA-Z0-9]+\z", RegexOptions.Compiled);
```
to11mtm, line 46: Side note, have we looked into whether it's worth doing sourcegen regex here for NET8+? If not I can file issue to investigate further.

to11mtm (approved):
Reviewed general cleanup, looks awesome!

src/NATS.Client.KeyValueStore/NatsKVStore.cs
mtmk: i didn't think of that actually we should please do

srs-adamr: You can do source generated as well. 

```
[GeneratedRegex(@"\A[-/_=.a-zA-Z0-9]+\z", RegexOptions.Compiled)]
private static partial Regex ValidKeyRegex();
```

PR #689, opened 2024-12-04
Add NatsClient to DI

mtmk:
This change modifies the default serializer and pending policy, hence it is technically a breaking change.

### DIFF
--- a/src/NATS.Client.Simplified/NATS.Client.Simplified.csproj
+++ b/src/NATS.Client.Simplified/NATS.Client.Simplified.csproj
@@ -20,4 +20,7 @@
         <ProjectReference Include="..\NATS.Client.Serializers.Json\NATS.Client.Serializers.Json.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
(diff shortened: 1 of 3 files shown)


rickdotnet:
These changes look good to me.

Does it make sense to keep `AddJsonSerialization(...)` around anymore? Seems confusing, and since it now defaults to JSON, I'm not sure that it's needed. To keep support for `JsonSerializerContext`, we could add some documentation around using `NatsJsonContextSerializerRegistry` and `WithSerializerRegistry()`.

mtmk:
> These changes look good to me.
> 
> Does it make sense to keep `AddJsonSerialization(...)` around anymore? Seems confusing and since it now defaults to JSON, I'm not sure that it's needed. There is documentation for [NatsJsonContextSerializerRegistry](https://nats-io.github.io/nats.net/documentation/advanced/serialization.html#using-json-serializer-context), could expand on that, if necessary. That combined with `WithSerializerRegistry()` should be the same functionality. If you wanted to keep the method, I'd still suggest renaming it to `WithJsonSerializerContext(params JsonSerializerContext[])` and track it in a similar way and take out the chained options.

@rickdotnet good point. I know I marked this as breaking already (actually not breaking compiler but behaviour) but do you think we'd upset a lot of code bases if we removed the method. hmm maybe obsolete the old one and keep it around for a while whether we introduce a new method or not?

rickdotnet:
Safest bet, in my opinion, would be to mark it as obsolete and include a nudge in the right direction, either a new method or a link to docs.

mtmk:
> Safest bet, in my opinion, would be to mark it as obsolete and include a nudge in the right direction, either a new method or a link to docs.

@rickdotnet sorry for the delay, this is done. let me know if you like the text goes with it.

scottf (approved):
LGTM

PR #703, opened 2025-01-07
Initial implementation of per-key-TTL for KV

stebet:
> [!NOTE]
> * to be merged into #733 

* Adds support to set the `AllowMsgTTL` flag for KV streams
* Adds a `Timestamp ttl` optional parameter to the `PutAsync<T>` method on `INatsKVStore`
  * Note that this does break API backwards compatibility if users are using their own serializers, but that's seems to be how it's done in Go client, so it's up to you if you want to create a separate overload for each method that can accept a TTL value
* Added a single test to test the TTL functionality. This currently requires the tests to have access to the latest nightly binary

TODO:
* More exhaustive tests
* Add support to the `CreateAsync/UpdateAsync` API methods

see also [ADR-43](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-43.md)

### DIFF
--- a/NATS.Net.sln
+++ b/NATS.Net.sln
@@ -17,23 +17,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NATS.Client.Core.Tests", "t
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MicroBenchmark", "sandbox\MicroBenchmark\MicroBenchmark.csproj", "{A10F0D89-13F3-49B3-ACF7-66E45DC95225}"
 EndProject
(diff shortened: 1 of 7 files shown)


mtmk:
we're hoping to merge this into server 2.11 branch before server release

jodydonetti:
Can't wait for per-key TTL, it will allow usage of NATS in FusionCache too.

Thanks for all the hard work!

mtmk (approved):
LGTM thanks @stebet 

There are still details to be worked out e.g. around `SubjectDeleteMarkerTTL` but this PR introduces support for per-message/key TTL feature for new server 2.11 soon to be released. We can have follow up PRs into the server-2.11-features branch when we have more implementation details available.

PR #706, opened 2025-01-08
Make sure we also install .net 6 for the tests

stebet:
.NET 6 was missing so it didn't run those tests and failed the CI builds.


mtmk (approved):
LGTM thanks @stebet

PR #708, opened 2025-01-08
Disable the Windows Memory Tests

stebet:
The Memory Tests seem to hang so let's skip them for now


mtmk (approved):
I agree we can disable the memory tests until we find the reason so that rest of the functional tests can run on windows too

LGTM thanks @stebet

PR #712, opened 2025-01-13
Add callback to NatsAuthOpts that allows refreshing a Token 

### LINKED ISSUE #356: Allow updating the Connection's token and/or JWT upon disconnect
### Proposed change

Support using the disconnected callback to refresh the token or JWT used to authenticate to allow reconnecting without the need to create a new connection and subscribe to all the topic again.
Something like this (working snippet using the V1 client):
```
opts.DisconnectedEventHandler += (sender, args) =>
{
    args.Conn.Opts.Token = RefreshToken();
};

opts.Token = GetToken();
using IConnection c = new ConnectionFactory().CreateConnection(opts);
``` 

### Use case

We have a long running application that subscribes to several NATS topics.
The authentication in our scenario is made using tokens, with expiration.
In order to allow the reconnect of the same connection upon disconnect, we need to provide a valid token, so it might need to be refreshed due to expiry date.
This proposal would void the need to create a new connection and re-subscribe to all topics, 

### Contribution

Yes, I'm glad to help (might lack some context on the internals though)

garrett-sutton:
fixes #356


garrett-sutton:
@mtmk are you the right person to review this? Or is there someone else that might be good to reach out to? Thanks!

src/NATS.Client.Core/Internal/UserCredentials.cs
```diff
@@ -31,6 +32,8 @@ public UserCredentials(NatsAuthOpts authOpts)
 
     public string? Token { get; }
 
+    public Func<Task<string>>? TokenHandler { get; }
```
mtmk, line 35: to be consistent with other callbacks should we make this `ValueTask`. Can't remember if we also pass cancellation token. Also what about JWT? would that benefit from a callback like this?

mtmk:
@garrett-sutton could you sign your commits please?

src/NATS.Client.Core/Internal/UserCredentials.cs
garrett-sutton: I modified my initial commit to be a `ValueTask` rather than a `Task`: 21c66540b10ccf0aee4f2771ee53374509e3eb64
From what I can tell in other places, it doesn't appear that a cancellation token is passed. 

In 58864dc1663a6e774a9b8e8839b2def3e5b2ff9e, I added callbacks for JWT, NKey, and seed. Based on my understanding, if the JWT is being refreshed, the seed also needs to be refreshed. And the same goes with NKeys. I added unit tests to show that the callbacks will take precedence over any raw values that are set on initialization of NatsAuthOpts.

garrett-sutton:
> @garrett-sutton could you sign your commits please?

Done. As a note, this should probably be called out in `CONTRIBUTING.md`. I missed this initially because I was only looking there.

src/NATS.Client.Core/Internal/UserCredentials.cs
```diff
@@ -25,30 +30,40 @@ public UserCredentials(NatsAuthOpts authOpts)
...
+    public Func<ValueTask<string>>? SeedHandler { get; }
+
     public string? NKey { get; }
 
+    public Func<ValueTask<string>>? NKeyHandler { get; }
+
     public string? Token { get; }
 
-    public string? Sign(string? nonce)
+    public Func<ValueTask<string>>? TokenHandler { get; }
```
mtmk, line 45: API usability point: should there be a single callback rather than individual ones? would there be a case where application might want to change its auth method? should we also pass in the host, is that helpful in any way?

darkwatchuk: +1 for single callback on this....

caleblloyd:
The most recent options callback I reviewed did supply a URI and a CancellationToken as an argument

https://github.com/nats-io/nats.net/blob/caa4befc6a4ba0c259d3c27679b57ed64eb4e1fe/src/NATS.Client.Core/NatsWebSocketOpts.cs#L24-L28

> API usability point: should there be a single callback rather than individual ones?

Sounds like a good idea, a signature could be

In `NatsAuthOpts`:

```csharp
   Func<Uri, CancellationToken, ValueTask<NatsAuthOpts>? Callback { get; init; } = null;
```

This would have to come with the caveat that a Callback could not return another `NatsOptsAuth` with a Callback

Or in `NatsOpts`:

```csharp
   Func<Uri, CancellationToken, ValueTask<NatsAuthOpts>? AuthOptsCallback { get; init; } = null;
```

Could also be a slipper slope though, what other options could be updated between reconnects?  And which ones could be updated after knowing the URI that will be connected to?  Is it worth putting just the auth ones in now, or coming up with a broader approach to allow for updating all potential options

mtmk:
Yes we have started to pass in CT in our callbacks only recently so we should carry on with that. I agree passing back the whole NatsAuthOpts is tricky. I was thinking of a simpler enum with a new type:
```csharp
enum NatsAuthType { Token, Jwt, NKey, Seed }

struct NatsAuthKey(NatsAuthType type, string Value)

Func<Uri, CancellationToken, ValueTask<NatsAuthKey>? AuthKeyCallback { get; init; } = null;
```

I feel name `AuthKey` may not be the name we end up with but this structure would be my proposal.

garrett-sutton:
> Yes we have started to pass in CT in our callbacks only recently so we should carry on with that. I agree passing back the whole NatsAuthOpts is tricky. I was thinking of a simpler enum with a new type:
> 
> ```cs
> enum NatsAuthType { Token, Jwt, NKey, Seed }
> 
> struct NatsAuthKey(NatsAuthType type, string Value)
> 
> Func<Uri, CancellationToken, ValueTask<NatsAuthKey>? AuthKeyCallback { get; init; } = null;
> ```
> 
> I feel name `AuthKey` may not be the name we end up with but this structure would be my proposal.

I like this approach. One follow-up question though. For using something like this, do we actually need to return an array of NatsAuthKey from the callback? 

I expect that if the JWT or the NKey change that the Seed should also change. Is that correct? If so, I think we need to provide a way for implementers to specify that multiple things need to be updated.

mtmk:
> > Yes we have started to pass in CT in our callbacks only recently so we should carry on with that. I agree passing back the whole NatsAuthOpts is tricky. I was thinking of a simpler enum with a new type:
> > ```cs
> > enum NatsAuthType { Token, Jwt, NKey, Seed }
> > 
> > struct NatsAuthKey(NatsAuthType type, string Value)
> > 
> > Func<Uri, CancellationToken, ValueTask<NatsAuthKey>? AuthKeyCallback { get; init; } = null;
> > ```
> >   
> > I feel name `AuthKey` may not be the name we end up with but this structure would be my proposal.
> 
> I like this approach. One follow-up question though. For using something like this, do we actually need to return an array of NatsAuthKey from the callback?
> 
> I expect that if the JWT or the NKey change that the Seed should also change. Is that correct? If so, I think we need to provide a way for implementers to specify that multiple things need to be updated.

you're right. we'd need to pass seed/secret next to value. how about this?
```csharp
enum NatsAuthType { Token, Jwt, NKey, UserInfo }
 
struct NatsAuthKey(NatsAuthType type, string Value, string Secret)
```

caleblloyd:
> you're right. we'd need to pass seed/secret next to value. how about this?

I think the combination of things they may want to supply is:
- `Username` + `Password` at same time
- `Token`
- `Seed` (from which we can derive the `NKey`, so  no need to really supply the `NKey`)
- `Jwt` + `Seed` at same time
- `NkeyFile`
- `CredsFile`
- No Auth

Too bad [TypeUnions](https://github.com/dotnet/csharplang/blob/main/proposals/TypeUnions.md) aren't in C# yet, it'd be nice to have records for

- `NatsAuthenticator.UsernamePassword(string Username, string Password)`
- `NatsAuthenticator.Token(string Token)`
- `NatsAuthenticator.NKey(string Seed)`
- `NatsAuthenticator.Jwt(string Jwt, string Seed)`
- `NatsAuthenticator.NkeyFile(string NKeyFile)`
- `NatsAuthenticator.CredsFile(string CredsFile)`

src/NATS.Client.Core/Internal/UserCredentials.cs
garrett-sutton: I used your latest suggestion in the other thread in beedf4291d6ef2054b6c3326769e4f96c41e3099 to consolidate things into a single callback. 

I ended up calling the new structure a `NatsAuthCred` and largely chose that because you mentioned you weren't thrilled about the idea of `NatsAuthKey`. I'm open to other ideas on it though.

src/NATS.Client.Core/Internal/UserCredentials.cs
```diff
@@ -31,24 +33,53 @@ public UserCredentials(NatsAuthOpts authOpts)
...
+            switch (a.Type)
+            {
+            case NatsAuthType.Jwt:
+                opts.JWT = a.Value;
+                seed = a.Seed;
+                break;
+            case NatsAuthType.Nkey:
+                opts.NKey = a.Value;
+                seed = a.Seed;
+                break;
```
caleblloyd, line 69: NKeys and Seeds are a pair.  It was probably a mistake in the initial implementation to require passing both, as the NKey can be derived from the seed.  I guess it acts as a discriminator so we can tell it apart from JWT + Seed

src/NATS.Client.Core/Internal/UserCredentials.cs
```diff
@@ -31,24 +33,53 @@ public UserCredentials(NatsAuthOpts authOpts)
...
+        {
+            using var cts = new CancellationTokenSource(timeout);
+            var a = await AuthCredCallback(uri.Uri, cts.Token).ConfigureAwait(false);
+
+            switch (a.Type)
+            {
+            case NatsAuthType.Jwt:
+                opts.JWT = a.Value;
+                seed = a.Seed;
+                break;
```
caleblloyd, line 65: JWTs and Seeds are a pair, the Sub in the JWT is the public NKey for the seed

mtmk:
> > you're right. we'd need to pass seed/secret next to value. how about this?
> 
> I think the combination of things they may want to supply is:
> 
> * `Username` + `Password` at same time
> * `Token`
> * `Seed` (from which we can derive the `NKey`, so  no need to really supply the `NKey`)
> * `Jwt` + `Seed` at same time
> * `NkeyFile`
> * `CredsFile`
> * No Auth
> 
> Too bad [TypeUnions](https://github.com/dotnet/csharplang/blob/main/proposals/TypeUnions.md) aren't in C# yet, it'd be nice to have records for
> 
> * `NatsAuthenticator.UsernamePassword(string Username, string Password)`
> * `NatsAuthenticator.Token(string Token)`
> * `NatsAuthenticator.NKey(string Seed)`
> * `NatsAuthenticator.Jwt(string Jwt, string Seed)`
> * `NatsAuthenticator.NkeyFile(string NKeyFile)`
> * `NatsAuthenticator.CredsFile(string CredsFile)`

that's good analysis! I'm happy with the name `NatsAuthenticator`. we can achieve the same syntax using the above enum + struct combo with these static helpers. One question about JWTs: does the seed change as well or just JWT in which case we could have an overload for that:

```csharp
enum NatsAuthType { None, Token, Jwt, UserInfo, CredFile .... }

struct NatsAuthenticator
{
    NatsAuthType _type;
    string? _value;
    string? _secret;

    private NatsAuthenticator(NatsAuthType, string?, string?) { ... }
    
    public static NatsAuthenticator NoAuth() { ... }

    public static NatsAuthenticator UsernamePassword(string Username, string Password) { ... }

    public static NatsAuthenticator Token(string Token) { ... }

    public static NatsAuthenticator Jwt(string Jwt, string Seed) { ... }

    public static NatsAuthenticator Jwt(string Jwt) { ... } // maybe?

    ...
}


```

src/NATS.Client.Core/Internal/UserCredentials.cs
```diff
@@ -31,24 +33,53 @@ public UserCredentials(NatsAuthOpts authOpts)
...
+    internal async Task AuthenticateAsync(ClientOpts opts, ServerInfo? info, NatsUri uri, TimeSpan timeout)
     {
-        opts.JWT = Jwt;
-        opts.NKey = NKey;
-        opts.AuthToken = Token;
-        opts.Sig = info is { AuthRequired: true, Nonce: { } } ? Sign(info.Nonce) : null;
+        string? seed = null;
+        if (AuthCredCallback != null)
+        {
+            using var cts = new CancellationTokenSource(timeout);
```
mtmk, line 57: We should take in CT from above as param and use it here. if we need the timeout we should link the tokens. The idea is if we're e.g. shutting down callbacks should get the signal as well and quit whatever they might be doing and not hanging.

src/NATS.Client.Core/NatsAuthOpts.cs
```diff
@@ -1,5 +1,21 @@
 namespace NATS.Client.Core;
 
+public enum NatsAuthType
+{
+    Token,
+    Jwt,
+    Nkey,
+}
+
+public struct NatsAuthCred(NatsAuthType type, string value, string seed)
```
mtmk, line 10: We should use a private .ctor and hide the fields. We can expose static factory methods as public only to minimize the API surface not exposing internals.

src/NATS.Client.Core/Internal/UserCredentials.cs
garrett-sutton: In the latest, I used the "add a seed" and we can derive the nkey from it approach: 21c5e74a8ff2bac5f45405d10eb7c43f94c93515

src/NATS.Client.Core/Internal/UserCredentials.cs
garrett-sutton: In the static factory methods, I implemented it such that you need to input the jwt and seed for jwt auth but only the seed for nkey auth: 21c5e74a8ff2bac5f45405d10eb7c43f94c93515

src/NATS.Client.Core/NatsAuthOpts.cs
garrett-sutton: This is done in 21c5e74a8ff2bac5f45405d10eb7c43f94c93515

src/NATS.Client.Core/Internal/UserCredentials.cs
garrett-sutton: A question about this. I don't necessarily object to this comment, but I'm not sure of whether 1) we need the timeout and if we do 2) what token to link it to. 3) Otherwise, I'm not sure if it is just OK to pass in `CancellationToken.None` to this method?

Do you have thoughts on these points?

I took the current approach because it seems to be what other AuthenticateAsync type methods did (i.e. `sslConnection.AuthenticateAsClientAsync`

mtmk: You're right we haven't been consistent across the board:
```csharp
// None
public Func<SslClientAuthenticationOptions, ValueTask>? ConfigureClientAuthentication { get; init; }

// Uses CTS with Connection timeout - but not passing in dispose token from connection object
public Func<Uri, ClientWebSocketOptions, CancellationToken, ValueTask>? ConfigureClientWebSocketOptions { get; init; }
```

I think we should start using the CT from connection dispose to have a clean shutdown. Should we pass `_disposedCancellationTokenSource` into Authenticate [here](https://github.com/nats-io/nats.net/pull/712/files#diff-aaf322e0ed00d67e3106e54d66e4f71704eeefcb79350aa24d760251a82aad20R464-R467):
```csharp
await _userCredentials.AuthenticateAsync(_clientOpts, WritableServerInfo, _currentConnectUri, Opts.ConnectTimeout, _disposedCancellationTokenSource.Token).ConfigureAwait(false);
```

then we can link them before passing to the callback [here](https://github.com/nats-io/nats.net/pull/712/files#diff-d1f1dfc7f712aaf1e0287005b7a0d8f8ae6592d24fcc1adb58b2a137efc8fb96R58-R59) e.g.:
```csharp
using var cts = new CancellationTokenSource(timeout);

#if NETSTANDARD

using var ctr = cancellationToken.Register(static state => ((CancellationTokenSource)state!).Cancel(), cts);

#else

await using var ctr = cancellationToken.UnsafeRegister(static state => ((CancellationTokenSource)state!).Cancel(), cts);

#endif

var authCred = await AuthCredCallback(uri.Uri, cts.Token).ConfigureAwait(false);
```

garrett-sutton: Ok, that makes sense! Thank you for the help on this. Those changes are reflecting in the most recent update of the commit with functional changes: 90880408f8a477ee88dc68266a394b7b49af4eff

src/NATS.Client.Core/Internal/ClientOpts.cs
```diff
@@ -56,11 +56,11 @@ private ClientOpts(NatsOpts opts)
 
     /// <summary>Connection username (if auth_required is set)</summary>
     [JsonPropertyName("user")]
-    public string? Username { get; init; } = null;
+    public string? Username { get; set; } = null;
 
     /// <summary>Connection password (if auth_required is set)</summary>
     [JsonPropertyName("pass")]
-    public string? Password { get; init; } = null;
+    public string? Password { get; set; } = null;
```
caleblloyd, line 63: Can these two be switched back to init?

garrett-sutton: I believe they need to have setters because without them the callback can't allow for a client to specify a new username and password (i.e. we are passing in an existing instance of `ClientOpts` to the AuthenticateAsync method that invokes the new callback).

mtmk: yes as @garrett-sutton said they are being set [here](https://github.com/nats-io/nats.net/pull/712/files#diff-d1f1dfc7f712aaf1e0287005b7a0d8f8ae6592d24fcc1adb58b2a137efc8fb96R70-R72)

We should probably make `ClientOpts` a readonly record all together and return a new `ClientOpts` object from the auth method. that may be too much change for this PR we can follow up.

mtmk:
@garrett-sutton I'd like to allow a few days for others to have a chance to review and comment. All being well, I'd say we can merge and release early next week.

garrett-sutton:
@mtmk sounds good. I'm looking to utilize the new feature in a new project at my org that we're trying to release soon. I just wanted to check in on where we are at.  Thanks!

src/NATS.Client.Core/NatsAuthOpts.cs
```diff
@@ -1,5 +1,47 @@
...
+        Secret = secret;
+    }
+
+    internal NatsAuthType Type { get; }
+
+    internal string? Value { get; }
+
+    internal string? Secret { get; }
+
+    public static NatsAuthCred NoAuth() => new(NatsAuthType.None, string.Empty, string.Empty);
```
mtmk, line 29: we should remove this. to me it's not well defined what it's for and I don't believe there is a known use case for it now. I suggest we take it out to avoid confusion unless I'm missing something (sorry I may have suggested it, not sure).

src/NATS.Client.Core/NatsAuthOpts.cs
```diff
@@ -1,5 +1,47 @@
 namespace NATS.Client.Core;
 
+public enum NatsAuthType
```
mtmk, line 3: could this be made internal?

src/NATS.Client.Core/NatsAuthOpts.cs
```diff
@@ -1,5 +1,47 @@
...
+{
+    None,
```
mtmk, line 5: we should remove this too as per NoAuth comment elsewhere.

src/NATS.Client.Core/NatsAuthOpts.cs
```diff
@@ -1,5 +1,47 @@
...
+
+    internal NatsAuthType Type { get; }
+
+    internal string? Value { get; }
+
+    internal string? Secret { get; }
+
+    public static NatsAuthCred NoAuth() => new(NatsAuthType.None, string.Empty, string.Empty);
+
+    public static NatsAuthCred UserInfo(string username, string password)
```
mtmk, line 31: Can we call these methods `FromUserInfo(), FromToken(), ...`? I think that would be more inline with other APIs like `TimeSpan`.

mtmk (changes requested):
sorry @garrett-sutton a few last minute changes from me nothing too big hopefully.

src/NATS.Client.Core/NatsAuthOpts.cs
garrett-sutton: I did this in the latest iteration of the commit with functional changes: 333c03bc223ad94af6514f2c6f4a97e6af296498

src/NATS.Client.Core/NatsAuthOpts.cs
garrett-sutton: removed in latest iteration of functional commit: 333c03bc223ad94af6514f2c6f4a97e6af296498

src/NATS.Client.Core/NatsAuthOpts.cs
caleblloyd: I think the use case was you had auth before, but you don't now.  Or you have a pool of servers, some with auth and some without.  So when you see the Url that is being connected to, you can specify no auth if you would like.

I believe this could be a `public static readonly NatsAuthCred None` though if you like that better

src/NATS.Client.Core/NatsAuthOpts.cs
garrett-sutton: changed in the latest iteration of the functional commit: 333c03bc223ad94af6514f2c6f4a97e6af296498

src/NATS.Client.Core/NatsAuthOpts.cs
garrett-sutton: agreed. I removed this in the latest iteration of the functional commit: 333c03bc223ad94af6514f2c6f4a97e6af296498

src/NATS.Client.Core/NatsAuthOpts.cs
caleblloyd: Not sure if removing this is a good idea, since `NatsAuthCred` is a struct if an empty struct is declared, it's going to have the default value of this enum.  The `None` value makes sense there.

src/NATS.Client.Core/NatsAuthOpts.cs
garrett-sutton: @caleblloyd Do you have a sense for how likely that workflow could be? I'm torn here that yes, this is technically possible but it seems much less likely than the other workflows. We're also not taking away someone's ability to connect w/o auth. 

Curious to hear your thoughts.

Also, impeccable timing on your comment and mine.

caleblloyd: I don't think switching auth methods is all that common, but I do like how this Struct represents the Union of all available Auth types.  I could see us using it in the future to construct a `NatsAuthOpts`, so it'd be nice to have the `NoAuth` option.  Also since it's a Struct that does represent an accurate default for an empty struct that is declared.  (well it did when the `NatsAuthCred` enum had `None` as the default :smile:

mtmk: good point about `None` but currently there is no legit way of creating this struct and the enum is internal now. I'm not sure about NoAuth behavior though. Would it make the auth options to fallback to initial values, make the connection anonymous or use the options from last set values? also it feels like a super edge case to me, idk. I think we need to put more thought into that. Would you be happy to leave it to a follow up PR?

src/NATS.Client.Core/NatsAuthOpts.cs
mtmk: ah I see one can say `default`. thanks for catching this. @garrett-sutton could you bring back the `None` if you still have some spare time for us please? sorry about messing you around on this.

src/NATS.Client.Core/NatsAuthOpts.cs
caleblloyd: > Would it make the auth options to fallback to initial values, make the connection anonymous or use the options from last set values?

I had not thought about that… I think I was assuming it would make the connection anonymous.

> Would you be happy to leave it to a follow up PR?

that is a good idea. Restore the None enum, but don’t make a specific static type here.

But we still have to deal with the case that an empty struct can be returner from the callback. What should we do there? Make the connection anonymous? Throw?

mtmk: currently (when the None is back) it's going to keep the settings from the last call if I ran the code correctly in my head. Could we get away with stating the behavior is undefined at this stage?

garrett-sutton: I think we're in agreement here that after bringing back `None` we can specify in the code that the `None` behavior from the callback is not defined. That behavior is specified [here](https://github.com/nats-io/nats.net/pull/712/files#diff-d1f1dfc7f712aaf1e0287005b7a0d8f8ae6592d24fcc1adb58b2a137efc8fb96R69)

src/NATS.Client.Core/NatsAuthOpts.cs
garrett-sutton: This makes sense. Thanks for the insight on this.

None is now specified in the most recent version of the functional commit: e1ac12d9b0f67cdafbe9aa0052c683d9f69d063e

src/NATS.Client.Core/NatsAuthOpts.cs
```diff
@@ -20,11 +60,21 @@ public record NatsAuthOpts
 
     public string? NKeyFile { get; init; }
 
+    /// <summary>
+    /// Callback to provide NATS authentication credentials.
+    /// When specified, the <see cref="NatsAuthType"/> of the
+    /// returned <see cref="NatsAuthCred"/> will take precedence
+    /// over other authentication options.
+    /// </summary>
+    public Func<Uri, CancellationToken, ValueTask<NatsAuthCred>>? AuthCredCallback { get; init; }
```
mtmk, line 69: Needs rewording. `NatsAuthType` is internal now we shouldn't mention it in public docs. I like the bit about precedence. Can we also add a remark that retuning `default(NatsAuthCred)` is undefined?

mtmk: it's fine actually I can fix that quick and we're ready to merge.

caleblloyd (approved):

mtmk (approved):
LGTM thanks @garrett-sutton

PR #719, opened 2025-01-17
Fix telemetry header readonly error

### LINKED ISSUE #713: Update on ReadOnly Headers after Retry
#### Some context on how we found the issue:
We noticed the [exception](https://github.com/nats-io/nats.net/blob/6bf76e222457e958d04a29f72096f22c963b51a0/src/NATS.Client.Core/NatsHeaders.cs#L430) "The response headers cannot be modified because the response has already started." in our logs due to misuse of the library on our part when processing our custom retries of jetstream publish requests. It looked something like this: 

Try 1: Some random exception we got 
Try 2: "The response headers cannot be modified because the response has already started."
Try 3: "The response headers cannot be modified because the response has already started."

This was because we were using the same `NatsHeader` object during each retry and the NATS client [sets](https://github.com/nats-io/nats.net/blob/b27eb7cc3459040c494e2d8910e4040cc9c51cda/src/NATS.Client.Core/NatsConnection.Publish.cs#L44) those headers as `ReadOnly` when a request is being traced. 

However, after fixing that, we noticed some strange behavior. Now after fixing that, the exceptions upon each retry look like this: 

Try 1: "The response headers cannot be modified because the response has already started."
Try 2: "The response headers cannot be modified because the response has already started."
Try 3: "The response headers cannot be modified because the response has already started."

We are getting that same exception but now even when we are not retrying (or are we). Upon taking a closer look at the NATS internals, we realized that there is an internal retry of requests [here](https://github.com/nats-io/nats.net/blob/6bf76e222457e958d04a29f72096f22c963b51a0/src/NATS.Client.JetStream/NatsJSContext.cs#L122) in the PublishAsync of the NATSJSContext. 

Stack Trace:
```
The response headers cannot be modified because the response has already started.
at NATS.Client.Core.Internal.Telemetry.AddTraceContextHeaders(Activity activity, NatsHeaders& headers)
   at NATS.Client.Core.NatsConnection.PublishAsync[T](String subject, T data, NatsHeaders headers, String replyTo, INatsSerialize`1 serializer, NatsPubOpts opts, CancellationToken cancellationToken)
   at NATS.Client.Core.NatsConnection.CreateRequestSubAsync[TRequest,TReply](String subject, TRequest data, NatsHeaders headers, INatsSerialize`1 requestSerializer, INatsDeserialize`1 replySerializer, NatsPubOpts requestOpts, NatsSubOpts replyOpts, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.PublishAsync[T](String subject, T data, INatsSerialize`1 serializer, NatsJSPubOpts opts, NatsHeaders headers, CancellationToken cancellationToken)
```

What seems to be happening is there is an exception with the publish request, but during the second retry, there is an internal NATS exception because of the reuse of headers during each retry.

### Expected behavior

The expected behavior would be for there to be no attempts at modifying `NatsHeaders` after they have been set to `ReadOnly`

### Server and client version

NATS.NET Client Version: 2.5.5

### Host environment

_No response_

### Steps to reproduce

1. Publish request with jetstream (ensure [Telemetry.HasListeners()](https://github.com/nats-io/nats.net/blob/b27eb7cc3459040c494e2d8910e4040cc9c51cda/src/NATS.Client.Core/NatsConnection.Publish.cs#L37) returns true)
2. Make sure there is an exception with the request
3. Retry the request once

mtmk:
* fixes #713


src/NATS.Client.Core/NatsConnection.Publish.cs
```diff
@@ -11,6 +11,7 @@ public ValueTask PublishAsync(string subject, NatsHeaders? headers = default, st
         if (Telemetry.HasListeners())
         {
             using var activity = Telemetry.StartSendActivity($"{SpanDestinationName(subject)} {Telemetry.Constants.PublishActivityName}", this, subject, replyTo);
+            Telemetry.AddTraceContextHeaders(activity, ref headers);
```
mtmk, line 14: this was a bug fix

scottf (approved):
LGTM

PR #720, opened 2025-01-22
Update sandbox apps to net8.0

### LINKED ISSUE #717: Upgrade Sandbox and Example console apps to net8.0
### Proposed change

Upgrade the Sandbox and Example console apps to net8.0

### Use case

Developers looking for examples should see them using the most recent LTS framework

### Contribution

Yes, I can do it!

caleblloyd:
Resolves #717 

- Update sample apps to net8.0
- Also update dependencies to all projects except for those in `src/`
  - Version pin FluentAssertions to v7, because v8 is [no longer FOSS](https://github.com/fluentassertions/fluentassertions/pull/2943)
  
Note: I did not run each of these individually to test that they don't have runtime errors.  I did run BlazorWasm and fix some runtime errors in that.

### DIFF
--- a/sandbox/BlazorWasm/Client/BlazorWasm.Client.csproj
+++ b/sandbox/BlazorWasm/Client/BlazorWasm.Client.csproj
@@ -9,7 +9,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="8.0.12" />
-    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="9.0.1" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="8.0.12" PrivateAssets="all" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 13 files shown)


tests/NATS.Extensions.Microsoft.DependencyInjection.Tests/NATS.Extensions.Microsoft.DependencyInjection.Tests.csproj
```diff
@@ -9,14 +9,14 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />
-        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.6.0" />
+        <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="9.0.1" />
```
mtmk, line 12: I'm seeing a warning here. should we drop to 8.0.1 instead?
```
Warning  : Microsoft.Extensions.DependencyInjection.Abstractions 9.0.1 doesn't support net6.0
```

mtmk (commented):
thanks @caleblloyd just seeing three build warnings one commented below, other two [here](https://github.com/nats-io/nats.net/pull/720/files#file-tools-schema-generation-program-cs-L86) and [here ](https://github.com/nats-io/nats.net/pull/720/files#file-sandbox-microbenchmark-defaultbench-cs-L94) nullable warnings - not a code change but probably because target was updated

caleblloyd:
For good measure I went ahead and changed the MSFT packages from 9.x to 8.x

mtmk (approved):
LGTM thanks @caleblloyd

PR #722, opened 2025-01-23
Use more threads for xUnit tests

caleblloyd:
~~Attempt at speeding up test runs by using the `aggressive` ParallelAlgorithm for test assemblies using xUnit~~

Speed up xunit tests by using `MaxParallelThreads = 4.0x`

Also make test NatsServer start and create connection methods async, so these don't block the thread pool during tests

### DIFF
--- a/tests/NATS.Client.TestUtilities/NatsServer.cs
+++ b/tests/NATS.Client.TestUtilities/NatsServer.cs
@@ -201,7 +201,9 @@ public class NatsServer : IAsyncDisposable
         {
             try
             {
-                await client.ConnectAsync("127.0.0.1", Opts.ServerPort, loopCts.Token);
+                var attemptTimeoutCts = new CancellationTokenSource(250);
+                var attemptCts = CancellationTokenSource.CreateLinkedTokenSource(loopCts.Token, attemptTimeoutCts.Token);
+                await client.ConnectAsync("127.0.0.1", Opts.ServerPort, attemptCts.Token);
                 if (client.Connected)
                     break;
             }
@@ -210,7 +212,7 @@ public class NatsServer : IAsyncDisposable
                 // ignore
             }
 
-            await Task.Delay(500, loopCts.Token);
+            await Task.Delay(250, loopCts.Token);
         }
 
         if (_processOut.IsFaulted)


caleblloyd:
This is ready, just the usual flappers failing in CI

Slight improvement, Linux runs down from 12m to 10m

If xUnit v3 implements parallel test runs within a class (https://github.com/xunit/xunit/issues/1986) then we can probably enable that and get even faster, I have a feeling we are limited by the longest running test class right now

tests/NATS.Client.TestUtilities/Utils.cs
```diff
@@ -140,7 +140,7 @@ await Retry.Until("service is found", async () =>
             var count = 0;
 
             // NATS cli sends an empty JSON object '{}' as the request payload, so we do the same here
-            await foreach (var msg in nats.RequestManyAsync<string, T>(subject, "{}", replySerializer: serializer, replyOpts: replyOpts, cancellationToken: ct).ConfigureAwait(false))
```
caleblloyd, line 143: I removed this to standardized tests not calling `ConfigureAwait` - was it needed for some particular reason?

tests/NATS.Client.TestUtilities/Utils.cs
```diff
@@ -150,7 +150,7 @@ await Retry.Until("service is found", async () =>
         });
 
         var count = 0;
-        await foreach (var msg in nats.RequestManyAsync<string, T>(subject, "{}", replySerializer: serializer, replyOpts: replyOpts, cancellationToken: ct).ConfigureAwait(false))
```
caleblloyd, line 153: here too

tests/NATS.Client.Simplified.Tests/ClientTest.cs
```diff
@@ -127,7 +127,7 @@ public async Task Client_works_with_all_expected_types_and_falls_back_to_JSON()
                         continue;
                     }
 
-                    await msg.ReplyAsync(new MyData(id, $"foo{id}"), cancellationToken: ctsTestTimeout.Token).ConfigureAwait(false);
```
caleblloyd, line 130: and here too... these were the only 3 I found

tests/NATS.Client.TestUtilities/Utils.cs
mtmk: I don't think so. it's fine.

mtmk (approved):
LGTM thanks @caleblloyd 🚀

PR #726, opened 2025-01-25
Fix JetStream test flappers

mtmk:
* Added pre connect retry loop for clients to be ready
* moved a few test to use a single nats-server rather than spawning a new one in each test (look for `[Collection("nats-server")]` on test classes)
* fixed test proxy sync issues
* extended timeouts to 30s (this one probably not that helpful but my assumption is if they are running on really small VMs with a single CPU than the extra wait might help)


piotrpio (approved):
LGTM!

PR #728, opened 2025-01-27
Added Additional Try Commands to KVStore

### LINKED ISSUE #441: Add versions of KV commands that don't throw exceptions?
### Proposed change

Add versions of KV commands that don't throw exceptions.   Throwing exceptions in all cases when you're just interested in the outcome of an operation seems a bit overkill?

### Use case

Some good uses for KV stores including leader election where we regularly try and get a lock on a resource.  So a TryCreateAsync call where we get a bool might be better than calling CreateAsync and getting exceptions thrown constantly from the API.  We may call this method every 1-3 seconds...

Likewise, just being interested in whether a kv pair exists or not might not need exceptions thrown all the time?

### Contribution

Happy to look at this if can agree on the signatures...

darkwatchuk:
Added the following commands to KVStore ;

- TryCreateAsync
- TryUpdateAsync
- TryDeleteAsync
- TryPurgeAsync

* resolves #441

### DIFF
--- a/src/NATS.Client.Core/NatsResult.cs
+++ b/src/NATS.Client.Core/NatsResult.cs
@@ -4,6 +4,7 @@ namespace NATS.Client.Core;
 
 public readonly struct NatsResult
 {
+    private static readonly NatsResult _defaultSuccess = new();
     private readonly Exception? _error;
 
     public NatsResult()
@@ -15,6 +16,8 @@ public readonly struct NatsResult
     {
         _error = error;
     }
+
+    public static NatsResult DefaultSuccess => _defaultSuccess;
 
     public Exception Error => _error ?? ThrowErrorIsNotSetException();
 
--- a/src/NATS.Client.Core/NatsResult.cs
+++ b/src/NATS.Client.Core/NatsResult.cs
@@ -2,6 +2,32 @@ using System.Runtime.CompilerServices;
 
 namespace NATS.Client.Core;
 
+public readonly struct NatsResult
+{
+    private readonly Exception? _error;
+
+    public NatsResult()
+    {
+        _error = null;
+    }
+
+    public NatsResult(Exception error)
+    {
+        _error = error;
+    }
+
+    public Exception Error => _error ?? ThrowErrorIsNotSetException();
+
+    public bool Success => _error == null;
+
+    public static implicit operator NatsResult(Exception error) => new(error);
+
+    private static Exception ThrowErrorIsNotSetException() => throw CreateInvalidOperationException("Result error is not set");
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static Exception CreateInvalidOperationException(string message) => new InvalidOperationException(message);
+}
+
 public readonly struct NatsResult<T>
 {
     private readonly T? _value;
--- a/src/NATS.Client.Core/NatsResult.cs
+++ b/src/NATS.Client.Core/NatsResult.cs
@@ -4,7 +4,6 @@ namespace NATS.Client.Core;
 
 public readonly struct NatsResult
 {
-    private static readonly NatsResult _defaultSuccess = new();
     private readonly Exception? _error;
 
     public NatsResult()
@@ -17,7 +16,7 @@ public readonly struct NatsResult
         _error = error;
     }
 
-    public static NatsResult DefaultSuccess => _defaultSuccess;
+    public static NatsResult Default => new();
 
     public Exception Error => _error ?? ThrowErrorIsNotSetException();
 
--- a/src/NATS.Client.JetStream/NatsJSContext.cs
+++ b/src/NATS.Client.JetStream/NatsJSContext.cs
@@ -82,6 +82,53 @@ public partial class NatsJSContext
         NatsHeaders? headers = default,
         CancellationToken cancellationToken = default)
     {
+        var result = await TryPublishAsync(subject, data, serializer, opts, headers, cancellationToken);
+        if (!result.Success)
+        {
+            throw result.Error;
+        }
+
+        return result.Value;
+    }
+
+    /// <summary>
+    /// Sends data to a stream associated with the subject.
+    /// </summary>
+    /// <param name="subject">Subject to publish the data to.</param>
+    /// <param name="data">Data to publish.</param>
+    /// <param name="serializer">Serializer to use for the message type.</param>
+    /// <param name="opts">Publish options.</param>
+    /// <param name="headers">Optional message headers.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the publishing call or the wait for response.</param>
+    /// <typeparam name="T">Type of the data being sent.</typeparam>
+    /// <returns>
+    /// The ACK response to indicate if stream accepted the message as well as additional
+    /// information like the sequence number of the message stored by the stream.
+    /// </returns>
+    /// <exception cref="NatsJSException">There was a problem receiving the response.</exception>
+    /// <remarks>
+    /// <para>
+    /// Use this method to avoid exceptions.
+    /// </para>
+    /// <para>
+    /// Note that if the subject isn't backed by a stream or the connected NATS server
+    /// isn't running with JetStream enabled, this call will hang waiting for an ACK
+    /// until the request times out.
+    /// </para>
+    /// <para>
+    /// By setting <c>msgId</c> you can ensure messages written to a stream only once. JetStream support idempotent
+    /// message writes by ignoring duplicate messages as indicated by the Nats-Msg-Id header. If both <c>msgId</c>
+    /// and the <c>Nats-Msg-Id</c> header value was set, <c>msgId</c> parameter value will be used.
+    /// </para>
+    /// </remarks>
+    public async ValueTask<NatsResult<PubAckResponse>> TryPublishAsync<T>(
+        string subject,
+        T? data,
+        INatsSerialize<T>? serializer = default,
+        NatsJSPubOpts? opts = default,
+        NatsHeaders? headers = default,
+        CancellationToken cancellationToken = default)
+    {
         if (opts != null)
         {
             if (opts.MsgId != null)
@@ -150,7 +197,7 @@ public partial class NatsJSContext
                 {
                     if (msg.Data == null)
                     {
-                        throw new NatsJSException("No response data received");
+                        return new NatsJSException("No response data received");
                     }
 
                     return msg.Data;
@@ -169,7 +216,7 @@ public partial class NatsJSContext
 
         // We throw a specific exception here for convenience so that the caller doesn't
         // have to check for the exception message etc.
-        throw new NatsJSPublishNoResponseException();
+        return new NatsJSPublishNoResponseException();
     }
 
     public async ValueTask<NatsJSPublishConcurrentFuture> PublishConcurrentAsync<T>(
--- a/src/NATS.Client.JetStream/NatsJSContext.cs
+++ b/src/NATS.Client.JetStream/NatsJSContext.cs
@@ -189,7 +189,14 @@ public partial class NatsJSContext
 
             await foreach (var msg in sub.Msgs.ReadAllAsync(cancellationToken).ConfigureAwait(false))
             {
-                if (msg.Data == null)
+                // If JetStream is disabled, a no responders error will be returned.
+                // No responders error might also happen when reconnecting to cluster.
+                // We should retry in those cases.
+                if (msg.HasNoResponders)
+                {
+                    break;
+                }
+                else if (msg.Data == null)
                 {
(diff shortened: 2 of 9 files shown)


tests/NATS.Client.KeyValueStore.Tests/KeyValueStoreTest.cs
```diff
@@ -695,4 +695,73 @@ await Retry.Until(
...
+
+        var deleteResult = await store.TryDeleteAsync("k1");
+        Assert.True(deleteResult.Success);
+
+        result = await store.TryCreateAsync("k1", "v3");
+        Assert.True(result.Success);
+
+        var finalValue = await store.TryGetEntryAsync<string>("k1");
+        Assert.True(finalValue.Success && finalValue.Value.Value == "v3");
+    }
```
mtmk, line 726: should we have couple of error cases as well in these tests?

darkwatchuk: There are, e.g. invalid revision. Besides, all the other kv tests call these indirectly. You want more added?

mtmk: ah yes I see checking if success is false i was looking for `result.Error` asserts. do you think if we check Errors as well that could help?

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -105,32 +105,96 @@ public async ValueTask<ulong> CreateAsync<T>(string key, T value, INatsSerialize
         throw new NatsKVCreateException();
     }
 
+    /// <inheritdoc />
+    public async ValueTask<NatsResult<ulong>> TryCreateAsync<T>(string key, T value, INatsSerialize<T>? serializer = default, CancellationToken cancellationToken = default)
+    {
+        ValidateKey(key);
```
mtmk, line 111: validate key throws exceptions as well.

darkwatchuk: Yes, it does.  We could trap that too, but is that wise?  I'd say that's more of a dev issue really, whereas the idea behind the Try command was to attempt something on the server, and if that failed, that could be ok.   I can go either way on this one. Up to you guys...

mtmk: What would be devs expectation? but you're right it's more about dealing with 'expected' scenarios. We can defer this for later if needed.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -105,32 +105,96 @@ public async ValueTask<ulong> CreateAsync<T>(string key, T value, INatsSerialize
...
+
+    /// <inheritdoc />
+    public async ValueTask<NatsResult<ulong>> TryUpdateAsync<T>(string key, T value, ulong revision, INatsSerialize<T>? serializer = default, CancellationToken cancellationToken = default)
     {
         ValidateKey(key);
         var headers = new NatsHeaders { { NatsExpectedLastSubjectSequence, revision.ToString() } };
 
         try
         {
             var ack = await JetStreamContext.PublishAsync($"$KV.{Bucket}.{key}", value, headers: headers, serializer: serializer, cancellationToken: cancellationToken);
```
mtmk, line 159: We should create js.TryPublish as well in a follow up PR.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
darkwatchuk: Whilst developing, you're more likely to come up against the invalid key based on the choices you've made when using the KV. During runtime you'd probably never get these.  Am not sure if something like ValidateKey is publicly available?  That might be a nice option if it isn't at some point.  IsValidKey... etc..

As I said, happy to go either way on this one, but the Try is more - "try this operation against the server and don't crash if the key's not in the state we might  think it's in"

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -105,32 +105,96 @@ public async ValueTask<ulong> CreateAsync<T>(string key, T value, INatsSerialize
...
-            ack.EnsureSuccess();
 
-            return ack.Seq;
-        }
-        catch (NatsJSApiException e)
-        {
-            if (e.Error is { ErrCode: 10071, Code: 400, Description: not null } && e.Error.Description.StartsWith("wrong last sequence", StringComparison.OrdinalIgnoreCase))
+            if (ack == null)
```
mtmk, line 161: do we need this check? ack isn't nullable.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -157,23 +221,40 @@ public async ValueTask DeleteAsync(string key, NatsKVDeleteOpts? opts = default,
         try
         {
             var ack = await JetStreamContext.PublishAsync<object?>(subject, null, headers: headers, cancellationToken: cancellationToken);
-            ack.EnsureSuccess();
-        }
-        catch (NatsJSApiException e)
-        {
-            if (e.Error is { ErrCode: 10071, Code: 400, Description: not null } && e.Error.Description.StartsWith("wrong last sequence", StringComparison.OrdinalIgnoreCase))
+
+            if (ack == null)
```
mtmk, line 225: same, ack shouldn't be null.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
darkwatchuk: done

src/NATS.Client.KeyValueStore/NatsKVStore.cs
darkwatchuk: done

src/NATS.Client.KeyValueStore/NatsKVStore.cs
darkwatchuk: Have included this too.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -555,22 +612,24 @@ internal async ValueTask<NatsKVWatcher<T>> WatchInternalAsync<T>(IEnumerable<str
...
         if (!ValidKeyRegex.IsMatch(key))
         {
-            ThrowNatsKVException("Key contains invalid characters");
+            return KeyContainsInvalidCharactersException;
         }
+
+        return new NatsResult();
     }
 
     [MethodImpl(MethodImplOptions.NoInlining)]
```
mtmk, line 635: `ThrowNatsKVException` looks unused in Rider for me.

src/NATS.Client.JetStream/NatsJSContext.cs
```diff
@@ -150,7 +197,7 @@ public async ValueTask<PubAckResponse> PublishAsync<T>(
                 {
                     if (msg.Data == null)
                     {
-                        throw new NatsJSException("No response data received");
+                        return new NatsJSException("No response data received");
                     }
```
mtmk, line 202: we don't need this to be in a try-catch anymore. Also we don't need to set `ThrowIfNoResponders = true,` above since `NatsMsg` now has a property to tell us if it was a no responders message. We can check that and break out here which should give us the same function.

src/NATS.Client.Core/NatsResult.cs
```diff
@@ -16,6 +17,8 @@ public NatsResult(Exception error)
         _error = error;
     }
 
+    public static NatsResult DefaultSuccess => _defaultSuccess;
```
mtmk, line 20: Nice touch! I think we can just say `default` here no need to have a backing field since it's a struct. (looking at e.g. `ValueTask.CompletedTask`)

darkwatchuk: Ok, would also help I used this too....!

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -69,19 +72,62 @@ internal NatsKVStore(string bucket, INatsJSContext context, INatsJSStream stream
     /// <inheritdoc />
     public string Bucket { get; }
 
+    /// <inheritdoc />
```
mtmk, line 75: This needs its own docs. nothing to inherit the docs from.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -69,19 +72,62 @@ internal NatsKVStore(string bucket, INatsJSContext context, INatsJSStream stream
...
+        if (!keyValidResult.Success)
+        {
+            return keyValidResult.Error;
+        }
+
+        var publishResult = await JetStreamContext.TryPublishAsync($"$KV.{Bucket}.{key}", value, serializer: serializer, cancellationToken: cancellationToken);
+        if (publishResult.Success)
+        {
+            var ack = publishResult.Value;
+            if (ack.Error is { ErrCode: 10071, Code: 400, Description: not null } && ack.Error.Description.StartsWith("wrong last sequence", StringComparison.OrdinalIgnoreCase))
```
mtmk, line 102: No need to check this since we're not passing in revision.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -69,19 +72,62 @@ internal NatsKVStore(string bucket, INatsJSContext context, INatsJSStream stream
...
+        {
+            return publishResult.Error;
+        }
     }
 
     /// <inheritdoc />
     public async ValueTask<ulong> CreateAsync<T>(string key, T value, INatsSerialize<T>? serializer = default, CancellationToken cancellationToken = default)
     {
-        ValidateKey(key);
+        var keyValidResult = TryValidateKey(key);
```
mtmk, line 126: I missed this before I think. was there any reason why we didn't delegate this one to `TryCreate` like the others?

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -151,18 +201,19 @@ public async ValueTask<ulong> UpdateAsync<T>(string key, T value, ulong revision
...
+        if (!keyValidResult.Success)
+        {
+            return keyValidResult.Error;
+        }
+
         var headers = new NatsHeaders { { NatsExpectedLastSubjectSequence, revision.ToString() } };
 
-        try
+        var publishResult = await JetStreamContext.TryPublishAsync($"$KV.{Bucket}.{key}", value, headers: headers, serializer: serializer, cancellationToken: cancellationToken);
```
mtmk, line 212: btw we can now use `_kvBucket + key` (like we did in TryGet) for the key. turned out a simple string concat was the most performant.

src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -218,15 +274,11 @@ public async ValueTask<NatsResult> TryDeleteAsync(string key, NatsKVDeleteOpts?
 
         var subject = $"$KV.{Bucket}.{key}";
```
mtmk, line 275: same as above

src/NATS.Client.JetStream/NatsJSContext.cs
```diff
@@ -183,28 +183,18 @@ public async ValueTask<NatsResult<PubAckResponse>> TryPublishAsync<T>(
...
-                        ThrowIfNoResponders = true,
                     },
                     cancellationToken)
                 .ConfigureAwait(false);
 
-            try
+            await foreach (var msg in sub.Msgs.ReadAllAsync(cancellationToken).ConfigureAwait(false))
             {
-                await foreach (var msg in sub.Msgs.ReadAllAsync(cancellationToken).ConfigureAwait(false))
+                if (msg.Data == null)
```
mtmk, line 192: We still need to check for no-responders here otherwise we won't retry (actually it wasn't as bad, I thought it was going to hang). Please add this before msg.Data == null check:
```csharp
                // If JetStream is disabled, a no responders error will be returned.
                // No responders error might also happen when reconnecting to cluster.
                // We should retry in those cases.
                if (msg.HasNoResponders)
                {
                    break;
                }
```

Also it turns out we don't have a test for this. Could you also add something like this test to PublishTest.cs please?
```csharp
    [Fact]
    public async Task Publish_no_responders()
    {
        await using var server = await NatsServer.StartAsync();
        await using var connection = await server.CreateClientConnectionAsync();
        var js = connection.CreateJetStreamContext();
        var result = await js.TryPublishAsync("foo", 1);
        Assert.IsType<NatsJSPublishNoResponseException>(result.Error);
    }
```

mtmk (approved):
LGTM thanks @darkwatchuk

PR #745, opened 2025-02-18
Add server error event

### LINKED ISSUE #496: Server errors aren't throwing exceptions
### Observed behavior

When server responds with an error `-ERR` we're not propagating the error as an exception to relevant calls. One use case is the 'Permissions Violation' error for example when permissions are set on the server per subject.

### Expected behavior

Relevant calls should throw an exception. This should also propagate through to the higher level APIs like Services.

workaround: You should observe the errors in logs

### Server and client version

any

### Host environment

any

### Steps to reproduce

Setup server with JWT auth and permission sub/pub.
```csharp
var opts = new NatsOpts
{
    LoggerFactory = // register your factory here
    Url = "tls://connect.ngs.global:4222",
    Name = "test-service",
    AuthOpts = new NatsAuthOpts { Jwt = "xyz..." }
};
var cnn = new NatsConnection(opts);
var svc = new NatsSvcContext(cnn);
var cfg = new NatsSvcConfig("foo", "1.0.0") { Description = "test service" };
var service = await svc.AddServiceAsync(cfg);
await service.AddEndpointAsync<string>((x) => { return x.ReplyAsync("test"); }, "getstring", "foo.aaa.bbb.test");
```
Code above runs without any exceptions and it actually doesn't work. you can observe the errors in log.

Also if the `$SVC` subject is allowed but the endpoint subject isn't, service and the endpoint is registered but the endpoint does not respond.

Thanks @mkobaly for finding the issue and digging into it.

* see also https://github.com/nats-io/nats-architecture-and-design/pull/282

mtmk:
Raises a `ServerError` event on `INatsConnection` whenever the server sends a `-ERR` line, so applications can observe issues like permissions violations without scraping logs.

Refs #496


mtmk:
@claude review

scottf (approved):
LGTM

PR #751, opened 2025-02-27
Add NATS v2.11 requirement notes for TTL-related features

mtmk:
Updated XML comments and remarks to specify that TTL and delete marker features require NATS server v2.11 or later. This ensures clarity on version-specific functionality for developers.


scottf (approved):
LGTM

PR #756, opened 2025-03-03
Added an ability to specify pool size with a callback using IServiceProvider

uhfath:
When initializing NATS using `IServiceCollection.AddNatsClient` one has to specify a pool size in a callback, which doesn't support passing `IServiceProvider`. This involves setting pool size upfront as a constant or reading it from `IConfiguration` directly with something like `IConfiguration.GetSection(...).Get<...>()`.
This PR retains an ability to specify pool size constantly as it is now without breaking changes. And also adds an ability to use `IServiceProvider` which helps in cases where pool size is defined in one of app options. Something like:
```
.WithPoolSize(sp => sp.GetService<IOptions<AppOption>>().Value.NATSPoolSize)
```
But because of this there will be no more `SingleConnectionFactory` and `PooledConnectionFactory` will be always used even if pool size is 1.
This might come with a performance impact in very high load cases since there is another layer of delegates added.
But it should be negligible.

### DIFF
--- a/src/NATS.Extensions.Microsoft.DependencyInjection/NatsBuilder.cs
+++ b/src/NATS.Extensions.Microsoft.DependencyInjection/NatsBuilder.cs
@@ -10,7 +10,7 @@ public class NatsBuilder
 {
     private readonly IServiceCollection _services;
 
-    private int _poolSize = 1;
+    private Func<IServiceProvider, int> _poolSizeConfigurer = _ => 1;
     private Func<IServiceProvider, NatsOpts, NatsOpts>? _configureOpts;
     private Action<IServiceProvider, NatsConnection>? _configureConnection;
     private object? _diKey = null;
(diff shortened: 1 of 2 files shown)


mtmk:
admin: please sign your commit it's required by CNCF. also I don't know why it's showing conflicts already with NatsBuilder could you check that too please?

uhfath:
> admin: please sign your commit it's required by CNCF. also I don't know why it's showing conflicts already with NatsBuilder could you check that too please?

Yeah, I've noticed that too.
Not sure what happened.
I have added a signature to my commits pulled `main` branch before making a PR.
I'll investigate.

mtmk:
> > admin: please sign your commit it's required by CNCF. also I don't know why it's showing conflicts already with NatsBuilder could you check that too please?
> 
> Yeah, I've noticed that too. Not sure what happened. I have added a signature to my commits pulled `main` branch before making a PR. I'll investigate.

thank you. btw with the commit signing you'll need to force push since we need all commits signed. sorry about the hassle there it's something we have to do admin wise.

uhfath:
I have created another PR (https://github.com/nats-io/nats.net/pull/757) that should be correctly signed and based on main branch. Will close this one.

PR #757, opened 2025-03-03
Added an ability to specify pool size with a callback using IServiceProvider (fixed&signed)

uhfath:
When initializing NATS using `IServiceCollection.AddNatsClient` one has to specify a pool size in a callback, which doesn't support passing `IServiceProvider`. This involves setting pool size upfront as a constant or reading it from `IConfiguration` directly with something like `IConfiguration.GetSection(...).Get<...>()`.
This PR retains an ability to specify pool size constantly as it is now without breaking changes. And also adds an ability to use `IServiceProvider` which helps in cases where pool size is defined in one of app options. Something like:
```
.WithPoolSize(sp => sp.GetService<IOptions<AppOption>>().Value.NATSPoolSize)
```
But because of this there will be no more `SingleConnectionFactory` and `PooledConnectionFactory` will be always used even if pool size is 1.
This might come with a performance impact in very high load cases since there is another layer of delegates added.
But it should be negligible.


mtmk:
please merge main. fixes build.

uhfath:
Done.

mtmk:
cc @rickdotnet

mtmk (approved):
LGTM thanks @uhfath

PR #761, opened 2025-03-05
Add NatsPayloadTooLargeException class

### LINKED ISSUE #759: Provide NatsPayloadTooLargeException to handle payload size related situation
### Proposed change

We propose to create a `NatsPayloadTooLargeException` class derived from `NatsException` class.

### Use case

In our utilization of nats .net client (currently, at 2.5.9), we want to handle the situation where payload size exceeds the server limitation separately from the other NatsExceptions.

At the moment, the only way to do that is check if the exception message contains "exceeds server's maximum payload size" string, which is not good.

With the new Exception, we can handle the situation with confidence.

### Contribution

_No response_

aDisplayName:
Add NatsPayloadTooLargeException class to support special handling regarding excessive payload when writing to server.

Fix #759


mtmk:
thanks @aDisplayName although it's a simple one should we add a test?

aDisplayName:
Related test has been updated

mtmk:
please merge main. fixes build.

aDisplayName:
@mtmk , Changes in main has been merged.

mtmk:
thanks @aDisplayName. no worries failing test is probably a flapper. there was a problem with the windows tests i wanted that fixed. I need to double check the merge but other than that it should be all good.

PR #768, opened 2025-03-08
Make the parameter-less StreamConfig constructor public

darkwatchuk:
Simple change to make the StreamConfig class available for external (de)serialization, e.g. for advisory messages such as  StreamSnapshotResponse.


mtmk (approved):
LGTM

PR #771, opened 2025-03-11
Add support for Direct Get API with key in subject

### LINKED ISSUE #770: Subject in GetDirectAsync differs across client library implementations
### Observed behavior

When a client using the nats.net library calls `GetEntryAsync`, they can encounter this error:
`Server error Permissions Violation for Publish to "$JS.API.DIRECT.GET.KV_bucket_name"`. 

This indicates that the client is attempting to publish to a subject that it does not have permissions to publish to. 

I also have a golang app that makes the same `Get` call from a key value store and it does not encounter this error.

This is seemingly caused by utilizing the auth_callout functionality of the nats server where we assign subject level permissions for NATS clients. Specifically, for KV GET interactions, clients are granted the following: `$JS.API.DIRECT.GET.*.$KV.*.subjectPrefix.>`

In looking at [nats.net](https://github.com/nats-io/nats.net/blob/a6e11459626cd800e5ba9b78f52a33275c7e965e/src/NATS.Client.JetStream/NatsJSStream.cs#L177), it looks like the .NET client only constructs the request and sends a message to `$JS.API.DIRECT.GET.KV_bucket_name` with content `{"last_by_subj":"$KV.bucket_name.example.key"}`. 

In looking at [nats.go](https://github.com/nats-io/nats.go/blob/da7160e5eeab6d2654f94b1232af1e8939fa59cb/jsm.go#L1250), it looks like it constructs the request and sends a message to `$JS.API.DIRECT.GET.KV_bucket_name.$KV.bucket_name.example.key` with no content.

### Expected behavior

Expect the nats.net and golang (and others) libraries to have feature parity. 

Could we update the current implementation to match the golang library functionality? Or could we provide an override for the `GetEntryAsync` method to match the golang library functionality?

### Server and client version

- nats server = 2.10
- nats.net client = 2.5.7

### Host environment

_No response_

### Steps to reproduce

https://natsio.slack.com/archives/C069GSYFP/p1741624842560849 has details on the difference in behavior.

1. Grant this NATS permissions to KV clients: `$JS.API.DIRECT.GET.KV_bucket_name.$KV.bucket_name.example.key` (you'd also need PUT permissions but those aren't problematic here)
2. with a dotnet app, `GetEntryAsync` from `example.key`
3. See error described above
4. with a go app, `Get` from `example.key`
5. See you successfully get the latest value associated with that key

mtmk:
Proposed solution for #770 
* fixes #770

### DIFF
--- a/src/NATS.Client.KeyValueStore/NatsKVContext.cs
+++ b/src/NATS.Client.KeyValueStore/NatsKVContext.cs
@@ -19,15 +19,30 @@ public enum NatsKVStorageType
 /// </summary>
 public class NatsKVContext : INatsKVContext
 {
-    private const string KvStreamNamePrefix = "KV_";
+    internal const string KvStreamNamePrefix = "KV_";
     private static readonly int KvStreamNamePrefixLen = KvStreamNamePrefix.Length;
     private static readonly Regex ValidBucketRegex = new(pattern: @"\A[a-zA-Z0-9_-]+\z", RegexOptions.Compiled);
+    private readonly NatsKVOpts _opts;
 
     /// <summary>
     /// Create a new Key Value Store context
     /// </summary>
     /// <param name="context">JetStream context</param>
-    public NatsKVContext(INatsJSContext context) => JetStreamContext = context;
+    /// <param name="opts">Context options</param>
+    public NatsKVContext(INatsJSContext context, NatsKVOpts opts)
+    {
+        JetStreamContext = context;
+        _opts = opts;
+    }
+
+    /// <summary>
+    /// Create a new Key Value Store context
+    /// </summary>
+    /// <param name="context">JetStream context</param>
+    public NatsKVContext(INatsJSContext context)
+        : this(context, NatsKVOpts.Default)
+    {
+    }
 
     /// <inheritdoc />
     public INatsJSContext JetStreamContext { get; }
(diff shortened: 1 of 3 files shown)


garrett-sutton:
@mtmk This is great. Thanks for hopping on it so quickly. This approach will work for our use case.

garrett-sutton:
**potential for future**

One additional piece that I had to think about for a little bit. In my use case's current state, we rely on the extensions that are defined in `src/NATS.Client.KeyValueStore/NatsClientExtensions.cs` to create an `INatsKVContext `. With the enhancements that are proposed as part of this PR we'll need to transition to setting up the `INatsKVContext` directly (rather than via the extension off of the `INatsConnection`). 

Like I said in my other comment, I think the current proposal is good and we are OK with setting up the INatsKVConext directly though. This is largely just food for thought.

mtmk:
> One additional piece that I had to think about for a little bit. In my use case's current state, we rely on the extensions that are defined in `src/NATS.Client.KeyValueStore/NatsClientExtensions.cs` to create an `INatsKVContext `. With the enhancements that are proposed as part of this PR we'll need to transition to setting up the `INatsKVContext` directly (rather than via the extension off of the `INatsConnection`).

@garrett-sutton that is a good point. I think we have a similar case for JetStream as well off the top of my head. I'll put that in a separate PR.

scottf (approved):
LGTM

Issue #422, opened 2024-03-01
JetStream Backoff


alberk8:
### Observed behavior

With  code below it looks like the backoff is not in effect.  Also there is no reference anywhere in the [source](https://github.com/nats-io/nats.net.v2/blob/56f831f8e59091aa91be6169aaad835368905c37/src/NATS.Client.JetStream/Models/ConsumerConfig.cs#L227).

### Expected behavior

Backoff Setting works

### Server and client version

Server: v2.10.11
Client: NATS.Client.JetStream v2.1.2

### Host environment

Official NATS docker (latest) with JetStream enabled.

### Steps to reproduce

```
var consumer = await jsContext.CreateOrUpdateConsumerAsync("EVENTS", new ConsumerConfig(consumerName)
{
    DeliverPolicy = ConsumerConfigDeliverPolicy.All,
    FilterSubject = "data.>",
    AckPolicy = ackPolicy,
    AckWait = ackWait,
    
    MaxDeliver = 5,
    Backoff = new List<long>() { (long) TimeSpan.FromSeconds(1).TotalSeconds, (long)TimeSpan.FromSeconds(2).TotalSeconds }
});

 await foreach (NatsJSMsg<Person> msg in consumer.ConsumeAsync<Person>())
 {
     Console.WriteLine($"Subject: {msg.Subject} Size: {msg.Size}");
     var md = msg.Metadata;
     if (md.HasValue)
     {
        Console.WriteLine($"Num Delivered: {md.Value.NumDelivered} TimeStamp: {md.Value.Timestamp.ToLocalTime()}");
        Console.WriteLine($"Consumer Seq: {md.Value.Sequence.Consumer} Stream Seq: {md.Value.Sequence.Stream}");
     }
     await msg.NakAsync();
 }
```

mtmk:
it might be nanoseconds, could you try that?
We changed everything to be TimeSpan but missed this one I think 🤔

alberk8:
I have changed to MicroSeconds and there is no difference.  The Re-Delivery went through in quick succession.

mtmk:
> I have changed to MicroSeconds and there is no difference. The Re-Delivery went through in quick succession.

it's should be nano .. another x1000

> 
> Looking at the source code, it does not seems that Backoff is implemented.

I think this is just passed to the server, I don't think there is anything to implement on client side, is there?

alberk8:
[...]

PR #772, opened 2025-03-11
Use TimeSpan for backoff consumer config

yeongjonglim:
Proposed enhancement for #422.

This change introduces a breaking change that all existing definition is required to be replaced with TimeSpan type.


mtmk:
thanks @yeongjonglim sorry for the delay. looks good and i will give it a try soon. I'm planning to include in 2.6.x since technically a breaking change.

mtmk (approved):
LGTM thanks @yeongjonglim

I will try to add this to the 2.6 release.

PR #776, opened 2025-03-12
Add overloads for client extensions

### LINKED ISSUE #773: NatsClientExtensions option overloads
> One additional piece that I had to think about for a little bit. In my use case's current state, we rely on the extensions that are defined in `src/NATS.Client.KeyValueStore/NatsClientExtensions.cs` to create an `INatsKVContext `. With the enhancements that are proposed as part of this PR we'll need to transition to setting up the `INatsKVContext` directly (rather than via the extension off of the `INatsConnection`).

@garrett-sutton that is a good point. I think we have a similar case for JetStream as well off the top of my head. I'll put that in a separate PR.

_Originally posted by @mtmk in https://github.com/nats-io/nats.net/issues/771#issuecomment-2714975029_

mtmk:
+ fixes #773


scottf (approved):
LGTM

PR #798, opened 2025-03-29
NatsMsg add create method (#632)

### LINKED ISSUE #632: How to create and pass instances of new NatsMsg<T>() to the NatsConnection#PublishAsync<T>(in NatsMsg<T> msg....) API correctly?
### Proposed change

I want to use this API 

```
NatsConnection#public ValueTask PublishAsync<T>(in NatsMsg<T> msg, INatsSerialize<T>? serializer = default, NatsPubOpts? opts = default, CancellationToken cancellationToken = default) =>
        PublishAsync(msg.Subject, msg.Data, msg.Headers, msg.ReplyTo, serializer, opts, cancellationToken);
```

to publish instances of ```NatsMsg<T>```

But this requires from me to construct an instance of NatsMsg  in "low level" manner:

```
new NatsMsg<T>(
                Subject: subject,
                Data: payload,
                Headers: headers,
                ReplyTo: replyTo,
                Connection: connection,
                Size: size
            );
```
If I understood correctly, API's doc hints to provide the Size of the message in such a way:

```
 int size = subject.Length + replyTo.Length + headers.Length + payload.Length;
``` 
But to achieve this, it seems, one needs to provide the headers parsed (since headers.Length is not compilable).

_**Question: How to create and pass instances of ```NatsMsg<T>()``` to the ```NatsConnection#PublishAsync<T>(in NatsMsg<T> msg....)``` API** correctly?_

I'm asking such question since I used [nats-java](https://github.com/nats-io/nats.java) version of this API but now want to transfer the java code consistently (without loosing API usage semantic) to C#/.Net platform.
https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/impl/NatsMessage.java#L447

```java
Message message = NatsMessage.builder()
            .subject(subject)
            .headers(headers)
            .replyTo(replyTo)
            .data(payload)
            .build();     
// and then publish it:
connection.jetStream().publish(message)            
```

Note that a user of the Nats Java API is not enforced to provide "Size" of the message up front.

### Use case

Better user experience to call  ```NatsConnection#PublishAsync<T>(in NatsMsg<T> msg....)```  API.

### Contribution

_No response_

Ivandemidov00:
close #632 
* Change GetBytesLength long to int
* Add NatsMsg method Create


src/NATS.Client.Core/Internal/HeaderWriter.cs
```diff
@@ -20,7 +20,7 @@ internal class HeaderWriter
 
     private static ReadOnlySpan<byte> ColonSpace => new[] { ByteColon, ByteSpace };
 
-    internal static long GetBytesLength(NatsHeaders headers, Encoding encoding)
+    internal static int GetBytesLength(NatsHeaders headers, Encoding encoding)
```
mtmk, line 23: is there a need for this change?

src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -420,6 +421,116 @@ private void CheckReplyPreconditions()
     }
 }
 
+public static class NatsMsg
```
mtmk, line 424: the size calculation might not be accurate if for some reason a different serializer is used while sending the message. also it might mean the serialization will be done twice (in case of the `<T>` version). perhaps making the size calculation optional is a solution? wdyt?

src/NATS.Client.Core/Internal/HeaderWriter.cs
Ivandemidov00: I assumed that changing long to int in GetBytesLength is a safer change than changing the Size type in NatsMsg.
Need to calculate the size of all messages for NatsMsg

mtmk: ah no worries, just realized it's internal. totally fine. sorry.

src/NATS.Client.Core/NatsMsg.cs
Ivandemidov00: Making size optional is fine, but it would require changes to IsEmpty and HasNoResponders since they depend on it. We would need to evaluate size and flags lazily when those properties are available.
What do you think about this?

mtmk: of course you're right! can we set them to reasonable values if the calculation is opted out? e.g. if the data is null then isempty can be true. and evaluate the header to set hasnoresponders (although that is only meaningful coming from the server). btw we must set the headers to be readonly. if you're happy with that then the other question is if we make size calculation on by default or not.

Ivandemidov00: I’m thinking of adding a separate Create override that skips the size calculation, serialization, and other operations.

src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -139,7 +139,9 @@ 2               30
          Int.Max: 2,147,483,647
              8mb:     8,388,608
      */
+    private const int SizeNotSet = -1;
     private readonly uint _flagsAndSize;
+    private readonly bool _hasSize;
```
mtmk, line 144: unfortunately we can't add more fields here. the point of the flags was to avoid adding more fields to this struct to keep it's size under control. it has a nock on effect on the channels' backing arrays spilling over to LOH.

it's a readonly struct and we know everything at the time of creating. can't we just set size to `0` and set the flags accordingly? I feel we don't need to make changes to this struct but I might be missing something.

Ivandemidov00: Thanks for the feedback! Yes, I noticed later in the tests that the size is important to us. I made the processing using the maximum value (0x3FFFFFFF) for undefined sizes instead of adding additional fields.

mtmk:
@Ivandemidov00 just looking at this and the issue it's fixing from an API usability perspective and I think we need a little more .NET way of doing this. What about taking `UriBuilder` as an example and create a `NatsMsgBuilder<T>`?

```csharp
        Uri uri = new UriBuilder
        {
            Host = "localhost",
            Port = 4222,
        }.Uri;

        NatsMsg<string> msg = new NatsMsgBuilder<string>
        {
            Subject = "test",
            Data = "12345",
            ReplyTo = "reply",
            Headers = new NatsHeaders(),
        }.Msg;
```

Also can we not change the `NatsMsg<T>` and use `0` (zero) to indicate the size calculation hasn't taken palace since minimum messages size at least would have to be `1` given that we have to have a subject with at least one character. (maybe that could be a validation check for the builder as well?)

wdyt?

Ivandemidov00:
@mtmk I've added a builder to the creation of NatsMsg<T>.

mtmk (approved):
LGTM thanks @Ivandemidov00

PR #799, opened 2025-03-31
Fix reader buffer bug with typo

mtmk:


mtmk:
@Vlad-Zaytsev are you able to test this PR in your environment?

mtmk:
> > @Vlad-Zaytsev are you able to test this PR in your environment?
> 
> No. I'm sorry, but it occures only in production. (and no longer occures actualy)

no worries, thanks for confirming that. inspecting the code and the exception this is the only call that might have an issue and it's clearly wrong. since it's not happening anymore, just to confirm, the section of the code is only hit if there was a server error send back like authorization violation for example. Can you remember when the error occurred at the time there may have been such errors?

Vlad-Zaytsev:
> > > @Vlad-Zaytsev are you able to test this PR in your environment?
> > 
> > 
> > No. I'm sorry, but it occures only in production. (and no longer occures actualy)
> 
> no worries, thanks for confirming that. inspecting the code and the exception this is the only call that might have an issue and it's clearly wrong. since it's not happening anymore, just to confirm, the section of the code is only hit if there was a server error send back like authorization violation for example. Can you remember when the error occurred at the time there may have been such errors?

Maybe. I can not tell for sure.

caleblloyd (approved):
The change looks good to me!

Issue #801, opened 2025-04-02
Subscription Manager Inbox leak .NetStandard2.0


AdamsDaveAmag:
### Observed behavior

The current implementation fails to remove the subjects as the check for subscriptions can never be met, as the subscription is not removed from the subscription list of the Subject

`NATS.Client.Core:InboxSub.cs`

In `public ValueTask RemoveAsync(NatsSubBase sub)`

The line ` if (subTable.Count == 0)` can never be met as the entry is never removed from the `subTable`

### Expected behavior

The subscription to be removed from the list allowing the subject to be removed.

I believe the following was missed 
`if (subTable.RemoveAll(item => item.TryGetTarget(out var el) && el == sub) == 0)`
    `_logger.LogWarning(NatsLogEvents.InboxSubscription, "Unregistered message inbox received for {Subject}", sub.Subject);`

in the implementation before  `if (subTable.Count == 0)` matching the similar logic in the non `NETSTANDARD2_0` code

### Server and client version

Windows
Visual Studio 2022
.Net Framework Project

NATS.Net 2.5.12

### Steps to reproduce

Attempt any operation on the a KV Store `DeleteAsync` and this will cause the subjects to increase

AdamsDaveAmag:
[TestNatsInboxLeak.zip](https://github.com/user-attachments/files/19648711/TestNatsInboxLeak.zip)

This is Test Console App, Compile it up and run interactively through the debugger. The App will break into the debugger with the intention of following the comments from that point in the code.

I used Visual Studio 2022 Diagnostic Tool to snapshot the memory. The code also contains comments showing the typical output from the snapshot

ckasabula:
@mtmk Did you mean to leave the fix commented out in src/NATS.Client.Core/Internal/InboxSub.cs?

mtmk:
> [@mtmk](https://github.com/mtmk) Did you mean to leave the fix commented out in src/NATS.Client.Core/Internal/InboxSub.cs?

I'm going to uncomment it but I'm trying to have test working on the CI but proving to be unreliable. If it doesn't work (I'm about to give up) I shall remove the test and just leave your fix.

ckasabula:
> > [@mtmk](https://github.com/mtmk) Did you mean to leave the fix commented out in src/NATS.Client.Core/Internal/InboxSub.cs?
> 
[...]

PR #803, opened 2025-04-02
Inbox Subscription Memory leak .NetStandard2.0

AdamsDaveAmag:
Subjects are not cleared as the list of Subscriptions is not cleared, and therefore never removes the subject from th einbox. This causes memory consumption to increase

https://github.com/nats-io/nats.net/issues/801


ckasabula:
Any idea on what's going on w/ the failing tests? Looks like it's failing to download/run nats-server.

mtmk:
> Any idea on what's going on w/ the failing tests? Looks like it's failing to download/run nats-server.
@ckasabula checking...

mtmk:
> > Any idea on what's going on w/ the failing tests? Looks like it's failing to download/run nats-server.
> 
> @ckasabula checking...

@ckasabula thanks for the heads up. should be fixed now

mtmk:
@AdamsDaveAmag thank you for the test code you sent over [here](https://github.com/nats-io/nats.net/issues/801#issuecomment-2786232544)

Just added a unit test to confirm the leak and fix. let me know if you're happy with it then I think we should merge this.

mtmk (approved):
LGTM thanks @AdamsDaveAmag 

(Had to make a few changes to workflows to have the tests run smoothly in general. nothing to do with the change in this PR which is in InboxSub.cs only)

Issue #802, opened 2025-04-02
Communication with TLS bus behind proxy results in SslPolicyErrors.RemoteCertificateNameMismatch


wolfkor:
### Observed behavior

Our current implementation works unencrypted and TLS encrypted. With the help of #647, it also works unencrypted behind a proxy. However, when using TLS encryption behind a proxy, the certificate verification on the client side fails due to a SslPolicyErrors.RemoteCertificateNameMismatch error because the proxy hostname is used for checking.

### Expected behavior

functioning communication with TLS bus via proxy

### Server and client version

nats-server: v2.10.5
NuGet NATS.Client.Core: 2.5.12 net8.0

### Host environment

nats-server on RHEL 9

### Steps to reproduce

```
private INatsConnection GetConnection()
        {
            INatsConnection connection = null;
            int timeout = 5000;
            SocketProxy proxy = SystemUtilsService.SocketProxy;

            try
            {
                bool tlsUse = RegMan.CurrentVersion.INO_TLS_USE;
                NatsOpts opts = new()
                {
                    Url = proxy.ProxyHost != null ? $"{proxy.ProxyHost}:{proxy.ProxyPort}" : Url,
                    AuthOpts = new NatsAuthOpts
                        { Username = "xxx", Password = "xxx" },
                    ConnectTimeout    = TimeSpan.FromMilliseconds(timeout),
                    ReconnectWaitMin  = TimeSpan.FromMilliseconds(timeout),
                    MaxReconnectRetry = -1,
                    SubjectEncoding   = Encoding.UTF8,
                    TlsOpts = new NatsTlsOpts
                    {
                        Mode = tlsUse ? TlsMode.Require : TlsMode.Disable,
                        ConfigureClientAuthentication = async (sslOptions) =>
                        {
                            sslOptions.RemoteCertificateValidationCallback =
                                (sender, certificate, chain, sslPolicyErrors) =>
                                {
                                    if (sslPolicyErrors == SslPolicyErrors.None) return true;

                                    //  SslPolicyErrors.RemoteCertificateNameMismatch
                                    ErrorMessage = sslPolicyErrors.ToString();
                                    return false;
                                };
                            await Task.CompletedTask;
                        }
                    },
                };

                connection = new NatsConnection(opts);

                // Setzen der Event-Handler für Reconnect und Disconnect
                connection.Connection.ConnectionOpened +=
                    async (sender, args) => await BusConnectedEventHandlerAsync(sender, args);
[...]

PR #804, opened 2025-04-02
fix TLS proxy scenario (#802)

wolfkor:
Fix for #802


src/NATS.Client.Core/Internal/TcpConnection.cs
```diff
@@ -1,18 +1,15 @@
...
 namespace NATS.Client.Core.Internal;
 
-internal sealed class SocketClosedException : Exception
-{
-    public SocketClosedException(Exception? innerException)
-        : base("Socket has been closed.", innerException)
-    {
-    }
-}
+internal sealed class SocketClosedException(Exception? innerException) : Exception("Socket has been closed.", innerException);
```
mtmk, line 12: please rollback all the formatting changes. makes the review quite difficult. thanks.

wolfkor: hope it is easier now

mtmk: thank you

src/NATS.Client.Core/Internal/TcpConnection.cs
```diff
@@ -57,16 +59,49 @@ public ValueTask ConnectAsync(string host, int port, CancellationToken cancellat
...
         {
+            var proxyOpts = opts.ProxyOpts;
+            if (proxyOpts.Host != null)
+            {
 #if NETSTANDARD
-            await _socket.ConnectAsync(host, port).WaitAsync(timeout, cts.Token).ConfigureAwait(false);
+                await _socket.ConnectAsync(proxyOpts.Host, proxyOpts.Port).WaitAsync(opts.ConnectTimeout, cts.Token).ConfigureAwait(false);
 #else
-            await _socket.ConnectAsync(host, port, cts.Token).ConfigureAwait(false);
+                await _socket.ConnectAsync(proxyOpts.Host, proxyOpts.Port, cts.Token).ConfigureAwait(false);
```
mtmk, line 73: if this is giving the ability to swap out the host:port on top of the callback, can you also use OnConnectingAsync to take care of this swap instead?

https://github.com/nats-io/nats.net/blob/70d65febbcc2c9cb06e7b44010e9454f82ee63f1/src/NATS.Client.Core/NatsConnection.cs#L137

src/NATS.Client.Core/Internal/TcpConnection.cs
```diff
@@ -57,16 +59,49 @@ public ValueTask ConnectAsync(string host, int port, CancellationToken cancellat
...
 #endif
+
+                string? connectAuth = null;
+                if (proxyOpts.Auth != null)
+                {
+                    // Create the CONNECT request with proxy authentication
+                    var auth = Convert.ToBase64String(Encoding.UTF8.GetBytes(proxyOpts.Auth));
+                    connectAuth = $"Proxy-Authorization: Basic {auth}\r\n";
```
mtmk, line 81: other auth methods? Would there be other headers we need to think of?

src/NATS.Client.Core/Internal/TcpConnection.cs
```diff
@@ -57,16 +59,49 @@ public ValueTask ConnectAsync(string host, int port, CancellationToken cancellat
...
+                var serverWithPort = $"{host}:{port}";
+                var connectBuffer = Encoding.UTF8.GetBytes($"CONNECT {serverWithPort} HTTP/1.1\r\nHost: {serverWithPort}\r\n{connectAuth}Proxy-Connection: Keep-Alive\r\n\r\n");
+
+                // Send CONNECT request to proxy
+                await SendAsync(connectBuffer).ConfigureAwait(false);
+
+                // Validate proxy response
+                var receiveBuffer = new byte[4096];
+                var read = await ReceiveAsync(receiveBuffer).ConfigureAwait(false);
+                var response = Encoding.UTF8.GetString(receiveBuffer, 0, read);
```
mtmk, line 93: this needs to read the response until whatever. can't rely on a single read which might not return the whole response.

src/NATS.Client.Core/Internal/TcpConnection.cs
wolfkor: Isn't OnConnectingAsync intended for swaping the bus address? Why would it be needed for the proxy address as well?

src/NATS.Client.Core/Internal/TcpConnection.cs
wolfkor: I've thought about that, too. In my experience with our customers so far, I haven't encountered any other options, but there are certainly other proxy authentication options. Do you have a complete list of the methods to be supported here?

src/NATS.Client.Core/Internal/TcpConnection.cs
wolfkor: I don't really understand what you're getting at? Isn't the code almost identical to your final example for OnSocketAvailableAsync from #647?

src/NATS.Client.Core/Internal/TcpConnection.cs
mtmk: just trying to understand if we can use OnConnectingAsync + OnSocketThing instead of this change.

wolfkor: No, because OnConnecting would require the socket for the proxy, which is created afterward. Furthermore, you would also have to consider the WebSocket case.
I even considered deleting OnSocketAvailable again. But I wasn't sure if there were other scenarios besides proxy where it would make sense.

src/NATS.Client.Core/Internal/TcpConnection.cs
mtmk: I don't. my concern is how we'd support this going forward. maybe we can have a callback instead?

src/NATS.Client.Core/Internal/TcpConnection.cs
mtmk: that was just an example. you need to check the read count and read more if you need more etc.

src/NATS.Client.Core/Internal/TcpConnection.cs
wolfkor: That would be possible. But I think every methodology has its own fixed format. For additional methods, I would extend properties in NatsProxyOpts. Do you want to leave that entirely up to the user?

mtmk: Not sure. what's your take? I think if we can support it with reasonable confidence just setting up the proxy host and port would be easy for the application. then the edge cases would require a mew release. (btw we must think of the security implications as well)

caleblloyd:
What protocols are you trying to proxy:

- NATS Protocol - Unencrypted
- NATS Protocol - default Opportunistic TLS - [TlsMode.Require](https://nats-io.github.io/nats.net/api/NATS.Client.Core.TlsMode.html)
- NATS Protocol - Implicit TLS (sometimes called TLS Handshake First)  - [TlsMode.Implicit](https://nats-io.github.io/nats.net/api/NATS.Client.Core.TlsMode.html)
- WebSocket - Unencrypted
- WebSocket - TLS

And what kind of proxy are you trying to use?

- SOCKS Proxy
- HTTP Proxy
- Something else?

src/NATS.Client.Core/Internal/TcpConnection.cs
```diff
@@ -57,16 +59,56 @@ public ValueTask ConnectAsync(string host, int port, CancellationToken cancellat
...
+                string? connectAuth = null;
+                if (proxyOpts.Auth != null)
+                {
+                    // Create the CONNECT request with proxy authentication
+                    var auth = Convert.ToBase64String(Encoding.UTF8.GetBytes(proxyOpts.Auth));
+                    connectAuth = $"Proxy-Authorization: Basic {auth}\r\n";
+                }
+
+                var serverWithPort = $"{host}:{port}";
+                var connectBuffer = Encoding.UTF8.GetBytes($"CONNECT {serverWithPort} HTTP/1.1\r\nHost: {serverWithPort}\r\n{connectAuth}Proxy-Connection: Keep-Alive\r\n\r\n");
```
caleblloyd, line 85: I am confused here, this appears to be an HTTP Proxy?  Aren't HTTP Proxies only for proxying HTTP(S) connections?  But this is in the `TcpConnection` class which is used for the NATS protocol, which is not a HTTP-based protocol.

For WebSockets, there is [NatsWebSocketOpts.ConfigureClientWebSocketOptions](https://nats-io.github.io/nats.net/api/NATS.Client.Core.NatsWebSocketOpts.ConfigureClientWebSocketOptions.html) which could be used to set the [ClientWebSocketOptions.Proxy](https://learn.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocketoptions.proxy?view=net-9.0#system-net-websockets-clientwebsocketoptions-proxy) property to a [WebProxy](https://learn.microsoft.com/en-us/dotnet/api/system.net.webproxy?view=net-9.0)

caleblloyd: Ah I see that HTTP CONNECT can work with protocols other than HTTP - [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/CONNECT)

> Aside from enabling secure access to websites behind proxies, a HTTP tunnel provides a way to allow traffic that would otherwise be restricted (SSH or FTP) over the HTTP(S) protocol.

caleblloyd:
I think a more flexible API may be to allow for specifying a `SocketConnectionFactory` in `NatsOpts` that has a signature with something like:

`Func<NatsUri, NatsOpts, CancellationToken, ValueTask<ISocketConnection>>? SocketConnectionFactory`

And that would be responsible for returning an opened ISocketConnection

src/NATS.Client.Core/Internal/TcpConnection.cs
wolfkor: Our current problem to solve for the customer environment is indeed NATS protocol Unencrypted or default Opportunistic TLS with HTTP proxy. The current code solves it

mtmk: I'm concerned we won't be able to support the proxy ourselves giving it the attention it deserves with all the different scenarios. Can you instead implement a IScocketConnection factory like suggested below? we can add that to NatsConnection class switching between the blow two blocks of code:

https://github.com/nats-io/nats.net/blob/b6edd5f4099d14adff75b8477fb162af1663924c/src/NATS.Client.Core/NatsConnection.cs#L338-L357

https://github.com/nats-io/nats.net/blob/b6edd5f4099d14adff75b8477fb162af1663924c/src/NATS.Client.Core/NatsConnection.cs#L644-L666

mtmk:
cc @johannespfeiffer

wolfkor:
For our scenario i had to add 
```
public class ProxyTcpConnection(ILogger<NatsConnection> logger) : TcpConnection(logger)
{
    public new async ValueTask ConnectAsync(string host, int port, TimeSpan timeout)
    {

        using CancellationTokenSource cts = new CancellationTokenSource(timeout);
        try
        {
            if (proxy.Proxy != null)
            {
                await Socket.ConnectAsync(proxy.ProxyHost, proxy.ProxyPort, cts.Token).ConfigureAwait(false);

                string connectAuth = null;
                if (proxy.ProxyAuth != null)
                {
                }

                string serverWithPort = $"{host}:{port}";
                byte[] connectBuffer = Encoding.UTF8.GetBytes(
                    $"CONNECT {serverWithPort} HTTP/1.1\r\nHost: {serverWithPort}\r\n{connectAuth}Proxy-Connection: Keep-Alive\r\n\r\n");

                await SendAsync(connectBuffer).ConfigureAwait(false);

                // Validate proxy response
                byte[]        receiveBuffer   = new byte[4096];
                StringBuilder responseBuilder = new StringBuilder();
                int           read;
                do
                {
                    read = await ReceiveAsync(receiveBuffer).ConfigureAwait(false);
                    responseBuilder.Append(Encoding.UTF8.GetString(receiveBuffer, 0, read));
                } while (read > 0 && !responseBuilder.ToString().Contains("\r\n\r\n"));

                string response = responseBuilder.ToString();
            }
            else
                await Socket.ConnectAsync(host, port, cts.Token).ConfigureAwait(false);
        }
        catch (Exception ex)
        {
            await DisposeAsync().ConfigureAwait(false);
            if (ex is OperationCanceledException)
                throw new SocketException(10060); // 10060 = connection timeout.

            throw;
        }
    }
}
```
and 
```
public class BusSocketConnectionFactory : DefaultSocketConnectionFactory, ISocketConnectionFactory
{
    public new static       BusSocketConnectionFactory       Instance => LazyInstance.Value;
    private static readonly Lazy<BusSocketConnectionFactory> LazyInstance = new(() => new BusSocketConnectionFactory());

    public new async Task<ISocketConnection> OnConnectionAsync(NatsUri uri, NatsConnection natsConnection, ILogger<NatsConnection> logger)
    {
        (string Host, int Port) target = await BeforeConnectingAsync(natsConnection, logger, (uri.Host, uri.Port));

        logger.LogInformation(NatsLogEvents.Connection, $"Try to connect NATS '{uri}' with {nameof(BusSocketConnectionFactory)}");

        ProxyTcpConnection connection = new ProxyTcpConnection(logger);
        await connection.ConnectAsync(target.Host, target.Port, natsConnection.Opts.ConnectTimeout).ConfigureAwait(false);
        return connection;
    }

    public new async Task<ISocketConnection> OnReconnectionAsync(NatsUri uri, NatsConnection natsConnection, ILogger<NatsConnection> logger, int reconnectCount)
    {
        if (natsConnection.OnConnectingAsync != null)
        {
            (string Host, int Port) target = (uri.Host, uri.Port);
            logger.LogInformation(
                NatsLogEvents.Connection,
                $"Try to invoke OnConnectingAsync before connect to NATS [{reconnectCount}] with {nameof(BusSocketConnectionFactory)}",
                reconnectCount);
            (string Host, int Port) newTarget = await natsConnection.OnConnectingAsync(target).ConfigureAwait(false);

            if (newTarget.Host != target.Host || newTarget.Port != target.Port) uri = uri.CloneWith(newTarget.Host, newTarget.Port);
        }

        logger.LogInformation(NatsLogEvents.Connection, $"Tried to connect NATS {uri} [{reconnectCount}] with {nameof(BusSocketConnectionFactory)}");

        NatsOpts opts = natsConnection.Opts;
        logger.LogDebug(NatsLogEvents.Connection, $"Trying to reconnect using TCP {uri} [{reconnectCount}] with {nameof(BusSocketConnectionFactory)}");
        ProxyTcpConnection tcpConnection = new ProxyTcpConnection(logger);
        await tcpConnection.ConnectAsync(uri.Host, uri.Port, opts.ConnectTimeout).ConfigureAwait(false);
        ISocketConnection socket = await AfterReconnectingAsync(natsConnection, uri, opts, logger, reconnectCount, tcpConnection);

        return socket;
    }
}
```
and in NatsOpts instance
`SocketConnectionFactory = BusSocketConnectionFactory.Instance`

mtmk (changes requested):
Thanks @wolfkor appreciate the work but I think we can simplify this a little. I want to make sue the changes are minimal to avoid disturbing existing applications in case of unintended behavior changes we might introduce while not opening up internal APIs as much as possible. hope it makes sense.

src/NATS.Client.Core/DefaultSocketConnectionFactory.cs
```diff
@@ -0,0 +1,110 @@
+using Microsoft.Extensions.Logging;
+using NATS.Client.Core.Internal;
+
+namespace NATS.Client.Core;
+
+public class DefaultSocketConnectionFactory : ISocketConnectionFactory
```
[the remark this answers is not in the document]
wolfkor: no i don't see that. We need a change inside TcpConnection or WebSocketConnection for sending to the proxy. How to hide and give the customer the chance to change it.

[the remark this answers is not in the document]
wolfkor: probably the only way i see is a callback which returns the pure Socket given by the proxy and put it in the constructor of TcpConnection. But the current OnSocketAvailable imlementations will break with that

src/NATS.Client.Core/INatsConnection.cs
```diff
@@ -61,11 +61,6 @@ public interface INatsConnection : INatsClient
     /// </summary>
     Func<(string Host, int Port), ValueTask<(string Host, int Port)>>? OnConnectingAsync { get; set; }
 
-    /// <summary>
-    /// Hook when socket is available.
-    /// </summary>
-    Func<ISocketConnection, ValueTask<ISocketConnection>>? OnSocketAvailableAsync { get; set; }
```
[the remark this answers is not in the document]
wolfkor: we can leave it but down call it - not a smart way i think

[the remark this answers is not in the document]
mtmk: I prefer not to break existing applications

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -327,43 +352,10 @@ private async ValueTask InitialConnectAsync()
...
             {
-                var target = (uri.Host, uri.Port);
-                if (OnConnectingAsync != null)
-                {
-                    _logger.LogInformation(NatsLogEvents.Connection, "Try to invoke OnConnectingAsync before connect to NATS");
-                    target = await OnConnectingAsync(target).ConfigureAwait(false);
-                }
-
-                _logger.LogInformation(NatsLogEvents.Connection, "Try to connect NATS {0}", uri);
-                if (uri.IsWebSocket)
```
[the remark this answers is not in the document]
mtmk: here and in reconnect loop add these:
```diff
                    _logger.LogInformation(NatsLogEvents.Connection, "Tried to connect NATS {Url} [{ReconnectCount}]", url, reconnectCount);
+                    if (Opts.SocketFactory != null)
+                    {
+                        _logger.LogDebug(NatsLogEvents.Connection, "Trying to reconnect using SocketFactory {Url} [{ReconnectCount}]", url, reconnectCount);
+                        var conn = Opts.SocketFactory(url, _disposedCancellationTokenSource.Token);
+                        await conn.ConnectAsync(url, Opts).ConfigureAwait(false);
+                        _socket = conn;
+                    }
+                    else if (url.IsWebSocket)
-                    if (url.IsWebSocket)
                    {
```

with the option addition that should be it.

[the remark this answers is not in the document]
wolfkor: ISocketConnection has no ConnectAsync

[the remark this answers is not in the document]
mtmk: that's even better. we can just return a connected socket and avoid the extra connect call.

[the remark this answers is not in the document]
wolfkor: and the caller has to implement for the return value of  SocketConnectionFactory ISocketConnection on his own? That was one point for unsealing TcpConnection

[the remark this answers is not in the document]
mtmk: yes, we can start with that. there is very little code in TcpConnection with almost no logic. it's not worth trying to reuse it. creates more unnecessary coupling.

mtmk:
thanks for the changes @wolfkor!

btw you need to sign your commits. you will have to force push with a new commit. otherwise repo is set to block unsigned commits. all commits in a pr must be signed.

image

mtmk (approved):
LGTM thanks @wolfkor 

(after commits signed)

will give @caleblloyd a chance to review.

caleblloyd (changes requested):
I think something happened with the diff, it appears a folder was moved and now there are a bunch of changes in here due to that.  Can you revert that?

caleblloyd:
Oh maybe it was caused by merging `main` into this PR, I think this may need to be squashed and rebased

mtmk:
> Oh maybe it was caused by merging `main` into this PR, I think this may need to be squashed and rebased

yes looks like we need a rebase @wolfkor. changes for this PR are only in NatsConnection, NatsOpts and NatsUri classes.

wolfkor:
I had to sign all my commits including the main merge by amend command. Give me a hint what to do. Making changes again in new branch?

mtmk:
> I had to sign all my commits including the main merge by amend command. Give me a hint what to do. Making changes again in new branch?

yes, opening a new pr and reference this one in comments. that might be easier.

mtmk:
* replaced by #826

PR #812, opened 2025-04-07
Update workflow for server 2.10

mtmk:
Since the latest is now 2.11
also split workflows for more reliability


MauriceVanVeen (approved):
LGTM

PR #815, opened 2025-04-09
Release 2.5.14

mtmk:
* Fix heartbeat timer to prevent unnecessary pulls (#794)
* Inbox Subscription Memory leak .NetStandard2.0 (#803)
* Fix reader buffer bug with typo (#799)
* Fix handling for OnNoData when ignoring deletes (#790)


MauriceVanVeen (approved):
LGTM

PR #817, opened 2025-04-09
Fix netstandard bugs and tidy up tests

mtmk:
Move NATS server process management to the lightweight class. Created the slow tests project for the last bit of tests heavily depending on the old NatsServer class.

* bug fix: netstandard ArgumentNullException extension fix
* bug fix: netstandard JetStream name validation fix


galvesribeiro (approved):
LGTM

PR #819, opened 2025-04-10
Make NatsMsg<T>.Build Public

darkwatchuk:
This PR is just to make NatsMsg<T>.Build public , currently internal.

This allows for creation of INatsSub implementations from outside of the library.

Thanks!


mtmk (approved):
LGTM thanks @darkwatchuk

Added docs with a remark that the API is internal to give notice to application developers to be careful about using it if they are more concern about future release stability since i feel we should have the freedom to change this without much concern.

PR #825, opened 2025-04-13
Make NatsJSConstants public

### LINKED ISSUE #824: Make `NatsJSConstants` public
### Proposed change

The static `NatsJSConstants` contains constant fields for stuff like the ACK payload, etc. Consider making it public.

### Use case

These payloads can sometimes be used by applications when they want to perform, for example, ad-hoc acknowledgments (see [this](https://github.com/nats-io/nats.deno/discussions/579#discussioncomment-6823612)).

### Contribution

_No response_

aradalvand:
Closes #824

### DIFF
--- a/src/NATS.Client.JetStream/Internal/NatsJSConstants.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSConstants.cs
@@ -3,7 +3,7 @@ using System.Text;
 
 namespace NATS.Client.JetStream.Internal;
 
-internal static class NatsJSConstants
+public static class NatsJSConstants
 {
     public static readonly ReadOnlySequence<byte> Ack = new(Encoding.ASCII.GetBytes("+ACK"));
     public static readonly ReadOnlySequence<byte> Nak = new(Encoding.ASCII.GetBytes("-NAK"));


src/NATS.Client.JetStream/Internal/NatsJSConstants.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace NATS.Client.JetStream.Internal;
```
mtmk, line 4: could you also move it up a folder?

aradalvand: Oops! Missed that.
Sure.

mtmk: We should move the namespace as well so it doesn't say internal anymore.

mtmk (approved):
LGTM thanks @aradalvand

mtmk:
sorry @aradalvand simple change but taking too much of your time. we need all the commits signed which means you'd need to force push. thanks!

image

aradalvand:
@mtmk No worries!
Just a simple force push? I just did, doesn't seem to have fixed the issue.

PR #826, opened 2025-04-13
Fix proxy with TLS scenario with SocketConnectionFactory

### LINKED ISSUE #802: Communication with TLS bus behind proxy results in SslPolicyErrors.RemoteCertificateNameMismatch

wolfkor:
* fixes #802 
* retry for (#804)

### DIFF
--- a/src/NATS.Client.Core/NatsOpts.cs
+++ b/src/NATS.Client.Core/NatsOpts.cs
@@ -142,6 +142,8 @@ public sealed record NatsOpts
     /// </remarks>
     public BoundedChannelFullMode SubPendingChannelFullMode { get; init; } = BoundedChannelFullMode.DropNewest;
 
+    public Func<NatsUri, NatsOpts, CancellationToken, ValueTask<ISocketConnection>>? SocketConnectionFactory { get; set; } = null;
+
     internal NatsUri[] GetSeedUris(bool suppressRandomization = false)
     {
         var urls = Url.Split(',');
--- a/src/NATS.Client.Core/NatsOpts.cs
+++ b/src/NATS.Client.Core/NatsOpts.cs
@@ -1,9 +1,7 @@
-using System.Net.WebSockets;
 using System.Text;
 using System.Threading.Channels;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
-using NATS.Client.Core.Internal;
 #if NETSTANDARD
 using Random = NATS.Client.Core.Internal.NetStandardExtensions.Random;
 #endif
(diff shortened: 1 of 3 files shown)


mtmk (approved):
LGTM thanks @wolfkor

src/NATS.Client.Core/NatsOpts.cs
```diff
@@ -142,6 +142,8 @@ public sealed record NatsOpts
     /// </remarks>
     public BoundedChannelFullMode SubPendingChannelFullMode { get; init; } = BoundedChannelFullMode.DropNewest;
 
+    public Func<NatsUri, NatsOpts, CancellationToken, ValueTask<ISocketConnection>>? SocketConnectionFactory { get; set; } = null;
```
mtmk, line 145: missed this. it should be `get; init;` not set. we need to maintain opts as readonly.

PR #827, opened 2025-04-13
Use Stopwatch in PingCommand for more precise timing

mtmk:
It's important to use monotonic timer here for correctness.


caleblloyd (approved):
LGTM

PR #828, opened 2025-04-14
Socket Connection Factory

### LINKED ISSUE #802: Communication with TLS bus behind proxy results in SslPolicyErrors.RemoteCertificateNameMismatch

caleblloyd:
- Fixes #802 
- Builds on top of #826
- Introduces public `INatsSocketConnectionFactory`, which is implemented by `TcpConnection` and `WebSocketConnection` or custom in a `SocketConnectionFactory` opt
- Introduces public `INatsTlsUpgradeableSocketConnection` which is implemented by `TcpConnection` or custom in a `SocketConnectionFactory` opt


wolfkor:
This is more in line with my understanding of a factory, but it changes the API ;-). I tried it with an implementation of INatsSocketConnectionFactory and INatsTlsUpgradeSocketConnection, and it seems to work. But shouldn't it be called INatsProxyUpgradeSocketConnection?

caleblloyd:
> This is more in line with my understanding of a factory, but it changes the API ;-

Yes, it seemed more natural to use an interface that the internal and external implementations could both implement

> Shouldn't it be called INatsProxyUpgradeSocketConnection

I don't think that it is specific to proxies, you could implement other kinds of sockets like Unix sockets too

wolfkor:
but also not specific to TLS ;-)

caleblloyd:
`INatsSocketConnection` is the one that is not-specefic to TLS, the library will not attempt to do a TLS upgrade on that socket.

But `INatsTlsUpgradeableSocketConnection` (which extends `INatsSocketConnection`) will cause the library to attempt to do a TLS upgrade if one is warranted.

caleblloyd:
I reduced `INatsSocketConnection` to `SendAsync`, `ReceiveAsync`, and `DisposeAsync`.  Closing the socket is now done with `DisposeAsync`

src/NATS.Client.Core/Internal/TcpConnection.cs
```diff
@@ -14,71 +12,39 @@ public SocketClosedException(Exception? innerException)
...
-#endif
-    }
-
-    /// <summary>
-    /// Connect with Timeout. When failed, Dispose this connection.
-    /// </summary>
-    public async ValueTask ConnectAsync(string host, int port, TimeSpan timeout)
+    public async ValueTask ConnectAsync(Uri uri, CancellationToken cancellationToken)
     {
-        using var cts = new CancellationTokenSource(timeout);
```
mtmk, line 62: was timeout not working? why did we remove it?

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -417,6 +378,51 @@ private async ValueTask InitialConnectAsync()
...
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(_disposedCts.Token, timeoutCts.Token);
+        try
+        {
+            var socketConnection = await connectionFactory.ConnectAsync(uri.Uri, Opts, cts.Token).ConfigureAwait(false);
+            _socketConnection = new SocketConnectionWrapper(socketConnection);
+        }
+        catch (OperationCanceledException) when (timeoutCts.IsCancellationRequested)
+        {
+            throw new SocketException(10060); // 10060 = connection timeout.
+        }
```
caleblloyd, line 406: @mtmk timeout is here now

mtmk (approved):
Awesome feature! LGTM thanks @caleblloyd

PR #831, opened 2025-04-16
Update KV per key TTL

mtmk:
Based on [adr-48](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-48.md)

### DIFF
--- a/src/NATS.Client.KeyValueStore/INatsKVStore.cs
+++ b/src/NATS.Client.KeyValueStore/INatsKVStore.cs
@@ -27,18 +27,6 @@ public interface INatsKVStore
     ValueTask<ulong> PutAsync<T>(string key, T value, INatsSerialize<T>? serializer = default, CancellationToken cancellationToken = default);
 
     /// <summary>
-    /// Put a value into the bucket using the key
-    /// </summary>
-    /// <param name="key">Key of the entry</param>
-    /// <param name="value">Value of the entry</param>
-    /// <param name="ttl">Time to live for the entry (requires the <see cref="NatsKVConfig.AllowMsgTTL"/> to be set to true). For a key that should never expire, use the <see cref="TimeSpan.MaxValue"/> constant.</param>
-    /// <param name="serializer">Serializer to use for the message type.</param>
-    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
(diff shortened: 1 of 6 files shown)


stebet:
I'll take a look in a few hours

src/NATS.Client.KeyValueStore/INatsKVStore.cs
```diff
@@ -26,18 +26,6 @@ public interface INatsKVStore
...
-    /// </summary>
-    /// <param name="key">Key of the entry</param>
-    /// <param name="value">Value of the entry</param>
-    /// <param name="ttl">Time to live for the entry (requires the <see cref="NatsKVConfig.AllowMsgTTL"/> to be set to true). For a key that should never expire, use the <see cref="TimeSpan.MaxValue"/> constant.</param>
-    /// <param name="serializer">Serializer to use for the message type.</param>
-    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
-    /// <typeparam name="T">Serialized value type</typeparam>
-    /// <returns>Revision number</returns>
-    ValueTask<ulong> PutAsync<T>(string key, T value, TimeSpan ttl, INatsSerialize<T>? serializer = default, CancellationToken cancellationToken = default);
-
```
stebet, line 40: Since `Put` is no longer supported, does that mean to update a TTLd key one should delete and recreate it.

stebet (approved):

src/NATS.Client.KeyValueStore/INatsKVStore.cs
mtmk: I think so though not sure how that goes with delete markers etc. let me ask around.

mtmk: ok the proper solution is in the pipeline but for a workaround we purge with a TTL (to keep bucket size maintained) then recreate with new TTL. also for caching we probably want a 1-history buket.

stebet: I'll take a look at the `IDistributedCache` implementation I have to take that into account :)

mtmk: released in [2.6.0-preview.5](https://www.nuget.org/packages/NATS.Net/2.6.0-preview.5)

Issue #834, opened 2025-04-16
NatsJSMsg is missing a ReplyAsync<TReply> method


yeongjonglim:
### Proposed change

At the moment, the NatsJSMsg<T> only support replying message with an empty message. I would like to be able to reply with some data even when we are using NatsJSMsg<T>. I have looked into the [current implementation](https://github.com/nats-io/nats.net/blob/59c530539c2e9c546aa60664582dc115a2ffb143/src/NATS.Client.JetStream/NatsJSMsg.cs#L87), it would be a forwarding to the existing NatsMsg<T> implementation, I believe we can simply follow the same.

### Use case

To provide the convenience to reply to a message even when we are in the context of JetStream.

### Contribution

I'm happy to submit a PR if the above proposal is accepted.

mtmk:
thanks @yeongjonglim tbh i can't remember why we didn't implement that. please go ahead an open a PR. also please check if other implementations have the same concept e.g. Go, Java etc. it's important we maintain parity among different client implementations.

yeongjonglim:
@mtmk , I have taken a quick look into nats.go implementation of JetStream message, it does not seems like they have any `Reply` interface. Their [current implementation](https://github.com/nats-io/nats.go/blob/main/jetstream/message.go) of `Reply` returns the subject to publish to, which is a completely different thing to what the current `ReplyAsync` does in this library. Other than that, I do not find any mechanisms that interface a JetStream message to execute reply.

I will still submit a PR, please let me know if you think this method is not suitable to live in the library.

mtmk:
it should be fine i guess. i'm not sure if i know a case where one would need this. do you have a use case for it or just saw that missing from the interface?

mtmk:
ok I might be missing something but message from the consumer would have a server specific reply-to iirc. did you try the change you're thinking about locally? please feel free to open the PR if it's working for you. just want to make sure you have a solution.

yeongjonglim:
Hey @mtmk, I have done a round of manual testing with the following scenarios:

1. Client is using Request-Reply to a subject in a Stream, based on all default `StreamConfig` targeting the stream subject, The result of it is immediate response from the Stream that looks something like this `{"stream":"job", "seq":235}`
2. I thought that it's because of the Stream Ack that is causing the `RequestAsync` to return immediately, and added the `NoAck` configuration to the Stream, however, the current `ReplyAsync` (with empty message) is not actually received at all and continue to await indefinitely.
[...]

PR #835, opened 2025-04-17
Add methods to NatsJsMsg (#834)

yeongjonglim:
This is to resolve [834](https://github.com/nats-io/nats.net/issues/834).


mtmk:
thanks @yeongjonglim. sounds interesting! if you could share an example for your use case that would be appreciated by the community as well 💯

yeongjonglim:
Closed as this method is not supported in NATS

PR #837, opened 2025-04-18
Adding pinned_ids to ConsumerInfo.cs

scottf:


scottf:
This feature did not make it to the server

PR #839, opened 2025-04-24
Obsolete ReplyAsync method for NatsJsMsg

### LINKED ISSUE #834: NatsJSMsg is missing a ReplyAsync<TReply> method

yeongjonglim:
This is to resolve #834

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSMsg.cs
+++ b/src/NATS.Client.JetStream/NatsJSMsg.cs
@@ -84,6 +84,7 @@ public interface INatsJSMsg<out T>
     /// <param name="opts">A <see cref="NatsPubOpts"/> for publishing options.</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the command.</param>
     /// <returns>A <see cref="ValueTask"/> that represents the asynchronous send operation.</returns>
+    [Obsolete("ReplyAsync is not valid when using JetStream. The reply message will never reach the Requestor as NATS will reply to all JetStream publish by default.")]
     ValueTask ReplyAsync(NatsHeaders? headers = default, string? replyTo = default, NatsPubOpts? opts = default, CancellationToken cancellationToken = default);
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


mtmk (approved):
LGTM thanks @yeongjonglim

src/NATS.Client.JetStream/NatsJSMsg.cs
```diff
@@ -208,6 +208,7 @@ public NatsJSMsg(NatsMsg<T> msg, INatsJSContext context)
     /// <param name="opts">A <see cref="NatsPubOpts"/> for publishing options.</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the command.</param>
     /// <returns>A <see cref="ValueTask"/> that represents the asynchronous send operation.</returns>
+    [Obsolete("ReplyAsync is not valid when using JetStream. The reply message will never reach the Requestor as NATS will reply to all JetStream publish by default.")]
```
mtmk, line 211: just noticed actually. we have this method on the interface `INatsJSMsg` too

yeongjonglim: I can include the Obsolete message on the interface too. Do you think that will be too verbose? Or just putting it on the interface would be sufficient?

mtmk:
just moved this to 2.7 release

src/NATS.Client.JetStream/NatsJSMsg.cs
mtmk: we should put it on both

yeongjonglim: Just added on the interface as well.

mtmk (approved):

PR #844, opened 2025-05-01
Fix reconnect socket behavior

mtmk:
Received a report of reconnect failure in Azure container environment. Suspecting TCP TIME_WAITs because of potential proxy handling in between the server and the client.

changes:
* revert the dispose order to what it was before: first set waitForClosed then dispose the socket
* revert to using Socket.Shutdown rather than Socket.Disconnect: this might have an effect on the underlying socket usage with pending data being sent or received and/or how tcp behaves on the wire especially when proxied e.g. if the proxy keeps the connection open.

 I admit it's a bit of a shot in the dark since despite my efforts on various setups I could not reproduce the issue. but my thinking is that socket.shutdown is probably the right thing to do since we don't want to reuse the underlying socket (which Socket.Disconnect seems to be aimed at) also triggering waitForClosed before socket disposal would avoid timeouts or hangs on the socket immediately starting the reconnect process.

### DIFF
--- a/src/NATS.Client.Core/Internal/HeaderWriter.cs
+++ b/src/NATS.Client.Core/Internal/HeaderWriter.cs
@@ -20,7 +20,7 @@ internal class HeaderWriter
 
     private static ReadOnlySpan<byte> ColonSpace => new[] { ByteColon, ByteSpace };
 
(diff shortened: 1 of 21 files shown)


caleblloyd (approved):
LGTM

PR #852, opened 2025-05-11
Add TTL versions of UpdateAsync/TryUpdateAsync to INatsKvStore

caleblloyd:
Expose the ttl-versions of update


mtmk:
sorry for the delay on this one @caleblloyd. i'm guessing it's for extending the ttl i.e. sliding expirations? I'm not sure if this works as expected.

btw is this same subject would you say? https://github.com/nats-io/nats.net/pull/831#discussion_r2056658587

mtmk:
There is a suggestion against put here https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-48.md#storing-values

mtmk (approved):
LGTM

PR #853, opened 2025-05-12
#848 Tweak dependencies

### LINKED ISSUE #848: Optimise dependencies based on framework
### Proposed change

Remove dependencies which can be provided by the framework natively ie System.Text.Json & System.io.pipelines which are dependencies on the core package 

### Use case

I want to minimise dependencies in my project by utilising framework dependencies wherever possible

### Contribution

_No response_

thompson-tomo:
Closes #848

### DIFF
--- a/src/NATS.Client.Core/NATS.Client.Core.csproj
+++ b/src/NATS.Client.Core/NATS.Client.Core.csproj
@@ -37,7 +37,6 @@
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
         </PackageReference>
-        <PackageReference Include="System.IO.Pipelines" Version="8.0.0"/>
     </ItemGroup>
 
     <!--  Dependencies for netstandard only  -->
@@ -49,6 +48,7 @@
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
         </PackageReference>
+        <PackageReference Include="System.IO.Pipelines" Version="8.0.0"/>
     </ItemGroup>
 
     <!--  Dependencies for netstandard2.0 only -->
@@ -61,9 +61,8 @@
         <PackageReference Include="IndexRange" Version="1.0.3"/>
     </ItemGroup>
 
-    <!--  Dependencies for net6.0 and net8.0 only -->
-    <ItemGroup Condition="'$(TargetFramework)' == 'net6.0' OR '$(TargetFramework)' == 'net8.0'">
-        <PackageReference Include="System.Text.Json" Version="8.0.5"/>
+    <ItemGroup Condition="'$(TargetFramework)' == 'net6' OR '$(TargetFramework)' == 'net8'">
+        <PackageReference Include="System.IO.Pipelines" Version="8.0.0"/>
     </ItemGroup>
 
     <ItemGroup>


src/NATS.Client.Core/NATS.Client.Core.csproj
```diff
@@ -37,7 +37,6 @@
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
         </PackageReference>
-        <PackageReference Include="System.IO.Pipelines" Version="8.0.0"/>
```
mtmk, line 40: so i guess >net9 includes this in the sdk? (sorry didn't check)
since we're not targeting net9 i think this dependency will always come through.

thompson-tomo: Correct on both regards. This way when net 9+ is eventually added the dependency conditions doesn't need to be touched.

mtmk: should we also target net9 then?

thompson-tomo: Sure, we could either add the additional TFM via this PR or via a seperate issue/PR. Let me know your preference.

thompson-tomo:
> are the applications going to get a benefit as a result of this?.... exposing the applications to security issues if the build servers still running old sdks?

This depends on how the end application is packaged:
- Self contained: if build server sdk is not maintained then yes risk is increased as the deployment bundles in the version provided by sdk. But the security risk overall is not significantly increased due to other security risks existing 
- framework dependent: security risks can be addressed by updating framework on target machine ie windows update. This is because the version provided by the framework is used

> would we be breaking builds 
Only scenario I can foresee is where another library/app targets net 6/7, uses a feature introduced in STJ 7/8 and doesn't have either an explicit or transitive dependency for a version of STJ which includes the necessary feature.

src/NATS.Client.Core/NATS.Client.Core.csproj
mtmk: is there a benefit you see us supporting net9 tfm? i guess not having a reference to SIOP is one.

src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
```diff
@@ -27,6 +27,12 @@
   <!--  Dependencies for netstandard only  -->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netstandard2.1'">
     <PackageReference Include="System.ComponentModel.Annotations" Version="5.0.0" />
+    <PackageReference Include="System.Text.Json" Version="8.0.5"/>
+  </ItemGroup>
+
+  <!--  Dependencies for net6.0 only -->
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <PackageReference Include="System.Text.Json" Version="8.0.5"/>
   </ItemGroup>
```
mtmk, line 36: did net6 not receive the security patch being out of support? we should add this to core as well for net6 tfm, no?

src/NATS.Client.Core/NATS.Client.Core.csproj
thompson-tomo: Other than SIOP not aware of a benefit but when net 10 is out perhaps we add that due to being an LTS and we get the benefit.

src/NATS.Client.JetStream/NATS.Client.JetStream.csproj
thompson-tomo: What I discovered was that this project was using a new feature introduced with STJ 8 which it could previously use as STJ 8 was a transitive dependency being provided via the core project.

I much prefer to have my explicit dependencies only in the projects which actually need them.

thompson-tomo:
@mtmk any idea why package restore is only failing on 1 os. Maybe it needs to be re-triggered?

src/NATS.Client.Core/NATS.Client.Core.csproj
mtmk: ok sounds good. that is our current policy i.e. following LTS but I'm always wondering if there is a benefit that would warrant adding other TFMs if useful for projects.

thompson-tomo:
Agree @mtmk it seems very interesting and that the issue is infrastructure related given that the most recent run only failed 1 project which is now talking about an ssl verification issue. Can it get another kick?

mtmk:
just kicked the build again. i'm guessing something funny going on with ci workers nuget sources then when they failed. let's see.

mtmk:
seeing these at the bottom of the diff tab. i wonder if that's a netstandard/net6 issue
image

thompson-tomo:
Ok based on https://github.com/dotnet/runtime/issues/58770 they look to be false warnings. Let's see if the suggestion in that issue resolves the warnings. 🤞

thompson-tomo:
@mtmk did the last commit resolve the warnings?

mtmk:
> @mtmk did the last commit resolve the warnings?

just kicked the checks

mtmk:
may i move this PR onto release/2.7?

thompson-tomo:
Yes that is fine with me.

PR #856, opened 2025-05-20
Add Extension Points

mtmk:
Introduce `NatsExtensionPoints` to enable customization of the NATS client via interfaces for connection, message interception, and subscription management.

Extension Points formalize the existing extensions being used in the client and open the door for additional features such as message interceptors being able to transform messages including the headers.


mtmk:
release/2.7 branch is obsolete now. need to move this change to base main. (if still valid)

mtmk:
callbacks on NatsOps is working well. any other extensions can be introduced gradually rather than a big API change.

PR #859, opened 2025-05-22
Add support for `Filter` and `Enrich` for OpenTelemetry activities

### LINKED ISSUE #857: Add `Filter` hook for OpenTelemetry activities
### Proposed change

Apparently OpenTelemetry support was added to this client as part of https://github.com/nats-io/nats.net/pull/437 (which is great); but it's not really documented, and you have to manually add an `.AddSource("NATS.Net")` call like this:
```cs
builder.Services.AddOpenTelemetry()
	.WithTracing(b => b
		.AddAspNetCoreInstrumentation()
		.AddHttpClientInstrumentation()
		.AddSource("NATS.Net")
		.AddMongoInstrumentation()
		.AddRedisInstrumentation()
        );
```
I only found this out by digging into the source code. It would be great if a `.AddNatsInstrumentation()` extension method was provided out of the box, or at least this were mentioned in the client library's documentation.

But more importantly, by enabling the `NATS.Net` activity source name, MASSIVE AMOUNTS of spans are generated if your application subscribes to even a few streams, etc. (e.g. 10,000 spans in 3 minutes in my case):

Image

Consider providing a mechanism for filtering out/narrowing down trace data — similar to MongoDB's `ShouldStartActivity`, or ASP.NET Core's `Filter`, for example:
https://github.com/jbogard/MongoDB.Driver.Core.Extensions.DiagnosticSources/blob/cf248a881f12cd49754a1530a3b8e5f827ba2892/src/MongoDB.Driver.Core.Extensions.DiagnosticSources/InstrumentationOptions.cs#L6-L10

Image

Image

Most major instrumentation libraries provide a `Filter` parameter. And this is also something the OpenTelemetry docs suggest [as a common convention](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/extending-the-sdk#filtering-processor:~:text=Most%20instrumentation%20libraries%20shipped%20from%20this%20repo%20provides%20a%20built%2Din%20Filter%20option%20to%20achieve%20the%20same%20effect.).

### Use case

Outlined above.

### Contribution

_No response_

aradalvand:
Resolves #857

Several important considerations:
- ~~1 dependency added: `OpenTelemetry.Api` to `NATS.Client.Core` — needed for the `TracerProviderBuilder.AddNatsInstrumentation` extension method. This makes NATS dependent on the so-called OpenTelemetry API, [which is expected](https://opentelemetry.io/docs/specs/otel/overview/#api). But if you're strongly opposed to this, we could also remove this method. Won't take away much from the core functionality added here which are `Filter` and `Enrich`.~~ (Dependency removed)
- Since `Telemetry.Send*` methods were also used in places without dependency injection available (e.g. the `NatsMsg` struct), I couldn't integrate `NatsInstrumentationOptions` into the `IOptions` system (which would've probably been more ideal) — so I created a static `Default` property instead, which the `Action<NatsInstrumentationOptions>` input parameter effectively configures, and that property is then used by the `Telemetry` class.
- I think it would also be useful to add an `IsInternal` flag to `NatsInstrumentationContext` (equivalent to the old `NATS.Net.Internal` activity source name) so that people could easily filter out lower-level communication (e.g. for request-replies), which could be verbose and chatty.

Made a number of tangential improvements too (if you disagree with any of them let me know and I could revert them back):
- Made the methods of the internal `Telemetry` class public — this is the most conventional pattern in .NET (which this same codebase also follows in almost all other places).
- Added `isRemote: true` to `ActivityContext.TryParse` — the OpenTelemetry specification states that span contexts extracted from text maps using propagators (e.g. `DistributedContextPropagator`) should be `isRemote: true` — see [this link](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isremote).
- In `NatsMsg<T>.Build` I moved the code for activity creation before `Deserialize` is called, so that if the user's custom deserializer spawns an activity (which mine does, for example, so this would be useful for me), that's included in the trace. This makes things symmetric with how `RequestAsync` and `PublishAsync` work as well (they encompasses the ser/de step in their activities).


src/NATS.Client.Core/Internal/NatsInstrumentationOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// The return value for the filter function is interpreted as follows:
+    /// - If filter returns `true`, the request is collected.
+    /// - If filter returns `false` or throws an exception the request is NOT collected.
+    /// </remarks>
+    public Func<NatsInstrumentationContext, bool>? Filter { get; set; }
+
+    /// <summary>
+    /// Gets or sets an action to enrich an Activity.
+    /// </summary>
+    public Action<Activity, NatsInstrumentationContext>? Enrich { get; set; }
```
thompson-tomo, line 25: Wondering what the use case would be for this level of control and I feel this gives away too much control. At most an app should only be able to add tags.

thompson-tomo:
See couple of comments but I would also suggest waiting for #859 to be merged which would enable the  context & options to reside in the abstractions package.

src/NATS.Client.Core/NATS.Client.Core.csproj
```diff
@@ -1,88 +1,89 @@
...
+
+    <!--  Dependencies for all -->
+    <ItemGroup>
+        <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.1" />
+        <PackageReference Include="Microsoft.Extensions.Primitives" Version="6.0.0" />
+        <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.10.48">
+            <PrivateAssets>all</PrivateAssets>
+            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+        </PackageReference>
+        <PackageReference Include="OpenTelemetry.Api" Version="1.8.1" />
```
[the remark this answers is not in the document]
mtmk: yea we shouldn't add dependencies to otel in the core libs and only rely on existing .net types for telemetry so that people who don't need otel aren't forced to include assemblies they don't use.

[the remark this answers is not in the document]
aradalvand: Okay; I'll remove this. We can create a separate package in a different PR.

src/NATS.Client.Core/Internal/NatsInstrumentationOptions.cs
aradalvand: Nearly all major instrumentation libraries have this. This is a very common convention:
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/03c577940d9b3eb223e21af98ac8eeb692e8e5c9/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreTraceInstrumentationOptions.cs#L61-L68
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/03c577940d9b3eb223e21af98ac8eeb692e8e5c9/src/OpenTelemetry.Instrumentation.SqlClient/SqlClientTraceInstrumentationOptions.cs#L66-L83
and so on and so forth.

Also, quoted from the official OpenTelemetry docs:
> Many [instrumentation libraries](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/extending-the-sdk#instrumentation-library) shipped from this repo provides a built-in Enrich option, which may also be used to enrich activities. Instrumentation library provided approach may offer additional capabilities such as offering easy access to more context (library specific). [Here](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/extending-the-sdk#filtering-processor:~:text=Many%20instrumentation%20libraries%20shipped%20from%20this%20repo%20provides%20a%20built%2Din%20Enrich%20option%2C%20which%20may%20also%20be%20used%20to%20enrich%20activities.%20Instrumentation%20library%20provided%20approach%20may%20offer%20additional%20capabilities%20such%20as%20offering%20easy%20access%20to%20more%20context%20(library%20specific).)

aradalvand:
> See couple of comments but I would also suggest waiting for https://github.com/nats-io/nats.net/pull/859 to be merged which would enable the context & options to reside in the abstractions package.

Waiting for what to be merged, sorry? #859 is this same PR.

src/NATS.Client.Core/Internal/NatsInstrumentationOptions.cs
thompson-tomo: Ok, I was just surprised to see the whole activity being provided rather than just the tags.

thompson-tomo: Thoughts about making this Action<Activity, IConnection>EnrichWithConnection { get; set; }

aradalvand: `IConnection` is included in `NatsInstrumentationContext`.

thompson-tomo: It follows the examples which you pointed out and it eliminates the need for null checks as connection/headers could be null. The other values are already set as tags as per otel

thompson-tomo (changes requested):

src/NATS.Client.Core/Internal/Telemetry.cs
```diff
@@ -23,6 +22,18 @@ internal static class Telemetry
...
+        var instrumentationContext = new NatsInstrumentationContext(
+            Subject: subject,
+            Headers: null,
+            ReplyTo: replyTo,
+            QueueGroup: null,
+            BodySize: null,
+            Size: null,
+            Connection: connection);
+
+        if (NatsInstrumentationOptions.Default.Filter is { } filter && !filter(instrumentationContext))
```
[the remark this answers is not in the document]
aradalvand: Absolutely not. The whole point of `Filter` is that the activity becomes null.

src/NATS.Client.Core/Internal/NatsInstrumentationOptions.cs
aradalvand: The only example this follows is the ASP.NET Core one, which is not analogous. Those separate `Enrich*` methods correspond to different contexts/scenarios altogether (receiving a request, sending a response, an exception being thrown); rather than different pieces of data belonging to the same context (which is essentially what you're suggesting).

thompson-tomo: Not the case as both the request & the response attributes are in fact part of both the client span as well as the server span [http-spans](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span)

src/NATS.Client.Core/Internal/Telemetry.cs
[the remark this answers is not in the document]
thompson-tomo: Take a look at [extending-the-sdk](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/extending-the-sdk#filtering-processor) that is the docs on extending the sdk for filtering.

I am aware that a processor could apply additional filtering to the spans but as mentioned in my previous comment
> so we can leverage the span duration for timing metrics but also and more significantly the activity tags can then be easily added to the metric

I agree that you could get away without creating the span in the scenario where the span is not being used for metrics especially ones where it is capturing timing but this scenario is not one of them.

aradalvand:
No idea why the Windows test is falling — I thought my last change caused by I undid that it's still failing; very strange! 😅

aradalvand:
I have no idea why the Windows test is failing... Weird.

mtmk:
may i move this PR onto release/2.7 branch? it might make it easier to coordinate with other otel stuff for us also we can hopefully release preview quicker

aradalvand:
Sure @mtmk feel free to do that.

src/NATS.Client.Core/Internal/Telemetry.cs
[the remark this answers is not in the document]
aradalvand: I'll research this further and get back to you.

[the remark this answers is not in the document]
aradalvand: I frankly fail to understand your point here. If what you're after is for the recorded flag to be omitted from the span, you could already do that either with a processor or with `Enrich` — a `Filter` that prevents span creation altogether won't hurt your ability to achieve what you want, all it does is unlock a different use case.

Moreover, creating a non-recorded span is functionally different from not creating a span in one crucial respect: If we create non-recorded spans, and a `ParentBasedSampler` is used — which it almost always is [in a production setting](https://opentelemetry.io/docs/languages/go/sampling/#:~:text=When%20in%20a%20production%20environment%2C%20consider%20using%20the%20ParentBased%20sampler%20with%20the%20TraceIDRatioBased%20sampler) — any child spans (i.e. spans that happen to created while the the "filtered" (recorded-flag-removed) span is active, and they might even be in user code) won't be sampled at all (even though we might want them to be).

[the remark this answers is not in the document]
thompson-tomo: let me try this a different way to make it clearer and to show that in fact what you are proposing would become an issue for implementing metrics or at worse would lead to wrong Metrics. 

NATS generate spans only based on 1 of 2 cases:

- When a message is being sent
- When a message is received

Both of those spans are considered messaging operation's and as the OTEL documentation points out the duration is required to be recorded and emitted as a metric. The document even states:

> When this metric is reported alongside a messaging span, the metric value SHOULD be the same as the corresponding span duration.

As such to ensure that the durations always match, the most efficient way to do this is to simply use the duration of the span rather than introduce stopwatches for when the activity is not being created but when they are use the duration from the activity. 

I agree with you spans could be simply not created if they are not needed for scenarios where those spans were not also having their duration reported as Metrics ie serialisation, but NATS does not currently have a scenario such as that.

At the same time looking at the issue you raised, i wonder if the root issue has been missed and it could be addressed in another way so that all users get the benefit without needing to write a filter. What i mean by this is, is the activity being started in the correct location (when a message is being built) or should it be moved to PublishToClientHandlersAsync within SubscriptionManager for example? Could we have a flag to enable/disable

src/NATS.Client.Core/Internal/NatsInstrumentationOptions.cs
```diff
@@ -0,0 +1,26 @@
+using System.Diagnostics;
+
+namespace NATS.Client.Core.Internal;
+
+/// <summary>
+/// Options for the OpenTelemetry instrumentation.
+/// </summary>
+public sealed class NatsInstrumentationOptions
```
mtmk, line 8: if this is public we should move it out internal

src/NATS.Client.Core/Internal/Telemetry.cs
```diff
@@ -6,14 +6,13 @@ namespace NATS.Client.Core.Internal;
 // https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#messaging-attributes
 internal static class Telemetry
 {
-    internal static readonly ActivitySource NatsActivities = new(name: NatsActivitySource);
-
-    private const string NatsActivitySource = "NATS.Net";
+    public const string NatsActivitySource = "NATS.Net";
+    public static readonly ActivitySource NatsActivities = new(name: NatsActivitySource);
```
mtmk, line 10: do we want this to be public? the class is internal? maybe a separate public class?

aradalvand: No it will remain internal — making `internal` class members `public` is just a common convention; see [my original comment](https://github.com/nats-io/nats.net/pull/859#:~:text=Made%20the%20methods%20of%20the%20internal%20Telemetry%20class%20public%20%E2%80%94%20this%20is%20the%20most%20conventional%20pattern%20in%20.NET%20(which%20this%20same%20codebase%20also%20follows%20in%20almost%20all%20other%20places).).

mtmk (approved):
LGTM

PR #869, opened 2025-05-24
Cleanup Csproj

thompson-tomo:
Increase usage of directory.build.props to reduce replication of properties across files. Have also move to suppression in code rather than global,


thompson-tomo:
No worries and the dotnet format issue should be resolved now.

mtmk (approved):
LGTM thanks @thompson-tomo cleaned up the proj files really nicely. sorry for the delay.

PR #875, opened 2025-06-01
Fix server install for test

mtmk:


thompson-tomo:
@mtmk can you merge this into the 2.7 branch so that the failing PR pipelines can be rebased etc

PR #871, opened 2025-05-24
#316 conversion to OTEL Metrics

thompson-tomo:
This PR focusses on improving the open telemetry integration by introducing support for metrics and refinement of the spans

The duration of the following operations are captured as both spans & duration metrics 
- publish
- Create
- Send
- Process
- Subscribe
- Unsubscribe

Additional metrics include:
- subscription count
- Sent bytes
- Received bytes
- Qty of messages sent
- Qty of messages received

### Prequiste work
- #876 enabling easy access to the properties of the operation
- #877 promotion of serialisation to occur as close to app to enable apps to potentially do their own serialisation
- #878 consolidate incoming message handling to achieve minimal differences.

Closes #316

PR #876, opened 2025-06-02
Refactor: pass around record of properties for operation

thompson-tomo:
This does some basic refactoring to prepare for https://github.com/nats-io/nats.net/pull/871 with the key focus being to pass record of operation properties rather than multiple variables.

As part of this change the following has been introduced:
- Nats{{operation}}Props: which are used to package up the properties and past through the call stack

### Potential Breaking changes:

- ReceiveInternalAsync within NatsSubBase has introduced a new overload designed in a backward compatitable way


src/NATS.Client.Core/NatsSubBase.cs
```diff
@@ -262,7 +291,20 @@ public virtual ValueTask DisposeAsync()
     /// <param name="replyTo">Reply subject received for this subscription.</param>
     /// <param name="headersBuffer">Headers buffer received for this subscription.</param>
     /// <param name="payloadBuffer">Payload buffer received for this subscription.</param>
+    [Obsolete("Nats processing props should be passed instead")]
     public virtual async ValueTask ReceiveAsync(string subject, string? replyTo, ReadOnlySequence<byte>? headersBuffer, ReadOnlySequence<byte> payloadBuffer)
+    {
+        await ReceiveAsync(new NatsProcessProps(subject, 0, "UNKNOWN"), headersBuffer, payloadBuffer).ConfigureAwait(false);
```
mtmk, line 297: what is 'UNKNOWN' we keep passing?

thompson-tomo:
Agree @mtmk that we need to be careful not to impact existing dependencies hence the new internal methods  etc. This is almost done with the main outstanding being review of accessibility modifiers, code comments & the UNKNOWN param scenario.

thompson-tomo:
@mtmk So it might be difficult to get reliable benchmarks given the current setup is producing such significant variance in results but based on last results & comparing it to a test from when I cleaned up the csproj file. The throughput has increased but at the same time the memory usage also went up. At the same time the allocated memory also sharply rose but not sure if due to short test run.

src/NATS.Client.Core/NatsPublishProps.cs
```diff
@@ -0,0 +1,19 @@
+namespace NATS.Client.Core;
+
+public record NatsPublishProps : NatsMessagingProps
```
thompson-tomo, line 3: Needs additional constructor that takes a NatsSubject

src/NATS.Client.Core/NatsMessagingProps.cs
```diff
@@ -0,0 +1,44 @@
+namespace NATS.Client.Core;
+
+public record NatsMessagingProps : NatsOperationProps
```
thompson-tomo, line 3: Needs additional constructor that takes a NatsSubject

src/NATS.Client.Core/NatsMessagingProps.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    internal int HeaderLength { get; set; }
+
+    internal int TotalMessageLength { get; set; }
+
+    internal int FramingLength { get; set; }
+
+    internal int TotalEnvelopeLength => TotalMessageLength + FramingLength;
+
+    public void SetReplyTo(string replyToTemplate, object replyToId)
```
thompson-tomo, line 32: Needs additional overload that takes a NatsSubject

src/NATS.Client.Core/NatsProcessProps.cs
```diff
@@ -0,0 +1,41 @@
+namespace NATS.Client.Core;
+
+public record NatsProcessProps : NatsMessagingProps
```
thompson-tomo, line 3: Needs additional constructor that takes a NatsSubject

src/NATS.Client.Core/NatsSubject.cs
```diff
@@ -0,0 +1,69 @@
+using System.Text.RegularExpressions;
+
+namespace NATS.Client.Core;
+
+public record NatsSubject
```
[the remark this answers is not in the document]
thompson-tomo: Have taken out all subject handling and will be addressed via https://github.com/nats-io/nats.net/issues/879. Hence closing this

src/NATS.Client.JetStream/NatsJSConsumer.cs
```diff
@@ -308,8 +308,7 @@ internal async ValueTask<NatsJSConsume<T>> ConsumeInternalAsync<T>(INatsDeserial
             stream: _stream,
             consumer: _consumer,
             context: _context,
-            subject: inbox,
-            queueGroup: default,
+            props: new NatsSubscribeProps(inbox),
```
thompson-tomo, line 311: Need to find a way to ensure that subscription id is being set for this scenario.

mtmk:
release/2.7 branch is obsolete now. need to move this change to base main. (if still valid)

PR #877, opened 2025-06-03
Do Serialisation in main to enable simplification

thompson-tomo:
This does some basic refactoring to prepare for https://github.com/nats-io/nats.net/pull/871 with the focus being moving serialisation into nats connection. This is to enable splitting of sending vs serialisation which is needed for OTEL.

As part of this change the following has been introduced:
- The serialisation occurs within the methods of nats connection to enable the publishing to focus on dealing with the bytes

In the future further optimisations could be done to improve performance for retried messages by avoiding re-serialising the content


thompson-tomo:
Thanks normally I would not have rebased but the branches needed it to fix the pipeline issue.

src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
```diff
@@ -247,12 +247,11 @@ internal override async ValueTask WriteReconnectCommandsAsync(CommandWriter comm
                 MinAckPending = _priorityGroup?.MinAckPending ?? 0,
             };
 
-            await commandWriter.PublishAsync(
```
mtmk, line 250: reason for using command writer here is that this particular one is priorityCommandWriter which is processed right after reconnect before anything else is processed.

src/NATS.Client.JetStream/Internal/NatsJSFetch.cs
```diff
@@ -192,13 +192,11 @@ internal override async ValueTask WriteReconnectCommandsAsync(CommandWriter comm
...
         };
-
-        await commandWriter.PublishAsync(
+        await _context.Connection.PublishAsync(
             subject: $"{_context.Opts.Prefix}.CONSUMER.MSG.NEXT.{_stream}.{_consumer}",
-            value: request,
+            data: request,
             headers: default,
             replyTo: Subject,
-            serializer: NatsJSJsonSerializer<ConsumerGetnextRequest>.Default,
```
mtmk, line 201: same

mtmk (changes requested):
early buffer pool returns seems to be the main issue

src/NATS.Client.Core/NatsConnection.Publish.cs
```diff
@@ -55,18 +51,71 @@ public ValueTask PublishAsync<T>(string subject, T? data, NatsHeaders? headers =
...
+                payloadBuffer = new NatsPooledBufferWriter<byte>(_arrayPoolInitialSize);
+
+            if (headers != null)
+                _headerWriter.Write(headersBuffer!, headers);
+
+            if (data != null)
+                serializer.Serialize(payloadBuffer, data);
+            task = DoPublishAsync(subject, headersBuffer?.WrittenMemory, payloadBuffer.WrittenMemory, replyTo, cancellationToken);
+        }
+        catch (Exception ex)
```
mtmk, line 93: these aren't needed

src/NATS.Client.Core/NatsConnection.RequestReply.cs
```diff
@@ -42,10 +42,12 @@ public async ValueTask<NatsMsg<TReply>> RequestAsync<TRequest, TReply>(
 
                 if (Opts.RequestReplyMode == NatsRequestReplyMode.Direct)
                 {
-                    using var rt = _replyTaskFactory.CreateReplyTask(replySerializer, replyOpts.Timeout);
```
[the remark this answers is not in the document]
thompson-tomo: The serializer is only needed to build the natsmsg which has been moved out of the reply task and is now on line 50 hence argument would not be used.

mtmk:
> It looks to me like the deserialisation of the response is failing when using jetstream subject and serializing go a string.

this maybe related to removed serializer arguments

src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
thompson-tomo: Interesting as the method shows it is a standard commandwritter. The reason for moving to connection is so that it can be traced.

src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
```diff
@@ -247,12 +247,11 @@ internal override async ValueTask WriteReconnectCommandsAsync(CommandWriter comm
...
+            await Connection.PublishAsync(
                 subject: $"{_context.Opts.Prefix}.CONSUMER.MSG.NEXT.{_stream}.{_consumer}",
-                value: request,
+                data: request,
                 headers: default,
                 replyTo: Subject,
-                serializer: NatsJSJsonSerializer<ConsumerGetnextRequest>.Default,
```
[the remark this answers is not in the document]
thompson-tomo: Have an idea on how to address this without it becoming a breaking change.

src/NATS.Client.Core/Internal/ReplyTask.cs
```diff
@@ -53,7 +49,7 @@ public override void SetResult(string? replyTo, ReadOnlySequence<byte> payload,
     {
         lock (_gate)
         {
-            _msg = NatsMsg<T>.Build(Subject, replyTo, headersBuffer, payload, _connection, _connection.HeaderParser, _deserializer);
+            _msg = new NatsRecievedEvent(Subject, replyTo, headersBuffer, payload);
```
mtmk, line 52: buffers need to be processed here. deferring it to be processed later buffers will be in an unknown state because they are reused for socket reads.

src/NATS.Client.Core/NatsConnection.RequestReply.cs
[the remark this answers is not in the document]
mtmk: right saw that now. again similar issue with buffers. see my comment below on ReplyTask.cs.

src/NATS.Client.Core/NatsConnection.Publish.cs
thompson-tomo: This is a replica of the functionality that was in command writer with nested try catches etc. 

As per https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/exception-handling-statements the finally will be called even on the throw.

mtmk: i meant the catch block is empty it shouldn't be needed, i.e. try-finally is enough, no?

thompson-tomo:
> perhaps we can create a switch for otel enabled connections so that buffers are copied as raw bytes and passed efficiently as e.g. `NatsMemoryOwner` to minimize GC then the serialization can happen anywhere along the stack. keeping serialization close to the wire was done to minimize buffer copies but in this case it seems to be getting in the way.

This is not just about otel but also helps for retry logic which is included by default with jetstream.

src/NATS.Client.Core/NatsConnection.Publish.cs
```diff
@@ -55,18 +51,71 @@ public ValueTask PublishAsync<T>(string subject, T? data, NatsHeaders? headers =
...
+        {
+            throw;
+        }
+        finally
+        {
+            if (payloadBuffer != null)
+            {
+                payloadBuffer.Reset();
+                _pool.Return(payloadBuffer);
```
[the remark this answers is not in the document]
thompson-tomo: Will need to look if line 91 is behaving as desired but sending based on tests is working only reply is failing when coming from jetstream.

src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
mtmk: it's called from here

https://github.com/nats-io/nats.net/blob/55fb6dbfcba3049449d509215c7ed54b1f23f6a8/src/NATS.Client.Core/NatsConnection.cs#L540-L541

thompson-tomo: Ah it is not actually using a PriorityCommandWriter but rather a short-lived commandwriter, I will implement changes tomorrow.

thompson-tomo: Changes have been made

src/NATS.Client.JetStream/Internal/NatsJSFetch.cs
```diff
@@ -192,13 +192,11 @@ internal override async ValueTask WriteReconnectCommandsAsync(CommandWriter comm
             MinPending = _priorityGroup?.MinPending ?? 0,
             MinAckPending = _priorityGroup?.MinAckPending ?? 0,
         };
-
-        await commandWriter.PublishAsync(
```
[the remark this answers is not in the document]
thompson-tomo: changes have been made

src/NATS.Client.Core/NatsConnection.RequestReply.cs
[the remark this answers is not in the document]
thompson-tomo: serializer is not needed with the changes.

src/NATS.Client.Core/Internal/ReplyTask.cs
thompson-tomo: so the copy solution has in fact resulted in an improved throughput with only a small increase in memory allocation hence throughput provides benefits in the change.

src/NATS.Client.Core/Internal/ReplyTask.cs
```diff
@@ -53,7 +49,23 @@ public override void SetResult(string? replyTo, ReadOnlySequence<byte> payload,
...
+            {
+                var payloadData = new byte[payload.Length];
+                payload.CopyTo(payloadData);
+                payloadValue = new ReadOnlySequence<byte>(payloadData);
+            }
+
+            ReadOnlySequence<byte>? headerValue = null;
+            if (headersBuffer != null)
+            {
+                var headerData = new byte[headersBuffer.Value.Length];
```
mtmk, line 63: allocating here defeats the point of managing pooled buffers and increases GC pressure.

src/NATS.Client.Core/NatsConnection.Publish.cs
```diff
@@ -55,18 +51,81 @@ public ValueTask PublishAsync<T>(string subject, T? data, NatsHeaders? headers =
...
+        {
+            if (headers != null)
+            {
+                headers?.SetReadOnly();
+                if (!_pool.TryRent(out headersBuffer))
+                    headersBuffer = new NatsPooledBufferWriter<byte>(_arrayPoolInitialSize);
+            }
+
+            if (!_pool.TryRent(out payloadBuffer!))
+                payloadBuffer = new NatsPooledBufferWriter<byte>(_arrayPoolInitialSize);
```
mtmk, line 76: again if you don't return these to the pool we'd have increased GC and not much point using the pool

mtmk (changes requested):
The buffer issue is fixed by allocating and copying new buffers which defeated the point of pooling buffers hence increasing the allocations.

### release/2.7 branch:

RequestReplyBench
| Method                  | Mean     | Error     | StdDev  | Allocated |
|------------------------ |---------:|----------:|--------:|----------:|
| RequestReplyDirectAsync | 145.0 us |   8.69 us | 0.48 us |   2.61 KB |

PublishSerialBench (increased iter)
| Method       | Iter    | Mean     | Error    | StdDev  | Allocated |
|------------- |-------- |---------:|---------:|--------:|----------:|
| PublishAsync | 1000000 | 356.1 ms | 46.90 ms | 2.57 ms |   2.03 KB |

### this branch:

RequestReplyBench
| Method                  | Mean     | Error     | StdDev  | Allocated |
|------------------------ |---------:|----------:|--------:|----------:|
| RequestReplyDirectAsync | 166.9 us |  73.34 us | 4.02 us |   3.08 KB |

PublishSerialBench (increased iter)
| Method       | Iter    | Mean     | Error    | StdDev  | Allocated |
|------------- |-------- |---------:|---------:|--------:|----------:|
| PublishAsync | 1000000 | 387.6 ms | 15.99 ms | 0.88 ms |   3.34 KB |

a solution i can think of is to have a different path for otel (or similar applications requiring this type of behavior) where we can copy the buffers efficiently from the pool (e.g. using `NatsMemoryOwner`) with minimal allocation impact. or leave where serialization happens as it is and pass the struct around where we can attach more parameters optionally.

haven't checked the other changes.

src/NATS.Client.Core/NatsConnection.Publish.cs
thompson-tomo: not sure what you mean by not returning to the pool especially given that within the finally block i am in fact returning them to the pool in the exact way it was previously being done.

thompson-tomo:
> The buffer issue is fixed by allocating and copying new buffers which defeated the point of pooling buffers hence increasing the allocations.

That was only the case for direct replies and not inbox replies and certainly not for publish. Other cases have been fixed without copying.

In relation to the benchmarks, if the concern is about gc, how come we are not looking at how many GC operations are in fact occurring? Also if performance is that much of a concern, it would be useful to see comparisons with varying qty's so that the growth can be looked at to see if the initial is higher, which skews the data. For the publish

> a solution i can think of is to have a different path for otel (or similar applications requiring this type of behavior) where we can copy the buffers efficiently from the pool (e.g. using NatsMemoryOwner) with minimal allocation impact

I would much rather have a common solution which overall has a net positive impact which considers performance and use cases offered

src/NATS.Client.Core/NatsConnection.Publish.cs
mtmk: ah i see good. i missed that. the increase in allocations is curious though. maybe the object pool size is too low for that bench. need to check that.

mtmk:
release/2.7 branch is obsolete now. need to move this change to base main.

PR #882, opened 2025-06-10
Properly dispose of activity after reply is received in direct mode

### LINKED ISSUE #862: Using `NatsRequestReplyMode.Direct` screws up OpenTelemetry traces
### Observed behavior

Using `NatsRequestReplyMode.Direct` screws up traces.
```cs
var natsClient = new NatsClient(opts: new()
{
	Name = "NATS Client",
	Url = primarySettings!.NatsAddress.ToString(),
	RequestReplyMode = NatsRequestReplyMode.Direct,
});
```

### Expected behavior

My suspicion is that it might be due to the following:
https://github.com/nats-io/nats.net/blob/f674fca8c370497e5fe995afe151d50b0f92999e/src/NATS.Client.Core/NatsConnection.RequestReply.cs#L43-L59

`sub1.Msgs.ReadAllAsync` (which is invoked when `RequestReplyMode = SharedInbox`) disposes activity instances after receiving a response:
https://github.com/nats-io/nats.net/blob/f674fca8c370497e5fe995afe151d50b0f92999e/src/NATS.Client.Core/Internal/ActivityEndingMsgReader.cs#L123-L127

We might have to do something similar in the ``RequestReplyMode = Direct` case?

### Server and client version

Server: v2.11.0
Client: v2.6.0

### Host environment

_No response_

### Steps to reproduce

_No response_

aradalvand:
Closes #862

I haven't tested this yet so that's why it's in draft mode ATM.

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.RequestReply.cs
+++ b/src/NATS.Client.Core/NatsConnection.RequestReply.cs
@@ -45,7 +45,9 @@ public partial class NatsConnection
                     using var rt = _replyTaskFactory.CreateReplyTask(replySerializer, replyOpts.Timeout);
                     requestSerializer ??= Opts.SerializerRegistry.GetSerializer<TRequest>();
                     await PublishAsync(subject, data, headers, rt.Subject, requestSerializer, requestOpts, cancellationToken).ConfigureAwait(false);
-                    return await rt.GetResultAsync(cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


mtmk:
release/2.7 branch is obsolete now. need to move this change to base main.

mtmk:
done in #1047. thanks for your help @aradalvand

PR #900, opened 2025-07-15
Fix test flapper

mtmk:
* Fix race in test
* Added caching for go setup to reduce build times
* Added test solution with only required projects to reduce build times


scottf (approved):
LGTM

PR #907, opened 2025-07-18
fix: `NatsInstrumentationOptions.Default` gets reset each time

aradalvand:
`public static NatsInstrumentationOptions Default => new();` meant that any time `NatsInstrumentationOptions.Default` was accessed it would return a new instance and essentially all settings would get reset, defeating the whole purpose. It's weird this wasn't caught in #859! 😅

Also, made `GetActivityContext` an extension on `NatsHeaders` so that it can also be used in the context of `NatsJSMsg`s and even when you only have a `NatsHeaders` instance in scope.

Found out also that `NatsInstrumentationContext.ActivityContext` didn't need to be made nullabe; in fact, that would've been things confusing.


mtmk (approved):
LGTM thanks @aradalvand

PR #909, opened 2025-07-21
Fix kv ttl interface

mtmk:
KV TTL should only be allowed on Create and Purge


ripienaar (approved):
Doubt I've ever read .net code but this does appear to the do the right thing :)

PR #911, opened 2025-07-22
fix: `NatsJSConsumer` never disposing receive activities

### LINKED ISSUE #910: `NatsJSConsumer` is never disposing receive activities
### Observed behavior

The "receive" span/activity started in `NatsMsg<T>.Build` here:
https://github.com/nats-io/nats.net/blob/a68b68cbc3134f430f127677060f704ab22dd93d/src/NATS.Client.Core/NatsMsg.cs#L385-L399

... is meant to be disposed/stopped when the `NatsMsg<T>` instance is read through a `ChannelReader` — we already have a class purpose-built for this called `ActivityEndingMsgReader`:
https://github.com/nats-io/nats.net/blob/a68b68cbc3134f430f127677060f704ab22dd93d/src/NATS.Client.Core/Internal/ActivityEndingMsgReader.cs#L7-L17

... but this class is only being used in `NatsSub`:
https://github.com/nats-io/nats.net/blob/a68b68cbc3134f430f127677060f704ab22dd93d/src/NATS.Client.Core/NatsSub.cs#L26

... and crucially, NOT in `NatsJSConsume`:
https://github.com/nats-io/nats.net/blob/a68b68cbc3134f430f127677060f704ab22dd93d/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs#L138-L139

As a consequence, "receive" spans that are started for JS messages are never disposed/stopped, and are therefore never flushed — in addition to technically being a memory leak, this also means that these spans never show up in the OpenTelemetry back-end that's ingesting spans — because spans are only sent to the OLTP exporter when they're stopped — which was an issue that had me pulling my hair out in complete astonishment for hours 😆

### Expected behavior

Implied above.

### Server and client version

nats-server: v2.11.0
nats client: 2.7.0-preview.2

### Host environment

_No response_

### Steps to reproduce

_No response_

aradalvand:
Fixes #910

Let me know if you don't think this is the right approach. ~~The sub-optimal change here is that the newly-added `ActivityEndingJSMsgReader` is almost the duplicate of the already-existing `ActivityEndingMsgReader`, but I could not find a way to abstract over the both of them — not to mention they technically reside in different assemblies too (core and jetstream).~~ (Update: Managed to [consolidate](https://github.com/nats-io/nats.net/pull/911#issuecomment-3109020932)).


mtmk:
good catch, thanks @aradalvand 

There are a few other implementations of NatsSubBase. Perhaps we should apply to all of them. We can take it step by step though no need to do all in one PR. wdyt?

aradalvand:
In the [last commit](https://github.com/nats-io/nats.net/commit/38be1046a9b95fec6fd14038bb0050f1447fefd7), I managed to merge `ActivityEndingMsgReader` and `ActivityEndingJSMsgReader` into a single class — in order to do this, I had to declare a shared interface for both `INatsMsg<T>` and `INatsJSMsg<T>` with a `Headers` property (I've currently named it `IMsg` which might not be great). Let me know what you think

mtmk:
thanks @aradalvand looks good to me in general.

KV watchers are using push consumers so they are a little different. We can handle them later if needed.

The constrained reader using the new INatsMsg interface seems reasonable to me.

I know it might be tricky but we have a separate project just to test otle stuff. do you think there is a way we can test these changes there?

mtmk (approved):
LGTM thanks @aradalvand 

could you merge 2.7 which has build fixes and run a 'dotnet format' to turn ci green please? we can merge right after that.

aradalvand:
Done, sir :)

PR #918, opened 2025-07-25
Add RequestTimeout option to NatsJSOpts for request operations

johnweldon:
The previous implementation used the defaults from the connections NatsOpts RequestTimeout for JetStream specific request/reply operations.

This change allows setting a separate default value for JetStream operations from the core NatsOpts default.

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSOpts.cs
+++ b/src/NATS.Client.JetStream/NatsJSOpts.cs
@@ -38,7 +38,7 @@ public record NatsJSOpts
     /// <summary>
     /// Timeout for JetStream API calls.
     /// </summary>
-    public TimeSpan? RequestTimeout { get; init; }
+    public TimeSpan RequestTimeout { get; }
 
     /// <summary>
     /// Ask server for an acknowledgment.
--- a/src/NATS.Client.JetStream/NatsJSOpts.cs
+++ b/src/NATS.Client.JetStream/NatsJSOpts.cs
@@ -8,7 +8,7 @@ namespace NATS.Client.JetStream;
 /// </summary>
 public record NatsJSOpts
 {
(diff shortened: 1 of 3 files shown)


src/NATS.Client.JetStream/NatsJSOpts.cs
```diff
@@ -34,6 +35,11 @@ public NatsJSOpts(NatsOpts opts, string? apiPrefix = default, string? domain = d
     /// </summary>
     public string? Domain { get; }
 
+    /// <summary>
+    /// Timeout for JetStream API calls.
+    /// </summary>
+    public TimeSpan? RequestTimeout { get; init; }
```
mtmk, line 41: Can we make this just 'get' property? that way we won't need to the nullable type on it.
```csharp
public TimeSpan RequestTimeout { get; }
```
also in the context we won't need the `??` checks. unless I'm missing something, of course. wdyt?

mtmk:
sign-off check isn't working because when checked out I can't see the signature 😕 .. even though github seems to be happy saying it's 'verified'. as long as github is happy we're fine. we can ignore the failing action.

src/NATS.Client.JetStream/NatsJSOpts.cs
johnweldon: Wouldn't that change the default behaviour, and possibly break backward compatibility?

johnweldon: Yes, except that in the .ctor it could be null, and is so by default, meaning it'd fall back to the existing behaviour if we keep it nullable.

mtmk:
ah, yes. it was a problem with the sign-off check workflow. thanks for the fix @johnweldon!

src/NATS.Client.JetStream/NatsJSOpts.cs
johnweldon: I understand now... made the change.

PR #921, opened 2025-07-31
chore: rework NatsOpt.Default initialization

### LINKED ISSUE #920: NatsOpts.Default with { } shares references to nested options like AuthOpts
### Observed behavior

When using `var opts = NatsOpts.Default with { };`, the resulting `opts.AuthOpts` still references the same instance as in `NatsOpts.Default`. This causes unexpected side effects when modifying or binding options, as multiple option sets unintentionally share nested state.

### Example
```csharp
var opts1 = NatsOpts.Default with { };
var opts2 = NatsOpts.Default with { };

configuration.Bind(opts2);

// configuration like AuthOpts__Username=other
// -> opts1.AuthOpts.Username is now also "other"

### Expected behavior

Please provide a built-in method (e.g. NatsOpts.Default.Clone()) that returns a deep-copied instance where nested records like AuthOpts, TlsOpts, and WebSocketOpts are independent.

Or maybe even cleaner, do not use singletons within default ctor but only with NatsOpts.Default.

### Server and client version

Client version 2.6.5

### Host environment

_No response_

### Steps to reproduce

_No response_

regnrat:
Creating NatsOpts.Default with sub property default objects:
- WebSocketOpts = NatsWebSocketOpts.Default,
- TlsOpts = NatsTlsOpts.Default,
- AuthOpts = NatsAuthOpts.Default,

Using default constructors for these sub properties in default constructor of NatsOpts.
Closes #920

### DIFF
--- a/tests/NATS.Client.CoreUnit.Tests/OptsTests.cs
+++ b/tests/NATS.Client.CoreUnit.Tests/OptsTests.cs
@@ -1,5 +1,4 @@
 using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.Primitives;
 
 public class OptsUrlTests
 {


mtmk:
thanks @regnrat. Looking good. Are there any cases you can think of this might break an existing app?

mtmk:
> I think you can construct situations where this breaks an application, in theory. But that would require unclean code on the consumers side and wouldn't make any sense.
> 
> But theoretical yes. Shall I change the base?

Yes, i think we should. we've been using minor version updates to signal potentially breaking API changes.

mtmk (approved):
LGTM thanks @regnrat

PR #926, opened 2025-08-12
Fix CommandWriter.cs dispose

mtmk:
This pull request updates the `DisposeAsync` method in `CommandWriter.cs` to ensure that both `_cts` and `_ctsReader` cancellation tokens are properly canceled during disposal, improving resource cleanup and preventing potential hangs.


scottf (approved):
LGTM

PR #928, opened 2025-08-14
Add support for initial connection retry

mtmk:
Introduced `RetryOnInitialConnect` option in `NatsOpts` for retrying on failed initial connection attempts.


scottf (approved):
LGTM

PR #934, opened 2025-08-15
Remove v2.9 testing from GitHub workflows

mtmk:
This pull request updates the CI workflow configuration by removing the NATS Server v2.9.x from the test and performance job matrices. This streamlines the workflows to only run against supported or current server versions.


neilalexander (approved):
LGTM

PR #937, opened 2025-08-26
Fix Object Store and JetStream publish retry logic

### LINKED ISSUE #917: NATS.Client.ObjectStore.NatsObjException: SHA-256 digest mismatch
### Observed behavior

We use NATS ObjectStore as a data ingestion buffer. Lots of devices connect to our API and send some blobs-like data, we store them in NATS ObjectStore and then slowly process it.
We have variety of file sizes, but files are usually quite small, below 100k, but they can do up to 30MB (but I doubt it).

It is absolutely fine for the whole day, but it seems like load is the trigger for errors. During rush hours we observe "NATS.Client.ObjectStore.NatsObjException: SHA-256 digest mismatch". It is around 1% of files (for example: 7 out of 820 within window of ~1 minute).

Bucket is created with:

```csharp
await ObjStoreContext.CreateObjectStoreAsync(
    new NatsObjConfig(BucketName) { Compression = true, MaxAge = RetentionPeriod },
    token);
```

Objects are added using (under load):

```csharp
await Store.PutAsync(ObjectKey, Stream, cancellationToken: cancellationToken);
```

and then read (one after another, no more pressure):

```csharp
_ = await store.GetAsync(ObjectKey, Stream, true, cancellationToken);
```

I did take a look at both: [PutAsync](https://github.com/nats-io/nats.net/blob/29c5fb6de4298e0841127e9a3297733199bd15d3/src/NATS.Client.ObjectStore/NatsObjStore.cs#L193) and [GetAsync](https://github.com/nats-io/nats.net/blob/29c5fb6de4298e0841127e9a3297733199bd15d3/src/NATS.Client.ObjectStore/NatsObjStore.cs#L90) and I have no idea what could possibly go wrong, but also I'm pretty sure there is nothing I could do in my code, could I?

Is there a possibility that in `PutAsync` (the "under pressure side") object is sent but metadata is not updated (like [this does not happen](https://github.com/nats-io/nats.net/blob/29c5fb6de4298e0841127e9a3297733199bd15d3/src/NATS.Client.ObjectStore/NatsObjStore.cs#L277)).

No idea, tbh.

Is hash calculation purely client side?
Can I disable it?
Can compression on bucket mess with it?

### Expected behavior

Hash is either valid or object is not added.

### Server and client version

NATS.Net: 2.6.1
NATS Server: 2.11.6

### Host environment

Environment: Kubernetes (EKS)
CPU: ARM
.NET Runtime: 9.0.7

### Steps to reproduce

The problems is that we have in production under pressure (pushing side is under pressure, the reading side throw the exception but is processing data sequentially so it is not really under pressure).

I could not manage so far to reproduce it locally, but I guess the steps would be:
* upload a lot of files concurrently at different speeds
* cancel uploads (using cancellation token) randomly

mtmk:
JetStream publishing was retrying on receiving no response (different from 503 no responders) usually on a timeout. This was causing object store digest mismatches potentially because of duplicate publishes. it is also possible to publish duplicate messages when timeouts starting to occur.

* Fixes #917

### DIFF
--- a/src/NATS.Client.ObjectStore/NatsObjStore.cs
+++ b/src/NATS.Client.ObjectStore/NatsObjStore.cs
@@ -25,6 +25,8 @@ public class NatsObjStore : INatsObjStore
     private const string NatsRollup = "Nats-Rollup";
     private const string RollupSubject = "sub";
 
+    // Disable retries as we want to fail fast and avoid duplicate chunks
+    private readonly NatsJSPubOpts _natsJSPubOpts = new() { RetryAttempts = 1 };
     private readonly NatsObjContext _objContext;
     private readonly INatsJSStream _stream;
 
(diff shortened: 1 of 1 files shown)


scottf (approved):
LGTM

PR #938, opened 2025-08-27
Don't suppress SystemExceptions in NatsSubBase.ReceiveAsync

desmondgc:
We just experienced an issue where our application stopped processing messages because an OutOfMemoryException was thrown and swallowed in ReceiveAsync. Framework docs explicitly state:

> Because [SystemException](https://learn.microsoft.com/en-us/dotnet/api/system.systemexception?view=net-9.0) serves as the base class of a variety of exception types, your code should not throw a [SystemException](https://learn.microsoft.com/en-us/dotnet/api/system.systemexception?view=net-9.0) exception, nor should it attempt to handle a [SystemException](https://learn.microsoft.com/en-us/dotnet/api/system.systemexception?view=net-9.0) exception unless you intend to re-throw the original exception.

### DIFF
--- a/src/NATS.Client.Core/NatsSubBase.cs
+++ b/src/NATS.Client.Core/NatsSubBase.cs
@@ -301,7 +301,7 @@ public abstract class NatsSubBase
             // interested in the messages anymore. Hence we ignore any messages being
             // fed into the channel and rejected.
         }
-        catch (Exception e)
+        catch (Exception e) when (!(e is SystemException))
         {
             _logger.LogError(NatsLogEvents.Subscription, e, "Error while processing message");
 


mtmk:
Thanks @desmondgc sounds reasonable. I wonder as a side effect we would break any other app and should specifically avoid OOM and SO exceptions only for example?

desmondgc:
> I wonder as a side effect we would break any other app and should specifically avoid OOM and SO exceptions only for example?

I think there are a number of other SystemException subtypes that are equally dangerous to swallow, like AccessViolationException and TypeInitializationException. I can't speak to whether this would have unexpected side-effects, but it would certainly be surprising.

> we have few places we're doing this

Indeed. Removing / updating this pattern everywhere would be a significant change. (Although in some cases the exception is "correctly" rethrown.) In the ReceiveAsync case there's more specific exception handling above, and I have real evidence of a problem, which is why I limited the change to single line. I wasn't bold enough to generalize.

mtmk (approved):
LGTM thanks @desmondgc

PR #939, opened 2025-08-27
Fix JetStream publish retry defaults

mtmk:
Decision to retry JetStream publish requests should be left to the application since it depends on the delivery and durability requirements of their solution.

PS This will go into next minor release 2.7.x with other minor braking changes


scottf (approved):
LGTM

PR #950, opened 2025-09-12
Support counters

mtmk:


mtmk:
implemented in #960 against main

PR #954, opened 2025-09-16
Fix Object Store Rename

### LINKED ISSUE #953: Object Store Updating Metadata with New Name - Failure to Remove Old Subject
### Observed behavior

When calling UpdateMetaAsync, it looks like when the PublishAsync is sent to the server to change the name, it only rolls up for the new subject (the new name), which leaves the old metadata (and name) in the object store. This leads to two entries that both lead to the same object. 

### Expected behavior

When calling UpdateMetaAsync, the old entry should be rolled up and removed, leaving only the new entry with the new name. 

### Server and client version

Server: 2.11.8
.NET Client: 2.6.7

### Host environment

_No response_

### Steps to reproduce

_No response_

mtmk:
This pull request fixes the handling of object metadata renaming in the NATS Object Store by ensuring that when an object's name is changed, the old metadata record is properly purged. It also adds a comprehensive test to verify this behavior.

**Object metadata renaming and cleanup:**

* Updated `UpdateMetaAsync` in `NatsObjStore.cs` to purge the old metadata record if the object's name is changed, preventing stale metadata from lingering after a rename.

**Testing and verification:**

* Added a new test `Rename_object_should_perge_old_named_meta` in `ObjectStoreTest.cs` to verify that renaming an object correctly purges the old metadata, updates the stream as expected, and ensures the object can only be accessed by the new name.

**Related**
* Fixes #953

### DIFF
--- a/src/NATS.Client.ObjectStore/NatsObjStore.cs
+++ b/src/NATS.Client.ObjectStore/NatsObjStore.cs
@@ -332,7 +332,7 @@ public class NatsObjStore : INatsObjStore
         if (key != meta.Name)
         {
             var metaSubject = GetMetaSubject(key);
-            await _stream.PurgeAsync(new StreamPurgeRequest{ Filter = metaSubject }, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


scottf (approved):
LGTM

PR #959, opened 2025-09-22
Fix connection state for consume

### LINKED ISSUE #855: Stuck after max retries exceeded in ConsumeAsync
### Observed behavior

When I set a `MaxReconnectRetry` and stop the nats server while a consumer is iterating over `ConsumeAsync` the `ConsumeAsync` method does neither return or throw after the reconnect loop stopped. I see no possibility to exit it (short of somehow externally registering that stop and then killing that task). 

### Expected behavior

When the reconnect loop exits, there is some way to escape from the ConsumeAsync method (either it returns/stops yielding or throws an exception). In that way  I can notice this in my program and try to handle it.

### Server and client version

client: 2.6.1
server: 2.11 (from official image 2.11-alpine)

### Host environment

Linux, server run via docker

### Steps to reproduce

A reproduction can be seen on https://github.com/NicholasTNG/reconnect.net/tree/trunk 
1. Start nats server via `docker compose up`
2. Start Program (with an integer as first argument to set the `MaxReconnectRetry`
3. Wait a moment and then stop the nats server via `docker compose down`

Now the reconnect loop will close but the program keeps running (doing nothing). The code after the ConsumeAsync call is never reached.

mtmk:
* Resolves #855

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -17,6 +17,7 @@ public enum NatsConnectionState
     Open,
     Connecting,
     Reconnecting,
+    Failed,
 }
 
 internal enum NatsEvent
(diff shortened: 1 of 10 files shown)


scottf (approved):
LGTM

PR #962, opened 2025-09-25
Add Atomic Publish Support

mtmk:
This pull request adds support for the new AllowAtomicPublish property in the JetStream stream configuration as described in [ADR-50](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-50.md), allowing atomic batch publishing into streams. It also includes comprehensive tests to verify the correct behavior of this property when creating, updating, and retrieving streams.

JetStream Stream Configuration Enhancements:

* Added the `AllowAtomicPublish` property to the `StreamConfig` class, enabling atomic batch publishing for streams. This property is serialized to JSON as `allow_atomic` and is ignored when set to its default value.

Testing Improvements:

* Introduced a new test, `AllowAtomicPublish_property_should_be_set_on_stream`, in `ManageStreamTest.cs` to verify that the `AllowAtomicPublish` property is correctly set, updated, and persisted on streams. The test is conditionally skipped for NATS server versions earlier than 2.12.
* Added a missing `using` directive for `NATS.Client.TestUtilities` in the test file to support the new test.


scottf (approved):
LGTM

PR #963, opened 2025-09-25
Add support for message schedules

mtmk:
This pull request introduces support for the new `AllowMsgSchedules` property in the JetStream `StreamConfig`, enabling the scheduling of messages as described in [ADR-51](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-51.md). It also adds a corresponding test to verify correct behavior when creating and updating streams with this property.

JetStream StreamConfig enhancements:

* Added the `AllowMsgSchedules` boolean property to the `StreamConfig` class, allowing message scheduling to be enabled or disabled for streams.

Testing improvements:

* Introduced a new test, `AllowMsgSchedules_property_should_be_set_on_stream`, which verifies that the `AllowMsgSchedules` property is correctly set, persisted, and enforces that disabling it after enabling returns the expected error.
* Added missing import for `NATS.Client.TestUtilities` in `ManageStreamTest.cs` to support the new test.


scottf (approved):
LGTM

PR #965, opened 2025-09-26
Add Prioritized Priority Consumer Policy

mtmk:
This pull request adds support for the "prioritized" priority policy for JetStream consumers as described in [ADR-42](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-42.md#prioritized-policy), enabling message delivery prioritization based on a specified priority value. It updates the consumer configuration, request models, and fetch options to include a `Priority` field used when the "prioritized" policy is set. The changes are validated and tested with new unit tests.

***Breaking change:***

`PriorityPolicy` on consumer config is now implemented as an `enum` an it would break applications using this field as `string` before. Applications not using this field won't be affected.

**Support for prioritized priority policy:**

* Added "prioritized" as a valid value for `PriorityPolicy` in `ConsumerConfig`, and updated validation to accept it. [[1]](diffhunk://#diff-6745059eea806687da349c43cbb33b277414c624ac58aac99401cfa57c58b8beL275-R275) [[2]](diffhunk://#diff-d9f5b99f7a2826ef012daa0f19c51104ee5dbf76f694604d8e1356a7de9ca7f5L272-R274)
* Added a `Priority` property to `ConsumerGetnextRequest` and `NatsJSPriorityGroupOpts`, allowing clients to specify delivery priority (0-9, with 0 as highest). [[1]](diffhunk://#diff-46b1a8315556e0f43a431d4563b714652b1dfae1bdd8f09d97acab601f56c7c5R85-R97) [[2]](diffhunk://#diff-e242e6ed01deda89352c949c84b0aa1095c24326954fcf4b2d82346d59d63ce4R268-R277)
* Updated `NatsJSConsumer` to pass the `Priority` value from fetch options into the consumer request. [[1]](diffhunk://#diff-c13f88be37e7f154bab245a644075c446bc5f4f56d3f9a982d7531c06e421d90R339) [[2]](diffhunk://#diff-c13f88be37e7f154bab245a644075c446bc5f4f56d3f9a982d7531c06e421d90R442) [[3]](diffhunk://#diff-c13f88be37e7f154bab245a644075c446bc5f4f56d3f9a982d7531c06e421d90R454)

**Testing and validation:**

* Added unit tests to verify correct behavior when using the "prioritized" policy and priority values, and to ensure the policy is accepted in consumer creation.

### DIFF
--- a/src/NATS.Client.JetStream/Models/ConsumerConfig.cs
+++ b/src/NATS.Client.JetStream/Models/ConsumerConfig.cs
@@ -272,7 +272,7 @@ public record ConsumerConfig
 #endif
 
     /// <summary>
-    /// Specifies the priority policy for consumer message selection, such as prioritizing <c>overflow</c> or <c>pinned_client</c>.
+    /// Specifies the priority policy for consumer message selection, such as prioritizing <c>prioritized</c>, <c>overflow</c>, or <c>pinned_client</c>.
     /// </summary>
     [System.Text.Json.Serialization.JsonPropertyName("priority_policy")]
     [System.Text.Json.Serialization.JsonIgnore(Condition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingDefault)]
--- a/src/NATS.Client.JetStream/Models/ConsumerConfig.cs
+++ b/src/NATS.Client.JetStream/Models/ConsumerConfig.cs
@@ -272,16 +272,15 @@ public record ConsumerConfig
 #endif
 
     /// <summary>
-    /// Specifies the priority policy for consumer message selection, such as prioritizing <c>prioritized</c>, <c>overflow</c>, or <c>pinned_client</c>.
+    /// Specifies the priority policy for consumer message selection.
     /// </summary>
     [System.Text.Json.Serialization.JsonPropertyName("priority_policy")]
     [System.Text.Json.Serialization.JsonIgnore(Condition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingDefault)]
-    [System.ComponentModel.DataAnnotations.StringLength(int.MaxValue, MinimumLength = 1)]
-    [System.ComponentModel.DataAnnotations.RegularExpression(@"^[^.*>]+$")]
 #if NET6_0
-    public string? PriorityPolicy { get; set; }
+    [System.Text.Json.Serialization.JsonConverter(typeof(NatsJSJsonStringEnumConverter<ConsumerConfigPriorityPolicy>))]
+    public ConsumerConfigPriorityPolicy? PriorityPolicy { get; set; }
 #else
-    public string? PriorityPolicy { get; init; }
+    public ConsumerConfigPriorityPolicy? PriorityPolicy { get; init; }
 #endif
 
     /// <summary>
--- a/src/NATS.Client.JetStream/Models/ConsumerGetnextRequest.cs
+++ b/src/NATS.Client.JetStream/Models/ConsumerGetnextRequest.cs
@@ -82,4 +82,17 @@ public record ConsumerGetnextRequest
     [System.Text.Json.Serialization.JsonPropertyName("id")]
     [System.Text.Json.Serialization.JsonIgnore(Condition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingDefault)]
     public string? Id { get; set; }
+
+    /// <summary>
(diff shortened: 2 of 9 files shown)


src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
```diff
@@ -263,17 +263,6 @@ private async ValueTask<NatsJSConsumer> CreateOrUpdateConsumerInternalAsync(
             throw new NatsJSException("Cannot create consumers with multiple priority groups.");
         }
 
-        if (config.PriorityPolicy is "pinned_client")
-        {
-            throw new NotImplementedException("Pinned clients are not supported yet.");
```
mtmk, line 268: pinned_client wan't implemented anyway, so just removed it here.

src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
```diff
@@ -263,17 +263,6 @@ private async ValueTask<NatsJSConsumer> CreateOrUpdateConsumerInternalAsync(
...
-        }
-
-        // TODO: enum these values?
-        if (config.PriorityPolicy != null && config.PriorityPolicy != "none" && config.PriorityPolicy != "overflow" && config.PriorityPolicy != "pinned_client")
```
mtmk, line 272: don't need this check any more. it's an enum now.

src/NATS.Client.JetStream/Models/ConsumerConfig.cs
```diff
@@ -272,16 +272,15 @@ public ConsumerConfig(string name)
 #endif
 
     /// <summary>
-    /// Specifies the priority policy for consumer message selection, such as prioritizing <c>overflow</c> or <c>pinned_client</c>.
+    /// Specifies the priority policy for consumer message selection.
```
mtmk, line 275: no need for documenting values. it's an enum now. it's a breaking change but since this is a fairly new feature, i don't expect many applications would be affected by it. required recompilation but the fix should be straight forward.

tests/NATS.Client.JetStream.Tests/EnumJsonTests.cs
```diff
@@ -167,4 +167,39 @@ public void ConsumerCreateRequestAction_Test(ConsumerCreateAction value, string
...
+        var json = Encoding.UTF8.GetString(bw.WrittenSpan.ToArray());
+        Assert.Contains(expected, json);
+
+        var result = serializer.Deserialize(new ReadOnlySequence<byte>(bw.WrittenMemory));
+        Assert.NotNull(result);
+        Assert.Equal(value, result.PriorityPolicy);
+    }
+
+    [Fact]
+    public void ConsumerConfigPriorityPolicy_null_test()
```
mtmk, line 276: we also have a test for `null` where it should not be serialized onto the wire at all when `null`.

src/NATS.Client.JetStream/Models/ConsumerConfig.cs
scottf: Yeah, this is fine. I've done the same on brand new things.

src/NATS.Client.JetStream/Models/ConsumerGetnextRequest.cs
```diff
@@ -82,4 +82,17 @@ public record ConsumerGetnextRequest
...
+    /// </summary>
+    /// <remarks>
+    /// Lower values indicate higher priority (0 is the highest priority).
+    /// Maximum priority value is 9. This field is only used when the consumer
+    /// has PriorityPolicy set to "prioritized".
+    /// </remarks>
+    [System.Text.Json.Serialization.JsonPropertyName("priority")]
+    [System.Text.Json.Serialization.JsonIgnore(Condition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingDefault)]
+    [System.ComponentModel.DataAnnotations.Range(0, 9)]
+    public byte Priority { get; set; }
```
scottf, line 97: when these are serialized, are the names lowercased?
probably what this is for
```
 new JsonStringEnumConverter<ConsumerConfigPriorityPolicy (JsonNamingPolicy.SnakeCaseLower),
```

scottf: I assume None does not get written either.

scottf (approved):
LGTM

PR #973, opened 2025-10-02
fix: handle 408 Requests Pending responses for fetch requests

### LINKED ISSUE #967: `FetchNoWaitAsync` non-deterministically blocks
### Observed behavior

This is the minimal repro:
```cs
using NATS.Net;

var nats = new NatsClient();
var jetStream = nats.CreateJetStreamContext();
try
{
    await jetStream.DeleteStreamAsync("ISTGITSATEST");
    Console.WriteLine("DELETED.");
}
catch
{

}

var stream = await jetStream.CreateOrUpdateStreamAsync(new(
    name: "ISTGITSATEST",
    subjects: [$"ISTGITSATEST.>"]
)
{
    MaxMsgsPerSubject = 1,
    Discard = StreamConfigDiscard.Old,
    DuplicateWindow = TimeSpan.Zero,
    Retention = StreamConfigRetention.Workqueue,
    AllowDirect = true,
});

_ = Task.Run(async () =>
{
    await jetStream.PublishAsync("ISTGITSATEST.hi", 123);
    await Task.Delay(500);
});

await Task.Delay(1_000);
await Task.WhenAll(Enumerable.Range(1, 3).Select(i => SpawnConsumer(i)));

async Task SpawnConsumer(int i)
{
    try
    {
        var consumer = await stream.CreateOrUpdateConsumerAsync(new("foo"));
        while (true)
        {
            var enumerable = consumer.FetchNoWaitAsync<int>(new() { MaxMsgs = 100 });
            await foreach (var item in enumerable)
            {
                Console.WriteLine($"({i}) RECEIVED: {item}");
            }
            Console.WriteLine($"({i}) LOOP EXITING");
            await Task.Delay(250);
        }
    }
    catch (Exception ex)
    {
        Console.WriteLine(ex);
    }
}
```
If you `dotnet run` this program, sometimes you get an output like this:
```
(1) RECEIVED: NATS.Client.JetStream.NatsJSMsg`1[System.Int32]
(1) LOOP EXITING
(2) LOOP EXITING
(3) LOOP EXITING
(2) LOOP EXITING
(1) LOOP EXITING
(3) LOOP EXITING
(2) LOOP EXITING
(3) LOOP EXITING
(1) LOOP EXITING
(2) LOOP EXITING
(1) LOOP EXITING
(3) LOOP EXITING
(1) LOOP EXITING
(2) LOOP EXITING
(3) LOOP EXITING
(3) LOOP EXITING
(2) LOOP EXITING
(1) LOOP EXITING
```
The important thing to notice is that you periodically see the `EXHAUSTED` message for all three of `(1)`, `(2)`, and `(3)`. However, if you run the program a couple more times, you occasionally get this type of result, which is very strange:
```
(2) RECEIVED: NATS.Client.JetStream.NatsJSMsg`1[System.Int32]
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
(2) LOOP EXITING
```
Where are the other numbers (other than `(2)`) this time?!
(This doesn't just happen to `(2)`, this is one instance, sometimes two of these numbers appear and one doesn't, etc.).

### Expected behavior

I would expect to see the first kind of output at all times, given the semantics of `FetchNoWaitAsync`, I don't see why its returned enumerator's `MoveNextAsync` would block.

### Server and client version

Client version: `2.7.0-preview.6`
Server version: 2.x

### Host environment

_No response_

### Steps to reproduce

_No response_

aradalvand:
Fixes #967

[This](https://github.com/nats-io/nats-server/blob/7726cba45cca93c8f1f760d50bdf74e20aea5934/server/consumer.go#L4216-L4235) is the relevant piece of source code in the NATS server codebase that's yielding that error.

This error type was added 4 years ago as part of [this PR](https://github.com/nats-io/nats-server/pull/2776), and [specifically here](https://github.com/nats-io/nats-server/commit/5592d923c4ab9d496866a75e5859d1ce26152a47#diff-c6e9a3ce0d55948f2e7ea75a4de309986654f2ab11e0f8750625b88b28bd9ebfR2127-R2129).

Support for this error type was added to the NATS Go client as part of [this PR](https://github.com/nats-io/nats.go/pull/967).

Now with this PR, the .NET client begins to explicitly recognize this error type and replicate the same behavior in the other `408` case in a `NatsJSFetch`, which is ending the subscription (since it [means no messages will be sent for that request](https://github.com/nats-io/nats.go/pull/967#discussion_r863251083)); thereby resolving the defect described in #967.

### DIFF
--- a/src/NATS.Client.Core/NatsHeaderParser.cs
+++ b/src/NATS.Client.Core/NatsHeaderParser.cs
@@ -119,6 +119,11 @@ public class NatsHeaderParser
                 headers.Message = NatsHeaders.Messages.MessageSizeExceedsMaxBytes;
                 headers.MessageText = NatsHeaders.MessageMessageSizeExceedsMaxBytesStr;
             }
+            else if (headerLine.SequenceEqual(NatsHeaders.MessageRequestsPendingBytes))
+            {
+                headers.Message = NatsHeaders.Messages.RequestsPending;
+                headers.MessageText = NatsHeaders.MessageRequestsPendingStr;
+            }
             else
             {
                 headers.Message = NatsHeaders.Messages.Text;
--- a/src/NATS.Client.Core/NatsHeaders.cs
+++ b/src/NATS.Client.Core/NatsHeaders.cs
@@ -27,6 +27,7 @@ public class NatsHeaders : IDictionary<string, StringValues>
         NoMessages,
         RequestTimeout,
         MessageSizeExceedsMaxBytes,
+        RequestsPending,
     }
 
     // Uses C# compiler's optimization for static byte[] data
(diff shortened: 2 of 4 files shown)


mtmk:
thanks @aradalvand!

mtmk (approved):
LGTM thanks @aradalvand

PR #979, opened 2025-10-06
fix: handle 408 Requests Pending responses for fetch requests (#973)

mtmk:
* Handle 408 Requests Pending responses for fetch requests

* Remove redundant `Console.WriteLine` that was put there for testing

(cherry picked from commit d313529caa843d0cedc4e1c49450c0a8701da408)


scottf (approved):
LGTM

PR #981, opened 2025-10-09
Added verification of the consumer sequence number for pull ordered consumers

### LINKED ISSUE #980: An ordered consumer has no internal check for the sequence numbers in ConsumeAsync
### Observed behavior

Submitting it as a defect to draw more attention, but it can equally be just an improvement if the current implementation is justified.
The  [ADR-17](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-17.md) says

> The subscription must track the last good stream and consumer sequences. When a gap is observed, the subscription closes its current subscription, releases its consumer and creates a new one starting at the proper stream sequence

Unlike with Golang client which has explicit verification of the consumer sequence number [here](https://github.com/nats-io/nats.go/blob/4dc7f1e4f613c88eade172f8fe9ebeea4bf28675/jetstream/ordered.go#L129)  Nats.Net just tracks the last stream sequence number and recreates the ephemeral consumer in case of exceptions [NatsJSOrderedConsumer.ConsumeAsync](https://github.com/nats-io/nats.net/blob/05c3fc39f3f97d291aafaee26c68aea337dd1e28/src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs#L78)

Without the explicit check for the consumer sequence on the client side, it can lead to gaps if the server starts misbehaving or the message was lost because of a network issue.  Was there a particular reason to skip this verification logic?  Will be happy to assist with the PR which introduces it if needed

### Expected behavior

Ordered consumer should track both consumer sequence and stream sequence

### Server and client version

nats server: 2.11.6
nats client: 2.6.7

### Host environment

_No response_

### Steps to reproduce

_No response_

sergeimam:
As per  https://github.com/nats-io/nats.net/issues/980 added explicit verification of the consumer sequence for an ordered consumer

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
+++ b/src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
@@ -79,6 +79,7 @@ public class NatsJSOrderedConsumer : INatsJSConsumer
                 consumerName = consumer.Info.Name;
                 _logger.LogInformation(NatsJSLogEvents.NewConsumer, "Created {ConsumerName} with sequence {Seq}", consumerName, seq);
 
+                ulong cseq = 0;
                 NatsJSProtocolException? protocolException = default;
 
                 await using (var cc = await consumer.OrderedConsumeInternalAsync(serializer, opts, cancellationToken))
@@ -126,7 +127,15 @@ public class NatsJSOrderedConsumer : INatsJSConsumer
                             if (msg.Metadata is not { } metadata)
                                 continue;
 
+                            var expected = cseq + 1;
+                            if (metadata.Sequence.Consumer != expected)
+                            {
+                                _logger.LogWarning(NatsJSLogEvents.Retry, $"Consumer sequence mismatch. Expected {expected}, was {metadata.Sequence.Consumer}  Retrying...");
+                                goto CONSUME_LOOP;
+                            }
+
                             seq = metadata.Sequence.Stream;
+                            cseq = metadata.Sequence.Consumer;
 
                             yield return msg;
                         }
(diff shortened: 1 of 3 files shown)


src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
```diff
@@ -180,23 +189,57 @@ public async IAsyncEnumerable<NatsJSMsg<T>> FetchAsync<T>(
         [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
         cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_cancellationToken, cancellationToken).Token;
+        var processed = 0;
```
mtmk, line 192: should we track max-bytes as well here e.g. `var bytesProcessed = 0;` then add it to options below?

src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
```diff
@@ -180,23 +189,57 @@ public async IAsyncEnumerable<NatsJSMsg<T>> FetchAsync<T>(
...
-        }
+            var mismatch = false;
+            ulong cseq = 0;
+
+            var consumer = await RecreateConsumer(_fetchConsumerName, _fetchSeq, cancellationToken);
+            _fetchConsumerName = consumer.Info.Name;
+
+            try
+            {
+                var fetchOpts = opts with { MaxMsgs = opts.MaxMsgs - processed };
```
mtmk: should we add the tracked bytes here? e.g.
```csharp
// Adjust both MaxMsgs and MaxBytes
var fetchOpts = opts with
{
    MaxMsgs = opts.MaxMsgs - processed,
    MaxBytes = opts.MaxBytes > 0 ? opts.MaxBytes - bytesProcessed : 0  // Add this?
};
```

mtmk (commented):
thanks @sergeimam looks good. Just a comment about the concern you raised regarding a potential infinite loop when using max-bytes option. let me know what you think.

src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
sergeimam: @mtmk I pushed a commit with the change. Does this capture your idea correctly? If yes, I will flip the PR into ready status.  
**Edited**: moved to ready for review

mtmk (approved):
LGTM thanks @sergeimam

tests/NATS.Client.CoreUnit.Tests/NuidTests.cs
```diff
@@ -129,15 +129,18 @@ public void GetNextNuid_PrefixRenewed_Char()
...
+            if (Nuid.TryWriteNuid(secondNuid))
+                Interlocked.Increment(ref _result);
         });
 
         executionThread.Start();
         executionThread.Join(1_000);
 
         // Assert
-        Assert.True(result);
+        Assert.Equal(2, Interlocked.CompareExchange(ref _result, 0, 0));
```
mtmk, line 143: unrelated test fix. rarely we get an error here. guessing potential race on the result bool.

src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
```diff
@@ -180,23 +189,65 @@ public async IAsyncEnumerable<NatsJSMsg<T>> FetchAsync<T>(
...
+
+                    processed++;
+                    bytesProcessed += msg.Size;
+
+                    yield return msg;
+                }
+            }
+            finally
+            {
+                var deleted = await TryDeleteConsumer(_fetchConsumerName, cancellationToken);
```
scottf, line 235: Why do you care if the delete succeeds? It should be ephemeral anyway. The key is to always generate an ordered consumer name. We agreed that the user can provide a prefix for ordered consumer names, which will allow them to be found more easily in logs

scottf (approved):
LGTM

src/NATS.Client.JetStream/NatsJSOrderedConsumer.cs
mtmk: `_fetchConsumerName` is kept around to delete the ephemeral when we finish with it just to be a good citizen. so if we couldn't delete it this time round and there is another fetch we try our best to delete again potentially. I don't think prefix feature is implemented in this client.

PR #984, opened 2025-10-17
Fix Ad-Hoc JSON Serializer to Use Default Options

### LINKED ISSUE #974: NatsJsonSerializerContext Default options suggestion
### Proposed change

The `NatsJsonSerializerContext` default options should be the same as `JsonSerializerOptions.Default`.

Right now, the property `DefaultIgnoreCondition` is set to `JsonIgnoreCondition.WhenWritingNull`, most likely for performance reasons. I don't think it's worth it.

This throws when trying to deserialize `required` properties that can be null, ex: `public required string? Name { get; init; }`. It's not a big deal when you know about it, but it's a footgun.

### Use case

I'm sending messages between two C# apps. I'm using the `required` keyword on properties to make sure I don't forget to initialize them. Since I'm using the same class in both apps (code is shared through a submodule), the serialization exception was unexpected.

### Contribution

_No response_

mtmk:
This pull request addresses an issue with the default JSON serialization options in `NatsJsonSerializer<T>`, specifically regarding round-tripping of required nullable properties. The main change is to use the default `JsonSerializerOptions` instead of ignoring null values by default, which fixes serialization/deserialization issues for required nullable properties. Additionally, new unit tests are added to verify the correct behavior, especially for .NET 8 and above.

**Serializer options update:**

* Changed the default constructor of `NatsJsonSerializer<T>` to use `JsonSerializerOptions.Default` (or an empty options instance on .NET 6), instead of setting `DefaultIgnoreCondition = WhenWritingNull`. This ensures that required nullable properties are correctly round-tripped during serialization and deserialization.

**Testing improvements:**

* Added new unit tests in `NatsJsonSerializerTests` (for .NET 8 or greater) to verify that required nullable properties are correctly serialized and deserialized, and that round-tripping works as expected with the updated serializer options.

* Fixes #974

### DIFF
--- a/src/NATS.Client.Serializers.Json/NatsJsonSerializer.cs
+++ b/src/NATS.Client.Serializers.Json/NatsJsonSerializer.cs
@@ -29,7 +29,11 @@ public sealed class NatsJsonSerializer<T> : INatsSerializer<T>
     /// This serializer is not suitable for native AOT deployments since it might rely on reflection
     /// </remarks>
     public NatsJsonSerializer()
-        : this(new JsonSerializerOptions { DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull })
+#if NET6_0
+        : this(new JsonSerializerOptions())
+#else
+        : this(JsonSerializerOptions.Default)
+#endif
     {
     }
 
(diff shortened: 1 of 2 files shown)


scottf (approved):
LGTM

PR #977, opened 2025-10-03
Object store item size fix

darkwatchuk:
ObjectStore ObjectMetaData.Size is currently an int and causes a JSON deserialization crash when pointing to an object store with large files. 

Changing to a ulong fixes this issue.

e.g. someone has uploaded large files to this object store on demo.nats.io
```csharp
   [Fact]
   public async Task LargeDemoTest()
   {
       var cts = new CancellationTokenSource(TimeSpan.FromSeconds(10));
       var cancellationToken = cts.Token;

       await using var nats = new NatsConnection(new NatsOpts { Url = "nats://demo.nats.io" });
       var js = new NatsJSContext(nats);
       var ob = new NatsObjContext(js);

       var objStore = await ob.GetObjectStoreAsync("omarchy-3_0_1", cancellationToken);

       await foreach (var thing in objStore.ListAsync(cancellationToken: cancellationToken))
       {
           Console.WriteLine($"{thing.Name} : {thing.Size}");
       }
   }
```

PR #985, opened 2025-10-17
Release 2.6.11

mtmk:
* fix: handle 408 Requests Pending responses for fetch (#973) (#979)
* Object store item size fix (#977)


MauriceVanVeen (approved):
LGTM

PR #986, opened 2025-10-20
Release 2.7.0-preview.8

mtmk:
* fix: handle 408 Requests Pending responses for fetch requests (#973)
* (merge from main) Object store item size fix (#977)


Jarema (approved):
LGTM

PR #995, opened 2025-11-04
Fix stream config adjustments on update

### LINKED ISSUE #994: Updating a stream with external sources doesn't set the external JS API string as expected
### Observed behavior

TL;DR, the `Domain` property of `StreamSource` is not respected when calling `UpdateStreamAsync`.

The `Domain` property of `StreamSource` can be used to configure an external source (that is, a stream coming from another jet stream domain). The property is just syntactic sugar which the NATS client should convert to a JS API prefix upon sending the config to the NATS server. That is the case when `CreateStreamAsync` is called on the `NatsJsContext`, but not when `UpdateStreamAsync` or `CreateOrUpdateStreamAsync` are called. The effect of that is any external source defined in a `StreamConfig` is lost when updating a stream, as all of them will be considered internal given they have no external JS API prefix configured.

### Expected behavior

`UpdateStreamAsync` and `CreateOrUpdateStreamAsync` should pre-process the given config in the same way `CreateStreamAsync` does.

### Server and client version

Server: 2.9
Client: 2.6.10

### Host environment

_No response_

### Steps to reproduce

_No response_

mtmk:
This PR is a fix for the KV Mirror related features implemented in PR **Added Domain support for stream mirroring and sourcing and KV full support for the same. #631**

This pull request refactors how domain-to-external API conversion is handled for JetStream stream configurations and adds comprehensive tests to ensure correct behavior. The main change is the introduction of a shared method to adjust stream configurations for domain settings, ensuring consistent and correct conversion of `Domain` fields to the `External.Api` property for both mirrors and sources across all stream-related operations. Additionally, new tests verify that this conversion is performed as expected.

**Core refactoring and consistency improvements:**

* Introduced a new private static method `AdjustStreamConfigForDomain` in `NatsJSContext.Streams.cs` to handle conversion of `Domain` fields to `External.Api` for both mirrors and sources, replacing duplicated logic in `CreateStreamAsync`, `UpdateStreamAsync`, and `CreateOrUpdateStreamAsync`. All three methods now call this helper for consistent behavior. [[1]](diffhunk://#diff-0ad6d8479defb75e034d860f997edc726b09c301762afdaf1e28cfeee404cab0L22-R22) [[2]](diffhunk://#diff-0ad6d8479defb75e034d860f997edc726b09c301762afdaf1e28cfeee404cab0L74-R43) [[3]](diffhunk://#diff-0ad6d8479defb75e034d860f997edc726b09c301762afdaf1e28cfeee404cab0L197-R166) [[4]](diffhunk://#diff-0ad6d8479defb75e034d860f997edc726b09c301762afdaf1e28cfeee404cab0R239-R276)
* Updated the `ConvertDomain` method to throw an exception if both `Domain` and an existing `External.Deliver` are set, ensuring correct configuration and preventing invalid states.

**Testing improvements:**

* Added a new test class `StreamDomainAdjustmentTest` that verifies domain-to-external API conversion for stream creation, update, and create-or-update operations, both with and without domains set on sources. These tests check both the outgoing payload and the returned configuration, ensuring the conversion is performed and that the `Domain` property is not serialized in the payload.

* fixes #994

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSContext.Streams.cs
+++ b/src/NATS.Client.JetStream/NatsJSContext.Streams.cs
@@ -19,6 +19,17 @@ public partial class NatsJSContext
         StreamConfig config,
         CancellationToken cancellationToken = default)
     {
+        config = AdjustStreamConfigForDomain(config);
+
+        var response = await JSRequestResponseAsync<StreamConfig, StreamInfo>(
+            subject: $"{Opts.Prefix}.STREAM.CREATE.{config.Name}",
+            config,
+            cancellationToken);
+        return new NatsJSStream(this, response);
+    }
+
+    private static StreamConfig AdjustStreamConfigForDomain(StreamConfig config)
+    {
         ThrowIfInvalidStreamName(config.Name, nameof(config.Name));
 
         // keep caller's config intact.
@@ -52,11 +63,7 @@ public partial class NatsJSContext
             config.Sources = sources;
         }
 
-        var response = await JSRequestResponseAsync<StreamConfig, StreamInfo>(
-            subject: $"{Opts.Prefix}.STREAM.CREATE.{config.Name}",
-            config,
-            cancellationToken);
-        return new NatsJSStream(this, response);
+        return config;
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


scottf (approved):
LGTM

PR #997, opened 2025-11-10
Fix NatsUri not reflecting host/port changes from OnConnectingAsync hook

### LINKED ISSUE #996: Wrong NATS URI used for logging (and exceptions) when using address from OnConnectingAsync hook
### Observed behavior

When using the `OnConncetAsync` hook, exceptions and logs show the original URI (from `NatsOpts`) because `ToString()` returns the cached `NatsUri` `_redacted` value from before the modification.

This is especially confusing if connecting fails, because logs and error messages e.g. says the default `localhost:4222` was used, and leads you to thinking that the address from the hook was not applied/used. 

https://github.com/nats-io/nats.net/blob/0d1194bd26ba9a2c077afdd19a46f8320c54ca00/src/NATS.Client.Core/NatsConnection.cs#L455
* The `with` expression creates a new `NatsUri` record with only the `Uri` property changed.
* `_redacted` field in NatsUri is only set in the constructor.
* Using `with`, the _redacted field is shallow copied from the original instance.
*  `ToString()` returns the original URI string, not the modified one.

### Expected behavior

When `OnConnectingAsync` changes the host/port, exceptions and logs should show the modified URI that was actually used for connection.

### Server and client version

**NATS Server:** 2.11
**NATS.NET:**  2.6.6

### Host environment

N/A

### Steps to reproduce

1. Set the `OnConnectingAsync` hook to return a host+port  for which a connection will fail (but different from `localhost:4222` or your `NatsOpts` value).
2. Observe URI used for log statements in `ConnectSocketAsync`.
3. Look at URI used in exception thrown when connection fails.

tormoder:
Fixes https://github.com/nats-io/nats.net/issues/996.


mtmk:
Thanks @tormoder good catch! this is fixing some of the logging but exceptions are still coming through with the original URI since the fix is done inside `ConnectSocketAsync(uri)` call and try/catch blocks wrapping it are still using the old uri. ~~Maybe it can return the uri?~~ **edit:** that won't work if it's throwing

src/NATS.Client.Core/NatsConnection.cs
mtmk: this one is still reporting the original uris

mtmk: this is still reporting the original

mtmk, line 482: this is fine

mtmk: maybe this could return the uri (actually that won't work with exception!)

src/NATS.Client.Core/NatsConnection.cs
```diff
@@ -452,7 +452,8 @@ private async Task ConnectSocketAsync(NatsUri uri)
             target = await OnConnectingAsync(target).ConfigureAwait(false);
             if (target.Host != uri.Host || target.Port != uri.Port)
             {
-                uri = uri with { Uri = new UriBuilder(uri.Uri) { Host = target.Host, Port = target.Port, }.Uri };
+                var modifiedUri = new UriBuilder(uri.Uri) { Host = target.Host, Port = target.Port }.Uri;
+                uri = new NatsUri(modifiedUri.ToString(), uri.IsSeed, uri.Uri.Scheme);
             }
         }
```
mtmk, line 479: maybe this can be extracted into a method?

mtmk (commented):
thanks @tormoder, I saw a couple of places exceptions were still reporting the original uri

tormoder:
Thanks, I will take another look.

tormoder:
A bit fiddly, but did another attempt.

mtmk:
> A bit fiddly, but did another attempt.

thanks @tormoder yes it looks a bit fiddly :) it seems to work though. I will have a closer look once the other PR is merged that fixes the tests as well.

mtmk:
hey @tormoder it looks like the problem was the NatsUri redaction caching. it's never on a hot path so just removed the caching of that string. That seems to be fixing the issue, let me know it works for you too.

tormoder:
> hey @tormoder it looks like the problem was the NatsUri redaction caching. it's never on a hot path so just removed the caching of that string. That seems to be fixing the issue, let me know it works for you too.

Yes, thank you.

scottf (approved):
LGTM

PR #999, opened 2025-11-17
Handel Unobserved Exceptions During Connection State Transitions

### LINKED ISSUE #998: Unobserved task exception when attempting to connect to not existing server
### Observed behavior

In below console app an Unobserved task exception is observed although the connect attempt is in a try/catch block.

### Expected behavior

All exceptions can be handled gracefully.

### Server and client version

Version 2.6.11

### Host environment

_No response_

### Steps to reproduce

```csharp
using NATS.Client.Core;

TaskScheduler.UnobservedTaskException += (sender, eventArgs) =>
{
    Console.WriteLine($"Unobserved task exception: {eventArgs.Exception}");
    eventArgs.SetObserved();
};

var connection
    = new NatsConnection(
        NatsOpts.Default with { Url = "nats:/user@pass:sorry" });

await using (connection.ConfigureAwait(false))
{
    try
    {
        await connection.ConnectAsync().ConfigureAwait(false);
        Console.WriteLine("Connected to NATS server.");
    }
    catch (Exception ex)
    {
        Console.WriteLine($"Failed to connect to NATS server: {ex.Message}");
    }

    await Task.Delay(500).ConfigureAwait(false);

    GC.Collect();

    Console.WriteLine("Exiting application.");
}
```

mtmk:
Ensure unobserved exceptions are handled properly during connection state transitions.

* Fixes #998 

This pull request introduces a small improvement to exception handling during the initial connection process in `NatsConnection.cs`. The change ensures that unobserved exceptions from the connection task are suppressed, preventing potential issues with unhandled exceptions surfacing unexpectedly.

Exception handling improvement:

* In `InitialConnectAsync`, after attempting to set an exception on `_waitForOpenConnection`, the code now explicitly accesses the task's `Exception` property to suppress unobserved exceptions, ensuring cleaner error handling during connection retries. [[1]](diffhunk://#diff-aaf322e0ed00d67e3106e54d66e4f71704eeefcb79350aa24d760251a82aad20L403-R411) [[2]](diffhunk://#diff-aaf322e0ed00d67e3106e54d66e4f71704eeefcb79350aa24d760251a82aad20L423-R439)

### DIFF
--- a/tests/NATS.Client.TestUtilities/FluentAssertionsExtension.cs
+++ b/tests/NATS.Client.TestUtilities/FluentAssertionsExtension.cs
@@ -35,24 +35,4 @@ public static class FluentAssertionsExtension
     {
         source.ToArray().Should().Equal(elements);
     }
-
-    public static GenericCollectionAssertions<T> Should<T>(this Memory<T> source)
-    {
-        return source.ToArray().Should();
-    }
-
-    public static GenericCollectionAssertions<T> Should<T>(this ReadOnlyMemory<T> source)
-    {
-        return source.ToArray().Should();
-    }
-
-    public static GenericCollectionAssertions<T> Should<T>(this Span<T> source)
-    {
-        return source.ToArray().Should();
-    }
-
-    public static GenericCollectionAssertions<T> Should<T>(this ReadOnlySpan<T> source)
-    {
-        return source.ToArray().Should();
-    }
 }
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,44 +1,44 @@
-﻿name: Update Docs
-
-on:
-  push:
-    branches:
-      - main
-
-permissions:
-  contents: write
-  pages: write
-  id-token: write
-
-jobs:
-  update-docs:
-    name: docfx
-    runs-on: ubuntu-latest
-    # if: contains(github.event.head_commit.message, '[nats:update-docs]')
-    environment:
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Setup dotnet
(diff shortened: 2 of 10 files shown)


scottf (approved):
LGTM

tests/NATS.Client.TestUtilities/FluentAssertionsExtension.cs
```diff
@@ -35,24 +35,4 @@ public static void ShouldEqual<T>(this ReadOnlyMemory<T> source, params T[] elem
...
-    {
-        return source.ToArray().Should();
-    }
-
-    public static GenericCollectionAssertions<T> Should<T>(this Span<T> source)
-    {
-        return source.ToArray().Should();
-    }
-
-    public static GenericCollectionAssertions<T> Should<T>(this ReadOnlySpan<T> source)
```
mtmk, line 54: This overload was causing stack overflow in tests. and since it wasn't really used by anything I removed them to fix the issue.

The reason for this starting to happen was the release of C# 14 last week where the [span overload resolution changed](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/csharp-overload-resolution) and since we're using the latest language by default and github ubuntu-lates now includes .net 10, and in our global.json we roll to latest major, we end up using the newest stable compiler.

mtmk:
hi @DieterSedlmayer are you able to verify this fixes the issue for you? thanks

DieterSedlmayer:
LGTM. Waiting for new NuGet release for re-testing.
Btw. Is it the expected behavior that, when using the defaults options with RetryOnInitialConnect = true, the connect attempt never times out?

mtmk:
> LGTM. Waiting for new NuGet release for re-testing.

thanks, will merge.

> Btw. Is it the expected behavior that, when using the defaults options with RetryOnInitialConnect = true, the connect attempt never times out?

Good point. we can add another option for that. wdyt?

PR #1004, opened 2025-11-21
Update INatsJSConsumer to return INatsJSMsg

### LINKED ISSUE #1002: INatsJsConsumer should return INatsJsMsg, not its concrete implementation
### Observed behavior

As per the title, `INatsJsConsumer` should return `INatsJsMsg` when any consumption method is called. Returning the concrete implementation makes it impossible to mock a consumer, in tests.

### Expected behavior

See above

### Server and client version

Server: N/A
Client: 2.6.10

### Host environment

_No response_

### Steps to reproduce

_No response_

mtmk:
Resolves #1002


scottf (approved):
LGTM

PR #1005, opened 2025-11-23
fix: ensure NatsConnectionPool cannot overflow

MiloszKrajewski:
Call me paranoid but assuming someone can try to send 5000 messages per second (which is not insane) by taking connection from pool every time (which is not insane either) it gives service 5 days before crash (2**31/5000/60/60/24 = 4.97).

This is just about ensuring division returns positive number.


mtmk:
thanks for the fix @MiloszKrajewski it sounds like a long shot but nevertheless it was a bug.

Can you also sign the commit please? CNCF policy we must have all commits signed in a PR.

MiloszKrajewski:
> Can you also sign the commit please? CNCF policy we must have all commits signed in a PR.

I think it should be fine now.

mtmk (approved):
LGTM thanks @MiloszKrajewski

PR #1006, opened 2025-11-26
Fix keyed NATS clients with configurations

### LINKED ISSUE #1003: IOptions<NatsOptsBuilder>> overwritten on multiple injections
### Observed behavior

**Background**
I am using NATS .NET in my application to communicate to two separate systems. 
The problem appeared when I injected NATS a second time in the dependency framework using `.AddNatsClient`, even though I was using the `.WithKey` method on the `NatsBuilder` used to configure the setting.
To be more specific, the URL and authentication method for these two NATS instances are different.

The problem itself was that the first injected NATS group, more specific the `NatsConnection` service, with key "FirstNATS", that was using the authentication from the second group.

Once I started debugging the application, it seemed like on line 190, from class `NatsBuilder`, when fetching the `NatsOptsBuilder` via the `IServiceProvider`, the `GetRequiredService` method is used, which, if there are multiple instances registered, will always return the last registered instance. Since the last registered instance was the `NatsOptsBuilder` from the second group, in the first group, the wrong configuration will be used.

The way we are injecting NATS is this:
 ```
        services.AddNatsClient(builder => builder.ConfigureOptions(
            optionsBuilder =>
                optionsBuilder.Configure<IOptions<NATSConfigOptions>>(
                    (builder, options) =>
                    {
                        NATSConfigOptions natsConfigOptions = options.Value;
                        NatsOptsBuilder natsOptsBuilder = builder;
                        NatsOpts opts = builder.Opts with { Url = natsConfigOptions.GetNatsUrl() };
                        opts.AuthOpts = NatsAuthOpts.Default with
                        {
                            AuthCredCallback = natsConfigOptions.NatsAuthCredCallback
                        };
                        natsOptsBuilder.Opts = opts;
                    })));
```

Moreover, in the `PoolFactory` method, the `NatsBuilder` class, accepts a `diKey` parameter, but it does not use it for anything. I would think its purpose was exactly to fetch the right `NatsOptsBuilder` service.

### Expected behavior

When injecting NATS twice, each injection should use the config resulted from it, and not from another injection round.

### Server and client version

NATS server version: 2.12.2
NATS .NET client: 2.6.7

### Host environment

_No response_

### Steps to reproduce

Based on the provided text, here are the reproduction steps for the described bug in the NATS .NET client library's dependency injection setup.

### Reproduction Steps

The core issue stems from multiple registrations of `NatsOptsBuilder` and the internal use of `GetRequiredService` without considering the dependency injection key (`diKey`).

#### 1. Setup Distinct Configurations

Define two distinct configurations (e.g., using `IOptions<NATSConfigOptions>` as referenced in the text), ensuring they have different URLs and/or authentication methods.

*   **Config A:** For the first NATS connection (e.g., URL: `nats://first:4222`, Auth: `Callback A`).
*   **Config B:** For the second NATS connection (e.g., URL: `nats://second:4222`, Auth: `Callback B`).

#### 2. Inject the First NATS Client (The Affected Instance)

Using the dependency injection framework, register the first NATS client. It is crucial to use the `.WithKey` method and the provided configuration structure:

```csharp
// Use Config A settings
services.AddNatsClient(builder => builder
    .WithKey("FirstNATS") // Explicitly setting a key
    .ConfigureOptions(
        optionsBuilder =>
            optionsBuilder.Configure<IOptions<NATSConfigOptions>>(
                (builder, options) =>
                {
                    // Logic to set NatsOptsBuilder.Opts with Config A's URL and AuthCredCallback
                    // ...
                })));
```

#### 3. Inject the Second NATS Client (The Overriding Instance)

Register the second NATS client *after* the first one. This step registers a new `NatsOptsBuilder` instance, which will become the "last registered instance" for that service type.

```csharp
// Use Config B settings
services.AddNatsClient(builder => builder
    .WithKey("SecondNATS") // Explicitly setting a different key
    .ConfigureOptions(
        optionsBuilder =>
            optionsBuilder.Configure<IOptions<NATSConfigOptions>>(
                (builder, options) =>
                {
                    // Logic to set NatsOptsBuilder.Opts with Config B's URL and AuthCredCallback
                    // ...
                })));
```

#### 4. Resolve the First NATS Service

Resolve the `NatsConnection` (or related service from the NATS group) that was intended for the first configuration, using its registered key:

```csharp
// Resolve the NatsConnection for "FirstNATS"
var firstNatsConnection = serviceProvider.GetRequiredKeyedService<INatsConnection>("FirstNATS"); 
// OR resolve the service that internally relies on the NatsBuilder.PoolFactory
```

#### 5. Observe the Configuration

Inspect the internal configuration (specifically the connection options, URL, and authentication details) of the resolved `NatsConnection` service for **"FirstNATS"**.

mtmk:
Refactored `NatsBuilder` to handle multiple keyed options, ensuring independent configurations for each. Added tests to validate correct behavior with keyed clients and enforce method call order for configuration setup. Fixes #1003.

### DIFF
--- a/src/NATS.Extensions.Microsoft.DependencyInjection/NatsBuilder.cs
+++ b/src/NATS.Extensions.Microsoft.DependencyInjection/NatsBuilder.cs
@@ -20,29 +20,31 @@ public class NatsOptsBuilder
 public class NatsBuilder
 {
     private readonly IServiceCollection _services;
-    private readonly OptionsBuilder<NatsOptsBuilder> _builder;
+    private OptionsBuilder<NatsOptsBuilder>? _builder;
     private Func<IServiceProvider, int> _poolSizeConfigurer = _ => 1;
     private object? _diKey = null;
+    private string _optionsName = Options.DefaultName;
 
     public NatsBuilder(IServiceCollection services)
     {
         _services = services;
-        _builder = _services
-            .AddOptions<NatsOptsBuilder>()
-            .Configure<IServiceProvider>((opts, provider) =>
-            {
-                opts.Opts = opts.Opts with
-                {
-                    LoggerFactory = provider.GetService<ILoggerFactory>() ?? NullLoggerFactory.Instance,
-                    SubPendingChannelFullMode = BoundedChannelFullMode.Wait,
-                };
-
-                if (ReferenceEquals(opts.Opts.SerializerRegistry, NatsOpts.Default.SerializerRegistry))
-                {
-                    opts.Opts = opts.Opts with { SerializerRegistry = NatsClientDefaultSerializerRegistry.Default, };
-                }
-            });
     }
+
+    private OptionsBuilder<NatsOptsBuilder> Builder => _builder ??= _services
+        .AddOptions<NatsOptsBuilder>(_optionsName)
+        .Configure<IServiceProvider>((opts, provider) =>
+        {
+            opts.Opts = opts.Opts with
+            {
+                LoggerFactory = provider.GetService<ILoggerFactory>() ?? NullLoggerFactory.Instance,
+                SubPendingChannelFullMode = BoundedChannelFullMode.Wait,
+            };
+
+            if (ReferenceEquals(opts.Opts.SerializerRegistry, NatsOpts.Default.SerializerRegistry))
+            {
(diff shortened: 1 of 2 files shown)


mtmk:
@andreyeurope apologies for the delay. Does this PR address the issue for you as well?

andreyeurope:
Yeah, looks good. Thank you for being so responsive.

scottf (approved):
LGTM

PR #1008, opened 2025-12-01
Add test to promote mirrored to regular stream

mtmk:
This pull request adds a new test to ensure that a mirrored stream can be promoted to a regular stream by removing its mirror configuration, and that messages are retained after the promotion.

JetStream mirror stream promotion:

* Added `Remove_mirror_config_to_promote_stream` test to verify that removing the mirror configuration from a stream successfully promotes it to a regular stream and retains all existing messages.


scottf (approved):
LGTM

PR #1009, opened 2025-12-02
Replace deprecated IndexRange package with Microsoft.Bcl.Memory to solve possible dependency conflicts (for netFramework/netStandard2.0)

### LINKED ISSUE #1010: Package IndexRange for netStandard2.0 projects can cause conflicts with Microsoft.Bcl.Memory
### Observed behavior

Dependency conflict between `IndexRange`and `Microsoft.Bcl.Memory` can cause duplicate `System.Index` definitions when using NATS from netframework/netstandard2.0. (Same issue for `Range`)
 
* `Nats.Client.Core` has a dependency to `IndexRange` in version 1.0.3
*  The project using NATS has a (transitive) dependency to `Microsoft.Bcl.Memory`
*  Both packages provide the C#8 Range and Index-Syntax
*  This will give compile time erros, as both packages define the sames types.

```
# Error Example
path-to-file-with-error.cs(420,69): error CS8356: Predefined type 'System.Index' is declared in multiple referenced assemblies:
'IndexRange, Version=1.0.3.0, Culture=neutral, PublicKeyToken=35e6a3c4212514c6' and
'Microsoft.Bcl.Memory, Version=9.0.0.8, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
```

### Expected behavior

NATS does not include the `IndexRange`-Packge.

I recommend removing this package entirely, as it can be replaced by the Microsoft one.
This is the suggested cause to action by the maintainer.
See [issue 14 in the IndexRange Repo](https://github.com/bgrainger/IndexRange/issues/14) and the Repo-Readme

> [IndexRange Repo readme](https://github.com/bgrainger/IndexRange/blob/master/README.md#indexrange)
For projects that target netstandard2.0 or .NET Framework 4.6.2 or later, use the [Microsoft.Bcl.Memory](https://www.nuget.org/packages/Microsoft.Bcl.Memory/) package instead. This package should be considered deprecated for those target frameworks.

I created the draft #1009 to show the code-impact of these changes. It's just a 1:1 repalcement of the packages.

### Server and client version

It's a compile time issue with the current released package: 2.6.11 

### Host environment

_No response_

### Steps to reproduce

Use a netFramework project that references NATS (eg. `Nats.Client.Core`) and any package that includes `Microsoft.Bcl.Memory`

FelixDamrau:
This solves a dependency conflict between `IndexRange` and `Microsoft.Bcl.Memory` that can cause duplicate `System.Index` definitions when using NATS from netframework/netstandard2.0.

### DIFF
--- a/src/NATS.Client.Core/NATS.Client.Core.csproj
+++ b/src/NATS.Client.Core/NATS.Client.Core.csproj
@@ -32,8 +32,8 @@
     <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
         <PackageReference Include="System.Memory" Version="4.5.5"/>
         <PackageReference Include="System.Buffers" Version="4.5.1"/>
-        <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0"/>
-        <PackageReference Include="Microsoft.Bcl.Memory" Version="9.0.11"/>
+        <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="9.0.0"/>
+        <PackageReference Include="Microsoft.Bcl.Memory" Version="9.0.0"/>
         <!--  Nerdbank.Streams for SequenceReader support https://github.com/dotnet/standard/issues/1493  -->
         <PackageReference Include="Nerdbank.Streams" Version="2.10.72"/>
     </ItemGroup>
--- a/src/NATS.Client.Core/NATS.Client.Core.csproj
+++ b/src/NATS.Client.Core/NATS.Client.Core.csproj
@@ -32,7 +32,7 @@
     <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
         <PackageReference Include="System.Memory" Version="4.5.5"/>
         <PackageReference Include="System.Buffers" Version="4.5.1"/>
-        <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="9.0.0"/>
+        <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0"/>
         <PackageReference Include="Microsoft.Bcl.Memory" Version="9.0.0"/>
         <!--  Nerdbank.Streams for SequenceReader support https://github.com/dotnet/standard/issues/1493  -->
         <PackageReference Include="Nerdbank.Streams" Version="2.10.72"/>
--- a/src/NATS.Client.Core/NATS.Client.Core.csproj
+++ b/src/NATS.Client.Core/NATS.Client.Core.csproj
@@ -33,9 +33,9 @@
         <PackageReference Include="System.Memory" Version="4.5.5"/>
         <PackageReference Include="System.Buffers" Version="4.5.1"/>
         <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0"/>
+        <PackageReference Include="Microsoft.Bcl.Memory" Version="9.0.11"/>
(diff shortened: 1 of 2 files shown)


FelixDamrau:
This PR is my suggestion to solve #1010

src/NATS.Client.Core/NATS.Client.Core.csproj
```diff
@@ -32,10 +32,10 @@
     <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
         <PackageReference Include="System.Memory" Version="4.5.5"/>
         <PackageReference Include="System.Buffers" Version="4.5.1"/>
-        <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0"/>
+        <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0"/>
+        <PackageReference Include="Microsoft.Bcl.Memory" Version="9.0.11"/>
```
FelixDamrau: **info**

The [current package `10.0.0`](https://www.nuget.org/packages/Microsoft.Bcl.Memory/10.0.0#dependencies-body-tab) needs `System.Memory` in version >= 4.6.3
I do not think these package-upgrades fit into this draft as a solution to #1010 

[`9.0.11`](https://www.nuget.org/packages/Microsoft.Bcl.Memory/9.0.11#dependencies-body-tab) requires `System.Memory` in version >= 4.5.5 which fits perfectly.

FelixDamrau:
**issue**

> Error: Commit 859fbbec15276dc645549f280732c79a4b6a575c is not signed using GPG, SSH, or S/MIME
Commit was not signed - my bad - will fix this later today.

---edit---
:heavy_check_mark: 
force pushed the signed commit.

mtmk:
thanks for the PR @FelixDamrau looking at the failing tests I think a couple of them are fixed as flappers in main as I mentioned in the issue #1010. the Request_reply_many_test_idle_timeout one I think I see rarely as well and consider it a flapper. It certainly doesn't look like your change might've cause these since the failures are happening on net8.0.

src/NATS.Client.Core/NATS.Client.Core.csproj
mtmk: To keep it all as consistent as possible, should we keep `Microsoft.Bcl.Memory` same version as `Microsoft.Bcl.AsyncInterfaces` for now if it's available?

edit: It is also my understanding of [NuGet dependency resolution](https://learn.microsoft.com/en-us/nuget/concepts/dependency-resolution?referrer=grok.com#dependency-resolution-rules) is that keeping it as low as possible would help avoiding transient resolution issues in general.

FelixDamrau: I think you are suggesting to increase the `Microsoft.Bcl.AsyncInterfaces"` version. I can do that.

9.0.0 is the lowest version available for both packages.

If you meant to lower the version of `Microsoft.Bcl.Memory`, then that's is not possible. There is no version < 9.

mtmk: I meant if we could lower Microsoft.Bcl.Memory version really. I shouldn't have mentioned Microsoft.Bcl.AsyncInterfaces apologies for the confusion.

It looks good now, if Microsoft.Bcl.AsyncInterfaces can be made 8.0.0 back again we can merge it. I'll try to get a quick release out with that.

mtmk (approved):
LGTM after Microsoft.Bcl.AsyncInterfaces version restored.

Thanks @FelixDamrau 👍

PR #1014, opened 2025-12-11
Add Timeout to Object Get Disposal

### LINKED ISSUE #1013: ObjectStore.GetAsync stuck for hours
### Observed behavior

ObjectStore.GetAsync(...) is stuck for hours.

Of course, I can use `CancellationTokenSource` with `.CancelAfter(...)`, which I'm going to do, but the fact I allowed it to run indefinitely actually allowed me to find this problem.

You maybe remember from #917 that we have a spike of traffic at 5AM in the morning. We use NATS JetStream and NATS ObjectStore to relieve the pressure.

Files get uploaded on API (high concurrency), put into ObjectStore bucket and message is sent that object/file is ready to be processed. Previous problem was related to files being corrupted because of a little bug how shared MemoryOwner was used.

Recently, after updating NATS server to 2.12.2, we've noticed new problem.

Current solution could be explainded like this:

```csharp
jetStream.Consume("log-received", m => {
   var logFile = new MemoryStream();
   objectStore.GetAsync(m.ObjectKey, logFile);
   var logEntries = Parse(logFile);
   database.Put(logEntries);
});
```

There is nothing fancy about it. Ok, maybe there is one things: we are able to keep it going for that long: we use `.AckProgress(...)` every 15 sec (or so) to tell NATS that we are still processing the message.

We noticed that all NATS communication (unrelated to this 'log-received' consumer) starts to behave very weird: periods of terrible performance (0 msg/s) and then catching up is small burst windows (20k msg/s).

Please note, we do measure handling of those messages and we are mostly sub 1 ms, while log processing is around 500 ms.

They are relatively fast to process, they are just not delivered fast enough. We see nothing in application code which would suggest our processing is slow.

After spotting such misbehaving services we decided to give them a little nudge, i mean, restart.

(Below: 3 services being gracefuly restarted, so service-wide cancellation token cancels all handlers)
```
2025-12-08 11:10:20.154 (ERROR) Failed to handle message DeviceLogUploadCompletedEvent after 21690854.0ms
2025-12-08 11:12:45.451 (ERROR) Failed to handle message DeviceLogUploadCompletedEvent after 21886341.6ms
2025-12-08 11:14:34.203 (ERROR) Failed to handle message DeviceLogUploadCompletedEvent after 21954748.5ms
```

They all have been sitting on one message for ~6h !
(Currently, our record holder is 33h)

All stack traces (when we finally cancel them) look like this:
```
System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at NATS.Client.Core.Internal.ActivityEndingMsgReader`1.ReadAllAsync(CancellationToken cancellationToken)+MoveNext()
   at NATS.Client.Core.Internal.ActivityEndingMsgReader`1.ReadAllAsync(CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at NATS.Client.JetStream.NatsJSContext.TryJSRequestAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.TryJSRequestAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.TryJSRequestAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.JSRequestAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.JSRequestResponseAsync[TRequest,TResponse](String subject, TRequest request, CancellationToken cancellationToken)
   at NATS.Client.JetStream.NatsJSContext.DeleteConsumerAsync(String stream, String consumer, CancellationToken cancellationToken)
   at NATS.Client.JetStream.Internal.NatsJSOrderedPushConsumer`1.DisposeAsync()
   at NATS.Client.ObjectStore.NatsObjStore.GetAsync(String key, Stream stream, Boolean leaveOpen, CancellationToken cancellationToken)
   at DeviceLogUpload.DeviceLogUploadCompletedHandler.Handle(DeviceLogUploadCompletedEvent notification, CancellationToken cancellationToken) 
```

What is possibly interesting: the problem is in `NatsJSOrderedPushConsumer.DisposeAsync()`. So it seems all the data has been already received, and then it sits for 33h in Dispose?

NATS logs indicated that our service gets declared as slow consumer, all messages we do process are fast (~1ms, often below) so our best guess for now is this PushConsumer which is the reason.

NATS server hosting this bucket is also spinning at 100% CPU during traffic spike, so it might be the clue.

### Expected behavior

Work properly, or just fail with exception, don't get stuck.

### Server and client version

Server: 2.12.2
Client: 2.6.11
.NET: 10

### Host environment

Kubernetes @ ARM

### Steps to reproduce

Hard to say. There is constant traffic of let's say two thousands messages all the time. This is background traffic, unrelated to this feature.
Then, massive upload spike happens (involving ObjectStore) and things get weird.
As last time, I am unable to reproduce it with my laptop.

mtmk:
* Potential fix for #1013

### DIFF
--- a/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
@@ -39,6 +39,12 @@ internal record NatsJSOrderedPushConsumerOpts
     public ConsumerConfigDeliverPolicy DeliverPolicy { get; init; } = ConsumerConfigDeliverPolicy.All;
 
     public bool HeadersOnly { get; init; } = false;
+
+    /// <summary>
+    /// Timeout for cleanup operations during disposal (e.g. deleting ephemeral consumers).
+    /// If the server is slow or unresponsive, disposal will not block longer than this.
+    /// </summary>
+    public TimeSpan CleanupTimeout { get; init; } = TimeSpan.FromSeconds(5);
 }
 
 internal class NatsJSOrderedPushConsumer<T>
@@ -158,7 +164,19 @@ internal class NatsJSOrderedPushConsumer<T>
 
         _msgChannel.Writer.TryComplete();
 
-        await _context.DeleteConsumerAsync(_stream, Consumer, _cancellationToken);
+        using var cts = new CancellationTokenSource(_opts.CleanupTimeout);
+        try
+        {
+            await _context.DeleteConsumerAsync(_stream, Consumer, cts.Token);
+        }
+        catch (OperationCanceledException)
+        {
+            // Timeout - server will clean up ephemeral consumer automatically
+        }
+        catch (NatsJSApiException)
+        {
+            // Consumer may already be gone, that's fine
+        }
     }
 
     internal void Init()
--- a/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
@@ -164,11 +164,11 @@ internal class NatsJSOrderedPushConsumer<T>
 
         _msgChannel.Writer.TryComplete();
 
-        using var cts = new CancellationTokenSource(_opts.CleanupTimeout);
-        using var ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, _cancellationToken);
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(_cancellationToken);
+        cts.CancelAfter(_opts.CleanupTimeout);
         try
         {
-            await _context.DeleteConsumerAsync(_stream, Consumer, ctsLinked.Token);
+            await _context.DeleteConsumerAsync(_stream, Consumer, cts.Token);
         }
         catch (OperationCanceledException)
         {
--- a/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
@@ -165,9 +165,10 @@ internal class NatsJSOrderedPushConsumer<T>
         _msgChannel.Writer.TryComplete();
 
         using var cts = new CancellationTokenSource(_opts.CleanupTimeout);
+        using var ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, _cancellationToken);
         try
         {
-            await _context.DeleteConsumerAsync(_stream, Consumer, cts.Token);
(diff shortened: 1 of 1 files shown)


src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
```diff
@@ -158,7 +164,19 @@ public async ValueTask DisposeAsync()
 
         _msgChannel.Writer.TryComplete();
 
-        await _context.DeleteConsumerAsync(_stream, Consumer, _cancellationToken);
+        using var cts = new CancellationTokenSource(_opts.CleanupTimeout);
```
MiloszKrajewski: I think combining both, timeout and original token, would be even better:

```csharp
using var cts = CancellationTokenSource.CreateLinkedTokenSource(_cancellationToken);
cts.CancelAfter(_opts.CleanupTimeout);
```

mtmk: fair point. I can link them.

src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
```diff
@@ -165,9 +165,10 @@ public async ValueTask DisposeAsync()
         _msgChannel.Writer.TryComplete();
 
         using var cts = new CancellationTokenSource(_opts.CleanupTimeout);
+        using var ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, _cancellationToken);
```
MiloszKrajewski: Don't want to be preachy about it, but I cannot resist ;-)

This approach:

```csharp
using var cts = CancellationTokenSource.CreateLinkedTokenSource(_cancellationToken);
cts.CancelAfter(_opts.CleanupTimeout);
```

creates one CancellationTokenSource and one (internal) Timer.
While this one:

```csharp
using var cts = new CancellationTokenSource(_opts.CleanupTimeout);
using var ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, _cancellationToken);
```

creates two CancellationTokenSources and one (internal) Timer. 

The first one is just a little bit lighter resource-wise.
Also one less implicit `finally`.

mtmk: nice! done

scottf (approved):
LGTM

PR #1016, opened 2025-12-12
Fix service hangs when stopping

mtmk:
I noticed this when running tests in Rider where process sticks around even after tests finishes.
While 'flushing' the buffers, *ping* in our implementation waits for the connecting to come back up and in the case of testing, that never happens, causing the process to hang.
There maybe other scenarios where service needs to be stopped and the server is not accessible, perhaps in IoT applications.


scottf (approved):
LGTM

PR #1021, opened 2025-12-17
PeerInfo.Lag is wrong type

### LINKED ISSUE #1020: PeerInfo.Lag is the wrong type
### Observed behavior

PeerInfo.Lag should be a ulong rather than an int32. 

Go version is a uint64

https://github.com/nats-io/nats-server/blob/daa61e2ed8d8e0dfc9099bfb665ade24071b6e6a/server/stream.go#L322

### Expected behavior

Shouldn't get a serialization error when number is deserialized if bigger than an int32

### Server and client version

latest client

### Host environment

_No response_

### Steps to reproduce

_No response_

darkwatchuk:
Changed PeerInfo.Lag to ulong to match GOLANG version.


mtmk (approved):
LGTM thanks @darkwatchuk

darkwatchuk:
Patched similar in the following according to the GOLANG version.

AccountLimits    int -> long

 .MaxMemory
 .MaxStorage
 .MemoryMaxStreamBytes
 .StorageMaxStreamBytes


ApiStats    int -> ulong
 .Total
 .Errors

PR #1031, opened 2026-01-05
Fix OTel INatsJSMsg extension

### LINKED ISSUE #1026: [v2.7.0] `StartActivity` does not work with `ConsumeAsync` return type
### Observed behavior

After updating to **v2.7.0**, the following code no longer compiles:

```csharp
await foreach (var message in consumer.ConsumeAsync<byte[]>(cancellationToken: stoppingToken))
{
    using var activity = message.StartActivity("nats.consume"); // <-- ERROR
    activity?.SetTag("messaging.system", "nats");
    activity?.SetTag("messaging.subject", message.Subject);

    // ...
}
```

In v2.7.0, `ConsumeAsync<T>()` now returns `INatsJSMsg<T>`, but the `StartActivity` extension method is defined on `NatsJSMsg<T>`, so it’s no longer available.

Would it make sense to change `StartActivity` to extend `INatsJSMsg<T>` instead?

### Expected behavior

See Observed behavior.

### Server and client version

See Observed behavior.

### Host environment

_No response_

### Steps to reproduce

_No response_

mtmk:
fixes #1026

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSTelemetryExtensions.cs
+++ b/src/NATS.Client.JetStream/NatsJSTelemetryExtensions.cs
@@ -26,4 +26,24 @@ public static class NatsJSTelemetryExtensions
             parentContext: msg.Headers.GetActivityContext(),
             tags: tags);
     }
+
+    /// <summary>Start an activity under the INatsJSMsg associated activity.</summary>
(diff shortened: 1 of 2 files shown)


scottf (approved):
LGTM

PR #1032, opened 2026-01-06
Fix 2.6/2.7 ABI for serializer interfaces

### LINKED ISSUE #1029: 2.7.0 INatsSerialize ABI break in INatsSerialize
### Observed behavior

Moving INatsSerialize to a new assembly caused ABI break in 2.7.0:

https://github.com/nats-io/nats.net/commits/872fb9578366cf835e5f355e1e6214244ebc6aaf/src/NATS.Client.Abstractions/INatsSerialize.cs

So any methods with serialize/deserialze as params are going to break when referenced in packages built targeting pre-2.7.0 NATS.Net

### Potential Fix

Short of moving them back, it may be possible to use [TypeForwardedToAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.typeforwardedtoattribute?view=net-10.0)

### Server and client version

Client: 2.7.0

mtmk:
Fixes #1029

### DIFF
--- a/.github/workflows/test_linux.yml
+++ b/.github/workflows/test_linux.yml
@@ -146,3 +146,8 @@ jobs:
             cd tests/NATS.Net.DocsExamples
             dotnet run
 
+      - name: Check ABI Compatibility (Type Forwarding)
+        run: |
+            cd tests/NATS.Client.CheckAbi
+            bash run-abi-check.sh
+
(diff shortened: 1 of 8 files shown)


caleblloyd (approved):
Fix looks good.  A future improvement to the test could be using something like `Microsoft.DotNet.ApiCompat.Tool` to check compat

scottf (approved):
LGTM

PR #1033, opened 2026-01-07
Add API Compat for TFMs

mtmk:
Thanks @caleblloyd recommending to use apicompat tool.

https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/compatible-framework-in-package-validator

I thought we can take advantage of this quite quickly checking our APIs across different targets.


caleblloyd (approved):
LGTM

PR #1039, opened 2026-01-11
resolves: 1035 - Use ConnectTimeout when awaiting INFO signal

### LINKED ISSUE #1035: NATS.Client.Core - Reconnect Retries Bound by RequestTimeout - NatsConection.cs:line:529
### Observed behavior

After a successful NATS Connection, upon loss of that connection the configure retry policy/settings are executed. During this reconnect/retry loop, if the NATS Server is unreachable, we land on NatsConnection.cs:line:529, specifically the `WaitAsync` hanging on the `waitForInfoSignal` Task. This respects the timeout provided via `RequestTimeout` of the connection's NatsOpts.
In my use-case, the default `RequestTimeout` needs to be relatively long, on the order of ~1 minute. That led to observing Retries seemingly hang well beyond expectations, I had expected the timeout to be the `ConnectTimeout`, not the `RequestTimeout`.
Working around this, a minimal Request Timeout can be set for the purpose of Retries, and the longer duration needs overridden via NatsSubOpts but this can become less than ideal and seems unnecessary

### Expected behavior

The expected behavior would be that during retry attempts, the connection timeout respects the `NatsOpts.ConnectTimeout` value.

Am I off base here, should this really be 'RequestTimeout' here? Seems like a simple fix I could PR quick if it's just mix up

### Server and client version

Server: tag 2.12
Client: 2.7.0

### Host environment

_No response_

### Steps to reproduce

`docker run nats/nats`

```csharp
var connection = new NatsConnection(NatsOpts.Default with
{
    RequestTimeout = TimeSpan.FromMinutes(5),
    ConnectTimeout = TimeSpan.FromSeconds(1),
    MaxReconnectRetry = 2
});
connection.OnConnectingAsync = ((string Host, int Port) args) =>
{
   count++;
   logger.LogCritical("CONNECTING {Count} {Host}:{Port}", count, args.Host, args.Port);
   return ValueTask.FromResult(args);
};
Debug.Assert(connection.ConnectionState is NatsConnectionState.Open)
Console.ReadLine()
```

`docker stop nats`

and observe only a single attempt is made and holds for the duration of RequestTimeout, as opposed to the 1 second Connect Timeout

ptsteward:
## Context: Issue #1035 
After a successful NATS Connection, upon loss of that connection the configured retry policy/settings are executed. During this reconnect/retry loop, if the NATS Server is unreachable, we land on NatsConnection.cs:line:529, specifically the `WaitAsync` hanging on the `waitForInfoSignal` Task. This respects the timeout provided via `RequestTimeout` of the connection's `NatsOpts`.
In my use-case, the default `RequestTimeout` needs to be relatively long, on the order of ~1 minute. That led to observing Retries seemingly hanging well beyond expectations, I had expected the timeout to be the `ConnectTimeout`, not the `RequestTimeout`.

## This Change
This PR swaps the timeout from `NatsOpts.RequestTimeout` to the expected `NatsOpts.ConnectTimeout`.

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -526,7 +526,7 @@ public partial class NatsConnection : INatsConnection
         {
             // Wait for an INFO message from server. If we land on a dead socket and server response
             // can't be received, this will throw a timeout exception and we will retry the connection.
-            await waitForInfoSignal.Task.WaitAsync(Opts.RequestTimeout).ConfigureAwait(false);
+            await waitForInfoSignal.Task.WaitAsync(Opts.ConnectTimeout).ConfigureAwait(false);
 
             // check to see if we should upgrade to TLS
             if (_socketConnection!.InnerSocket is INatsTlsUpgradeableSocketConnection tlsUpgradeableSocket)


mtmk (approved):
LGTM thanks @ptsteward

PR #1041, opened 2026-01-13
Fix slow consumer blocking

mtmk:
**Problem**
When internal bounded channels use `BoundedChannelFullMode.Wait` (the default), a slow consumer causes `WriteAsync()` to block until space becomes available. If these channels are written directly from the socket reader thread, the entire NATS connection freezes until space become available. This causes the other requests to timeout under load and in some cases it can create deadlocks. Note that slow consumers are handled at NATS Core calls but since channels are implemented at subscriber classes and not at the base class `NatsSubBase` to avoid or reduce channel chains, higher level APIs like JetStrean, KV, Object Store and Services have to implement it separately.

Issue manifest itself as some form of 'No Response' exception (e.g. `NatsJSPublishNoResponseException`) and I have seen it reported several times on GitHub (e.g. #763) or on Slack discussions. I always thought it was a Task scheduling or thread pool issue but in this case it turned out to be a lot simpler.

**Solution**
Channels directly on the socket reader path now use `BoundedChannelFullMode.DropNewest` (default from options) with an `OnMessageDropped` callback to notify applications. When message drops happen recovery is done via standard NATS mechanisms such as message redelivery or consumer reset on sequence mismatch. This behavior is now fully in line with other clients (Go, Java, Rust).

**Changes**
- Fix bounded channels blocking the socket reader thread when consumers are slow: they now use `DropNewest` mode (default from options) with `OnMessageDropped` callbacks
- Added tests for slow consumers

### DIFF
--- a/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
@@ -146,15 +146,19 @@ internal class NatsJSConsume<TMsg> : NatsSubBase
             Timeout.Infinite,
             Timeout.Infinite);
 
-        // This channel is used to pass messages
-        // to the user from the subscription channel (which should be set to a
-        // sufficiently large value to avoid blocking socket reads in the
-        // NATS connection).
-        _userMsgs = Channel.CreateBounded<NatsJSMsg<TMsg>>(1000);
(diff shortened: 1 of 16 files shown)


scottf (approved):
LGTM

caleblloyd (approved):
LGTM

PR #1047, opened 2026-01-22
Fix OTel activity leak in Direct request mode

### LINKED ISSUE #862: Using `NatsRequestReplyMode.Direct` screws up OpenTelemetry traces

mtmk:
Fix OpenTelemetry activity not being disposed when using `NatsRequestReplyMode.Direct`.

The activity stored in message headers is now explicitly disposed after receiving the response, matching the behavior of `SharedInbox` mode which disposes via `ActivityEndingMsgReader`.

Fixes #862

### DIFF
--- a/examples/Example.OpenTelemetry/ClientApp.cs
+++ b/examples/Example.OpenTelemetry/ClientApp.cs
@@ -24,7 +24,10 @@ public static class ClientApp
 
         Console.WriteLine("Client App is starting...");
 
-        await using var nats = new NatsConnection();
+        await using var nats = new NatsConnection(new NatsOpts
+        {
+            RequestReplyMode = NatsRequestReplyMode.Direct,
+        });
 
         using (var activity = activitySource.StartActivity("SayHi"))
         {
--- a/examples/Example.OpenTelemetry/ServiceApp.cs
+++ b/examples/Example.OpenTelemetry/ServiceApp.cs
@@ -24,7 +24,10 @@ public static class ServiceApp
 
         Console.WriteLine("Service App is starting...");
 
-        await using var nats = new NatsConnection();
+        await using var nats = new NatsConnection(new NatsOpts
+        {
+            RequestReplyMode = NatsRequestReplyMode.Direct,
+        });
 
         await foreach (var msg in nats.SubscribeAsync<string>("greet.>"))
         {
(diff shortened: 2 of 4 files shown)


mtmk:
sorry for the long delay @aradalvand I'd appreciate it if you can check it out when you get a minute please. thanks

scottf (approved):
LGTM

aradalvand:
Hi @mtmk; thank you for the effort. Did you actually test this? I say that because I remember doing something vaguely similar and it did not appear to work, so the solution may not be that simple.

mtmk:
@aradalvand 

> Hi @mtmk; thank you for the effort. Did you actually test this? I say that because I remember doing something vaguely similar and it did not appear to work, so the solution may not be that simple.

good point! i can/should write a test case for it. can you give me a rough scenario please?

aradalvand:
I think just doing a simple request and ensuring the span/activity is closed after the reply is received would suffice. It's just important to run any test you write against the previous code as well to make sure this fix is actually doing the right thing.

mtmk:
I tried Example.OpenTelemetry project with jaeger:

- (1) Direct mode before the fix:
image

- (2) Direct mode after the fix:
image

- (3) Shared inbox:
image

My interpretation:
`(1) Direct mode before the fix` doesn't close the span missing the last `ClientApp inbox.receive` span is missing. `(2) Direct mode after the fix` correctly have the `ClientApp inbox.receive` span.

the difference between `(3) Shared inbox` and `(2) Direct mode` is that `inbox.subscribe` (right after the `request`) doesn't exist since direct mode isn't creating a shared inbox.

scottf (approved):
LGTM

PR #1048, opened 2026-01-22
Add TermWithReason support to AckTerminateAsync

mtmk:
Add optional `reason` parameter to `AckTerminateAsync` that sends `+TERM <reason>` to the server, matching the Go client's `TermWithReason` functionality.

The reason appears in JetStream advisory events (requires NATS Server 2.10.4+).

Low-allocation implementation using `ArrayPool<byte>` with platform-specific optimizations for netstandard2.0 vs newer targets.

⚠️ This is a breaking change for applications implementing `INatsJSMsg<T>`

I only expect tests would break and can be detected and fairly easily fixed during build. However, if applications are using their own implementations of `INatsJSMsg<T>` they will break during build too but more severely if they are transitive dependencies using an earlier version of `NATS.Client.JetStream` they would fail at runtime.

### Interface `INatsJSMsg<T>` Changes

| Change | Binary Compatible | Source Compatible |
|--------|:-----------------:|:-----------------:|
| `default` → `null` on `ReplyAsync` params | ✅ Yes | ✅ Yes |
| `default` → `null` on `AckAsync` param | ✅ Yes | ✅ Yes |
| `default` → `null` on `NakAsync` param | ✅ Yes | ✅ Yes |
| `default` → `null` on `AckProgressAsync` param | ✅ Yes | ✅ Yes |
| `default` → `null` on `AckTerminateAsync` param | ✅ Yes | ✅ Yes |
| **New method `AckTerminateAsync(string reason, ...)`** | ❌ **No** | ❌ **No** |

---

### Struct `NatsJSMsg<T>` Changes

| Change | Binary Compatible | Source Compatible |
|--------|:-----------------:|:-----------------:|
| `default` → `null` on all method params | ✅ Yes | ✅ Yes |
| New overload `AckTerminateAsync(string reason, ...)` | ✅ Yes | ✅ Yes |
| `delay == default` → `delay == TimeSpan.Zero` | ✅ Yes | ✅ Yes |
| `Connection == default` → `Connection == null` | ✅ Yes | ✅ Yes |
| Doc comments → `<inheritdoc />` | ✅ Yes | ✅ Yes |
| Internal `NatsJSMsgConstants` class | ✅ Yes | ✅ Yes |

---

### Final Verdict

| Usage Pattern | Safe to Ship? |
|---------------|:-------------:|
| Applications using `NatsJSMsg<T>` directly | ✅ **Yes** |
| Applications calling methods via `INatsJSMsg<T>` | ✅ **Yes** |
| Applications/libraries implementing `INatsJSMsg<T>` | ❌ **No** |

---

### Breaking Change Details

**What breaks:**

```csharp
// Any type implementing INatsJSMsg<T> - e.g., user mocks for testing
public class MockJSMsg<T> : INatsJSMsg<T>
{
    // Must now implement:
    // ValueTask AckTerminateAsync(string reason, AckOpts? opts, CancellationToken ct)
}
```

**How it breaks:**

Scenario: Source break
Failure Mode: Compilation error when updating package and recompiling

Scenario: Binary break
Failure Mode: TypeLoadException at runtime if old compiled mock is loaded with new library version (transitive dependency scenario)

Who is affected:

- Users who created test doubles/mocks of INatsJSMsg<T>
- Libraries that expose their own implementations of INatsJSMsg<T>

Who is NOT affected:

- Users who only consume messages (vast majority)
- Users who only use the concrete NatsJSMsg<T> struct

---

**Risk Assessment**

Factor: Likelihood of user interface implementations
Assessment: Medium - Interface docs explicitly mention "creating test doubles in unit testing"

Factor: Severity of break
Assessment: High - TypeLoadException crashes app at startup

Factor: Detectability
Assessment: Low for transitive - User may not know they have a dependency with an implementation

Factor: SemVer classification
Assessment: Breaking change - requires major version bump per SemVer

---

**Conclusion**

The current implementation is a breaking change.

It will cause TypeLoadException at runtime for any code that implements INatsJSMsg<T>, including through transitive dependency updates where the user never modified their own code.

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSMsg.cs
+++ b/src/NATS.Client.JetStream/NatsJSMsg.cs
@@ -8,7 +8,7 @@ namespace NATS.Client.JetStream;
 
 /// <summary>
 /// This interface provides an optional contract when passing
-/// messages to processing methods which is usually helpful in
+/// messages to processing methods, which is usually helpful in
 /// creating test doubles in unit testing.
 /// </summary>
 /// <remarks>
@@ -30,12 +30,12 @@ namespace NATS.Client.JetStream;
 public interface INatsJSMsg<out T> : INatsMsg
 {
     /// <summary>
-    /// Subject of the user message.
+    /// Get subject of the user message.
     /// </summary>
     string Subject { get; }
 
     /// <summary>
-    /// Message size in bytes.
+    /// Get message size in bytes.
     /// </summary>
     /// <remarks>
     /// Message size is calculated using the same method NATS server uses:
@@ -46,12 +46,12 @@ public interface INatsJSMsg<out T> : INatsMsg
     int Size { get; }
 
     /// <summary>
-    /// Deserialized user data.
+    /// Get deserialized user data.
     /// </summary>
     T? Data { get; }
 
     /// <summary>
-    /// The connection messages was delivered on.
+    /// Get the connection messages were delivered on.
     /// </summary>
     INatsConnection? Connection { get; }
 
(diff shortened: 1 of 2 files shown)


scottf (approved):
LGTM

mtmk:
Alternative non breaking change using extension method:
```csharp
  public static class NatsJSMsgExtensions
  {
      public static ValueTask AckTerminateAsync<T>(
          this NatsJSMsg<T> msg,
          string reason,
          AckOpts? opts = null,
      {
          return msg.AckTerminateInternalAsync(reason, opts, cancellationToken);
      }
  }
```

**Down side** not useful for testing

mtmk:
we can probably take the risk of breaking `INatsJSMsg<T>` in this occasion.

PR #1051, opened 2026-01-26
FIXED unobserved exceptions surfacing due to unobserved tcs.

### LINKED ISSUE #1050: Authorization violation when connecting followed by recreating said connection causes unobserved SocketClosedException
### Observed behavior

If an authorization violation is thrown during connection via `NatsConnection` and you then recreate said connection, the garbage collector will find an unobserved `SocketClosedException`. This occurs due to the  `TaskCompletionSource` in `SocketConnectionWrapper` setting an exception but never observing it.

**Proposed Solution**
My proposed solution would be to create an extension for `TaskCompletionSource` with a method that both sets the exception and observes it immediately, then use this method at all locations where `TrySetException` is used:

```
internal static bool TrySetObservedException(this TaskCompletionSource source, Exception ex)
{
    var result = source.TrySetException(ex);

    if (result)
    {
        // Suppress unobserved exceptions as the exceptions will surface elsewhere
        _ = source.Task.Exception;
    }

    return result;
}
```
If this seems reasonable i can open a PR with the suggested change.

### Expected behavior

No unobserved SocketClosedException should surface.

### Server and client version

This was reproduced in blackbox and unit tests with version 2.7.2 preview

### Host environment

_No response_

### Steps to reproduce

The code below will cause an unobserved exception to occur if the server process is started with a user config that would cause an authentication error when connecting.
```
var exceptionThrown = false;
TaskScheduler.UnobservedTaskException += (_, _) =>
{
    // The unobserved exception will be a SocketClosedException.
    exceptionThrown = true;
};

// Connect twice to recreate the connection and trigger unobserved exception
await Task.Run(async () =>
{
    for (var i = 0; i < 2; i++)
    {
        var conn = new NatsConnection(new NatsOpts
        {
            // The server process should be setup with a user config that will cause an auth error
            Url = server.Url,
        });

        try
        {
            await conn.ConnectAsync();
        }
        catch (NatsException)
        {
            // Ignore the expected authentication error
        }

        await conn.DisposeAsync();
    }
});

// Wait for the unobserved exception to be raised
await Task.Delay(500);

GC.Collect();

Assert.True(exceptionThrown);
```

me064:
Resolves #1050 

Fixed unobserved exceptions surfacing from `TaskCompletionSource` objects setting exceptions but not observing them.

The fix includes an extension method `TrySetObservedException` for `TaskCompletionSource` that sets the exception and immediately observes it to avoid any unobserved exceptions. All usages of `TrySetException` has been replaced with `TrySetObservedException`.

Two new tests have been added and one unrelated test has been modified due to it causing unobserved exceptions that are then caught in one of the new tests, causing it to fail. 

My recommendation for the future would be to create a global garbage collection test or similar to ensure no future unobserved exceptions appear throughout the library.

### DIFF
--- a/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
+++ b/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
@@ -298,8 +298,8 @@ internal sealed class NatsReadProtocolProcessor : IAsyncDisposable
             catch (SocketClosedException e)
             {
                 _logger.LogDebug(NatsLogEvents.Protocol, e, "Socket closed during read loop");
-                _waitForInfoSignal.TrySetException(e);
-                _waitForPongOrErrorSignal.TrySetException(e);
+                _waitForInfoSignal.TrySetObservedException(e);
+                _waitForPongOrErrorSignal.TrySetObservedException(e);
                 return;
             }
             catch (Exception ex)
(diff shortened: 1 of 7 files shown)


mtmk:
@me064 thanks for the PR it's looking good. The test seems to be failing for some reason. Once you fix the test I can have one last look then merge.

mtmk (approved):
LGTM thanks @me064

PR #1052, opened 2026-01-26
Add SlowConsumerDetected event and warnings

mtmk:
Introduced `SuppressSlowConsumerWarnings ` in `NatsOpts` to control whether warnings about dropped messages are logged as slow consumers. Also implemented slow consumer detection for these warnings and `SlowConsumerDetected` event.


scottf (approved):
LGTM

PR #1025, opened 2025-12-19
Subject is always read as ASCII, not UTF-8

aafloy:
Fix the issue of #1023 by parsing subject with UTF8 as outlined in the protocol convention.

Since NatsOpts contains a SubjectEncoding, the fix ideally should use this, however, the protocol convention is clear on what encoding the subject is, so I would argue that the NatsOpts.SubjectEncoding should be removed.

PR #1054, opened 2026-01-29
Fix subject encoding to use UTF-8 per NATS protocol spec

### LINKED ISSUE #1023: Subject is always read as ASCII, not UTF-8
### Observed behavior

Protocol conventions states that:

"All UTF-8 characters except spaces/tabs and separators which are . and > are allowed."

https://docs.nats.io/reference/reference-protocols/nats-protocol#protocol-conventions

The subject is always read with Encoding.ASCII, instead of Encoding.UTF8 or NatsOpts.SubjectEncoding

https://github.com/nats-io/nats.net/blob/4856762f510ad2388c50810ed3e6764e4cb84bba/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs#L467

https://github.com/nats-io/nats.net/blob/4856762f510ad2388c50810ed3e6764e4cb84bba/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs#L519

### Expected behavior

Sending a message with UTF-8 characters in subject should be received with the matching subject.

### Server and client version

Server version 2.12.0, .NET Client version 2.6.14

### Host environment

Docker/Linux

### Steps to reproduce

Send a message with a UTF-8 encoded character in subject and receive the message with .NET client.

mtmk:
- Change subject and reply-to parsing from `Encoding.ASCII` to configurable `SubjectEncoding` (default: `Encoding.UTF8`) per the NATS protocol convention
- Update both read side (`NatsReadProtocolProcessor`) and write side (`NatsOpts.SubjectEncoding` default) so `UTF-8` subjects round-trip correctly
- Add XML docs for `HeaderEncoding` and `SubjectEncoding` options

Based on the work by @aafloy in #1025.

Resolves #1023.

### DIFF
--- a/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
+++ b/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
@@ -21,12 +21,14 @@ internal sealed class NatsReadProtocolProcessor : IAsyncDisposable
     private readonly Task _infoParsed; // wait for an upgrade
     private readonly ConcurrentQueue<PingCommand> _pingCommands; // wait for pong
     private readonly ILogger<NatsReadProtocolProcessor> _logger;
+    private readonly Encoding _subjectEncoding;
     private readonly bool _trace;
     private int _disposed;
 
     public NatsReadProtocolProcessor(SocketConnectionWrapper socketConnection, NatsConnection connection, TaskCompletionSource waitForInfoSignal, TaskCompletionSource waitForPongOrErrorSignal, Task infoParsed)
     {
         _connection = connection;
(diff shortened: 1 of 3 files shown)


scottf (approved):
LGTM

aafloy:
Thanks for pulling my quick fix in the right direction @mtmk. I don't have the full view of NATS in my head, so.. Thanks! :)

mtmk:
> Thanks for pulling my quick fix in the right direction @mtmk. I don't have the full view of NATS in my head, so.. Thanks! :)

no problem at all @aafloy thanks for the original PR 🚀 please feel free to drop more whenever you spot something. always welcome. Appreciate it! 😊

PR #1055, opened 2026-01-30
Add more tests for JetStream consumer behavior

### LINKED ISSUE #933: More test coverage: EphemeralConsumerTest
https://github.com/nats-io/nats.net/issues/927#issuecomment-3189294980

```csharp
using NATS.Client.Core;
using NATS.Client.JetStream;
using log4net;
using log4net.Appender;
using log4net.Config;
using log4net.Layout;
using NATS.Net;
using System.Diagnostics;
using Microsoft.Extensions.Logging.Abstractions;

public class EphemeralConsumerTest
{
    private static readonly ILog Log = LogManager.GetLogger(typeof(EphemeralConsumerTest));
    private static INatsJSConsumer _consumer;
    private static INatsJSContext _context;
    private const string Url = "127.0.0.1:4222";
    private static CancellationTokenSource _cts = null;
    private static Task _task = null;

    private const string DestName = "messaging.destination.name";
    private const string OpKey = "messaging.operation";
    private const string OpPub = "publish";
    private const string ConsumerNextVerb = "$JS.API.CONSUMER.MSG.NEXT";
    private const string NatsActivitySource = "NATS.Net";
    // Connect to NATS server. Since connection is disposable at the end of our scope we should flush

    // our buffers and close connection cleanly.
    private static NatsOpts opts = NatsOpts.Default with
    {
        Url = Url,
        Name = "Spiral-503-Example",
        AuthOpts = new NatsAuthOpts() { Username = "admin", Password = "admin" },
        LoggerFactory = new NullLoggerFactory(),//new NatsNetLoggerFactory(LogLevel.Trace),
        RequestTimeout = TimeSpan.FromSeconds(15),
        PingInterval = TimeSpan.FromSeconds(1800),
        MaxPingOut = 100
    };

    private static NatsConnection _connection = null;

    static async Task Main(string[] args)
    {
        Activity.DefaultIdFormat = ActivityIdFormat.W3C;
        Activity.ForceDefaultIdFormat = true;
        
        var consumerMsgNextStatement = $"{ConsumerNextVerb}.{_consumer.Info.StreamName}.{_consumer.Info.Name}";

        ActivitySource.AddActivityListener(new ActivityListener()
        {
            ShouldListenTo = (source) => string.Equals(source.Name, NatsActivitySource, StringComparison.OrdinalIgnoreCase),
            Sample = (ref ActivityCreationOptions<ActivityContext> options) => ActivitySamplingResult.AllDataAndRecorded,
            ActivityStarted = activity => LogActivityStarted(activity, consumerMsgNextStatement, _consumer.Info.Name),
            ActivityStopped = _ => { },
        });

        ConfigureLog4Net();
       
        _task = Task.Run(RunMessageLoop);
        await _task;
    }

    private static void LogActivityStarted(Activity activity, string consumerMsgNextStatement, string consumerName)
    {
        if (activity.OperationName == $"$JS.API {OpPub}" && activity.Tags.Any(t => t is { Key: OpKey, Value: OpPub })
            && activity.Tags.Any(t => t.Key == DestName && t.Value == consumerMsgNextStatement))
        {
            var tag = activity.Tags.First(t => t.Key == DestName);
            Log.Info($"Issuing another pull request for the consumer, Consumer={consumerName}, Destination={tag.Value}");
        }
    }

    private async static Task RunMessageLoop()
    {
        _cts = new CancellationTokenSource(TimeSpan.FromMinutes(20));
        var opts = new NatsJSConsumeOpts
        {
            Expires = TimeSpan.FromSeconds(30),
            IdleHeartbeat = TimeSpan.FromSeconds(15),
            MaxMsgs = 1000
        };

        await foreach (var msg in _consumer.ConsumeAsync<string>(opts: opts, cancellationToken: _cts.Token))
        {
            await Task.Delay(TimeSpan.FromSeconds(10));//emulate processing
            await msg.AckAsync();
            Console.WriteLine($"received msg on {msg.Subject} with data {msg.Data}");
        }
    }

    static EphemeralConsumerTest()
    {
        _connection = new NatsConnection(opts);
        _context = _connection.CreateJetStreamContext();
        _consumer = CreateConsumer().GetAwaiter().GetResult();
    }

    private static async Task<INatsJSConsumer> CreateConsumer()
    {
        var consumerCfg = new ConsumerConfig()
        {
            FilterSubjects = ["TS.TEST.Incoming", "TS.TEST.Incoming.>"],
            MemStorage = true,
            InactiveThreshold = TimeSpan.FromSeconds(600),
            Name = "Ephemeral1",
            DurableName = null,
            DeliverPolicy = ConsumerConfigDeliverPolicy.New,
            MaxWaiting = 1000,
            AckWait = TimeSpan.FromSeconds(30),
            MaxAckPending = 1000,
            MaxDeliver = 20
        };

        var _consumer = await _context.CreateOrUpdateConsumerAsync("TS-TEST", consumerCfg);

        return _consumer;
    }

    static void ConfigureLog4Net()
    {
        var layout = new PatternLayout("[%date] %-5level - %message%newline");

        var consoleAppender = new ConsoleAppender
        {
            Layout = layout,
            Threshold = log4net.Core.Level.Info
        };
        layout.ActivateOptions();
        consoleAppender.ActivateOptions();

        var repository = LogManager.GetRepository(System.Reflection.Assembly.GetEntryAssembly());
        BasicConfigurator.Configure(repository, consoleAppender);
    }
}
```

```pwsh
nats stream add 'TS-TEST' `
 --storage=file `
 --compression=none `
 --replicas 1 `
 --retention=limits `
 --discard=old `
 --max-age=2d `
 --max-bytes=3GB `
 --max-msg-size=32MB `
 --max-msgs=20000 `
 --subjects="TS.TEST.Incoming.>,TS.TEST.Incoming" `
 --defaults
```

mtmk:
- Added tests to validate consumer behavior when deleted or restarted:
  * Consume with ephemeral consumer deletion (409 response validation).
  * Handle server restart with memory storage consumer (503 response validation).
  * Managing slow message processing with consumer deletion.
  * Accumulation of 503 errors across heartbeat pulls.
Resolves #933

### DIFF
--- a/tests/NATS.Client.JetStream.Tests/ConsumerConsumeTest.cs
+++ b/tests/NATS.Client.JetStream.Tests/ConsumerConsumeTest.cs
@@ -961,6 +961,76 @@ public class ConsumerConsumeTest
     }
 
     [Fact]
+    public async Task Consume_buffered_messages_with_slow_processing_still_detects_503_after_server_restart()
+    {
+        // When messages are buffered and being slowly processed, the heartbeat timer
+        // is the only mechanism to drive new pull requests after 503 (since ResetPending
+        // sets pending to max, preventing CheckPending from issuing pulls).
+        // We verify that even with slow message processing, the 503 threshold is
+        // eventually hit via timer-driven pulls after the consumer vanishes.
+        var logger = new InMemoryTestLoggerFactory(LogLevel.Debug);
+        await using var server = await NatsServerProcess.StartAsync();
+        await using var nats = new NatsConnection(new NatsOpts { Url = server.Url, LoggerFactory = logger });
+        await nats.ConnectRetryAsync();
+
+        var cts = new CancellationTokenSource(TimeSpan.FromSeconds(60));
+
+        var js = nats.CreateJetStreamContext();
+        await js.CreateStreamAsync(new StreamConfig("s1", ["s1.*"]), cts.Token);
+        var consumer = await js.CreateOrUpdateConsumerAsync("s1", new ConsumerConfig("c1") { MemStorage = true }, cts.Token);
+
+        // Publish many messages so the buffer has data to process slowly
+        for (var i = 0; i < 20; i++)
+        {
+            var ack = await js.PublishAsync("s1.foo", i, cancellationToken: cts.Token);
+            ack.EnsureSuccess();
+        }
+
+        var firstMessageConsumed = new WaitSignal(TimeSpan.FromSeconds(30));
+        var messagesProcessed = 0;
+
+        var consumeTask = Task.Run(async () =>
+        {
+            var opts = new NatsJSConsumeOpts
+            {
+                MaxMsgs = 100,
+                MaxConsecutive503Errors = 3,
+                IdleHeartbeat = TimeSpan.FromSeconds(1),
+                Expires = TimeSpan.FromSeconds(2),
+            };
+
+            await foreach (var msg in consumer.ConsumeAsync<int>(opts: opts, cancellationToken: cts.Token))
+            {
+                await msg.AckAsync(cancellationToken: cts.Token);
+                var count = Interlocked.Increment(ref messagesProcessed);
+                firstMessageConsumed.Pulse();
+
+                // Slow processing on all messages
+                await Task.Delay(TimeSpan.FromMilliseconds(500), cts.Token);
+            }
+        });
+
+        // Wait until messages start flowing, then restart server
+        await firstMessageConsumed;
+
+        await server.RestartAsync();
+
+        // Even though the consumer is slowly processing buffered messages,
+        // the 503 detection should still terminate ConsumeAsync.
+        // After restart, the MemStorage consumer is gone, so timer-driven pulls
+        // return 503 and accumulate until the threshold is hit.
+        var exception = await Assert.ThrowsAnyAsync<NatsJSException>(async () => await consumeTask);
+        Assert.Contains("503", exception.Message);
+
+        // Verify some messages were processed before termination
+        Assert.True(messagesProcessed > 0, "Should have processed at least one message");
+
+        // Verify 503 errors accumulated
+        var noResponderLogs = logger.Logs.Count(m => m.EventId == NatsJSLogEvents.NoResponders);
+        Assert.True(noResponderLogs > 0, "Should have logged NoResponders events");
+    }
+
+    [Fact]
     public async Task Consume_503_counter_accumulates_across_heartbeat_timer_pulls()
     {
         var logger = new InMemoryTestLoggerFactory(LogLevel.Debug);
--- a/tests/NATS.Client.JetStream.Tests/ConsumerConsumeTest.cs
+++ b/tests/NATS.Client.JetStream.Tests/ConsumerConsumeTest.cs
@@ -794,4 +794,229 @@ public class ConsumerConsumeTest
 
         Assert.Equal(2, messagesReceived);
     }
+
+    [Fact]
+    public async Task Consume_ephemeral_consumer_deleted_on_server_terminates_with_409()
+    {
+        await using var server = await NatsServerProcess.StartAsync();
+        await using var nats = new NatsConnection(new NatsOpts { Url = server.Url });
+        await nats.ConnectAsync();
+
+        var cts = new CancellationTokenSource(TimeSpan.FromSeconds(60));
+
+        var js = nats.CreateJetStreamContext();
+        await js.CreateStreamAsync(new StreamConfig("s1", ["s1.*"]), cts.Token);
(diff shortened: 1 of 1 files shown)


scottf (approved):
LGTM

scottf (commented):
## Claude PR Review: #1055 — "Add more tests for JetStream consumer behavior"

This is a test-only PR (295 additions, 0 deletions, 1 file) that adds 5 new test cases to `ConsumerConsumeTest.cs`, resolving issue #933 about ephemeral consumer test coverage. Already approved by scottf with LGTM.

### Overall assessment

This is solid test coverage for important edge cases in the JetStream consumer lifecycle. The tests cover a good matrix of failure scenarios: explicit consumer deletion (409), implicit consumer loss via server restart (503), slow processing interacting with both, and 503 accumulation across heartbeat-driven pulls. These are exactly the kinds of integration tests that catch real-world bugs.

### What's good

**Clear scenario differentiation.** Each test targets a distinct failure mode: the first test covers the "clean" 409 path where the server knows the consumer was deleted and tells you. The second covers the "dirty" path where the server restarts and has no memory of the consumer, so you get 503s instead. Tests 3 and 4 add the slow-processing dimension to each. Test 5 specifically validates the accumulation mechanism. This is well-thought-out coverage.

**Good use of `WaitSignal` for synchronization.** Rather than brittle `Task.Delay` for coordination between the consume loop and the test harness, the tests properly signal when the first message is consumed before proceeding with the destructive action (delete/restart). This is consistent with the existing test patterns in the file.

**Proper `Interlocked` usage** for cross-thread message counters. Consistent with the established patterns in the file.

**Each test starts its own `NatsServerProcess`** rather than using the shared fixture. This is necessary since these tests restart or need isolated server state, and it matches what the existing restart tests in the file do.

**The comments in test 4 are valuable** — the explanation about `ResetPending` setting pending to max and the heartbeat timer being the only mechanism to drive new pulls after 503 documents a subtle interaction that would otherwise require reading the implementation to understand.

### Issues

**Potential timing sensitivity in test 5 — `Retry.Until` after exception.** The test `await`s `consumeTask` (which throws), and *then* does a `Retry.Until` to check for idle timeout logs. But once `consumeTask` completes with an exception, no more log events will be generated. So either the logs are already present by the time the exception is caught, or they never will be. The `Retry.Until` with a 10-second timeout gives the impression it's waiting for something that might still happen, but the consume loop is already dead at that point. This won't cause a false *failure* (the logs should already be there), but the `Retry.Until` is misleading — a direct `Assert` would be clearer about what's actually being checked. Or if the intent is to allow for log flushing latency, a brief comment would help.

**Test 3 has a 5-second delay per message processed.** Combined with the 60-second CTS timeout, this is fine for correctness, but it makes this one test inherently slow. The 5-second delay is what creates the window for the delete to happen during processing, so it's doing real work — but a comment noting this is intentionally slow (and that the test validates the consumer terminates *despite* the slow handler) would help future maintainers who might see it and think it could be shortened.

**Inconsistent assertion style for 503 tests.** The 409 tests (tests 1 and 3) do thorough validation: assert `NatsJSProtocolException` type, check `HeaderCode == 409`, and check the message contains "Consumer Deleted". The 503 tests (tests 2, 4, 5) only check `Assert.Contains("503", exception.Message)`. It would be stronger to also assert on the exception type and/or header code for consistency, if the exception type carries that information for 503s. If it doesn't (because 503 is a no-responders scenario rather than a protocol status), a quick comment explaining the asymmetry would be useful.

**Test 4 doesn't assert on `messagesProcessed` bound.** It asserts `messagesProcessed > 0` (at least one was processed), but given 20 messages published and 500ms delay per message, plus a server restart happening after the first message, it might be worth asserting `messagesProcessed < 20` to confirm the consume loop actually terminated early rather than processing everything. This would validate the "terminated by 503 threshold" path rather than just "eventually finished."

### Minor observations

- All 5 tests use hardcoded stream name `"s1"` and consumer name `"c1"` without the `_server.GetNextId()` prefix pattern used by some other tests in the file. This is fine because each test starts its own isolated server, but it's a pattern difference worth noting.

- The consumer config in all tests sets `MemStorage = true` explicitly. For the 409 tests (1 and 3), memory storage isn't strictly necessary since the consumer is explicitly deleted. For the server-restart tests (2, 4, 5) it's essential since that's what makes the consumer vanish on restart. Not a problem, just noting it's consistently applied even where not required.

- Tests 2, 4, and 5 use `ConnectRetryAsync()` while tests 1 and 3 use `ConnectAsync()`. The retry variant is needed for the restart tests since the connection needs to survive the restart. Tests 1 and 3 don't restart the server so `ConnectAsync()` is fine. The distinction is correct.

### Verdict

This is good test coverage for a real gap identified in #933. The tests are well-structured, follow established patterns, and cover meaningful failure scenarios. The issues I noted are all minor polish — nothing that would cause incorrect test results. The approval from scottf seems appropriate. Good to merge.

PR #1058, opened 2026-02-02
Fix StreamSourceInfo.Active handle -1 correctly

### LINKED ISSUE #1045: StreamSourceInfo Active property gets parsed as zero rather than -1
### Observed behavior

The `Active` property of a `StreamSourceInfo` should be -1 in case there was no activity, as per the documentation. However, it looks like "-1" gets parsed to `TimeSpan.Zero`. I think it may be a rounding issue.

### Expected behavior

-1 should get parsed into a negative `TimeSpan` rather than `TimeSpan.Zero`, to disambiguate the case where `Active` is genuinely zero, albeit rare.

An alternative would be to make `Active` nullable and set it to `null` in case the value returned by the NATS server is -1. That may be preferable given it forces the user to take this case into account, and is more C#-like.

### Server and client version

Server: 2.12.3-alpine
Client: 2.7.1

### Host environment

_No response_

### Steps to reproduce

_No response_

mtmk:
resolves #1045

### DIFF
--- a/src/NATS.Client.JetStream/Models/StreamSourceInfo.cs
+++ b/src/NATS.Client.JetStream/Models/StreamSourceInfo.cs
@@ -45,12 +45,12 @@ public record StreamSourceInfo
     public long Lag { get; set; }
 
     /// <summary>
-    /// When last the mirror had activity, in nanoseconds. Value will be -1 when there has been no activity.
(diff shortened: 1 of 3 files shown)


scottf (approved):
LGTM

PR #1063, opened 2026-02-11
Add validation for unsupported PinnedClient calls

mtmk:
Reject pinned client priority policy with fetch and next

Per [ADR-42](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-42.md#pinning) pinned client mode only applies to Consume. Fetch and Next are not supported in this mode:

> That applies only to Consume. Fetch should not be supported in this mode. At least initially.


scottf (approved):
LGTM

PR #1064, opened 2026-02-13
POC: Add .NET 10 Target

mtmk:
**Minimum required changes to add .NET 10 target:**

- [X509Certificate2 constructors obsoleted (SYSLIB0057)](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057) — .NET 10 deprecates loading certificates via new
  X509Certificate2(...) and X509Certificate2Collection.Import(...). The replacement is X509CertificateLoader with methods
  like LoadPkcs12(), LoadPkcs12FromFile(), and LoadPkcs12CollectionFromFile().

- [Span null comparison (CA2265)](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2265) — .NET 10 flags comparing Span<T> / ReadOnlySpan<T> to null as an error, since null is
  implicitly converted to Span.Empty, making the check redundant.

- TFM-conditional package references — csproj conditions that matched specific TFMs (e.g. == 'net8.0') didn't account for
  net10.0, causing it to fall through to older package versions missing required APIs (keyed DI services).

TODO:
- [ ] Remove all .NET 6 conditionals
- [ ] Look into using new .NET 10 APIs e.g.
   - [ ] [System.Threading.Lock](https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock?view=net-10.0)
   - [ ] [JsonSerializer.DeserializeAsync(PipeReader)](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserializeasync?view=net-10.0#system-text-json-jsonserializer-deserializeasync(system-io-pipelines-pipereader-system-type-system-text-json-serialization-jsonserializercontext-system-threading-cancellationtoken))

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -2,14 +2,14 @@
     <Import Project="..\Directory.Build.props" />
 
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.0;netstandard2.1;net6.0;net8.0</TargetFrameworks>
+        <TargetFrameworks>netstandard2.0;netstandard2.1;net8.0;net10.0</TargetFrameworks>
         <ImplicitUsings>enable</ImplicitUsings>
         <Nullable>enable</Nullable>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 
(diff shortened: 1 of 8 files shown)


mtmk:
copied to #1070

PR #1076, opened 2026-02-25
Fix code analyzer warning

mtmk:
Code analyzers flagging the use of '`Finalize`' as method name. Renamed to '`FinalizeHash`' to avoid confusion.

This is an internal method.


scottf (approved):
LGTM

PR #1082, opened 2026-03-03
Add message context to serialization interfaces

mtmk:
Adds opt-in `INatsSerializeWithContext<T>` and `INatsDeserializeWithContext<T>` interfaces (and a combined `INatsSerializerWithContext<T>`) that receive a `NatsMsgContext` (subject, reply-to, headers) during (de)serialization. Existing serializers continue to work unchanged: the library dispatches via runtime type check and falls back to the standard interface otherwise. Built-in serializers (`NatsUtf8PrimitivesSerializer`, `NatsRawSerializer`, `NatsJsonContextSerializer`) also propagate the context to `_next` so chains keep context-awareness at the leaf. `NatsMsgContext` is constructed via a required-subject constructor (no init setters, no netstandard `IsExternalInit` shim). The read-only flag on `NatsHeaders` is removed so context-aware serializers can mutate headers (e.g. set `Content-Type`); the trade-off is that a single `NatsHeaders` instance is no longer safe to share across concurrent publishes, documented on the type.


src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -519,7 +519,7 @@ public NatsMsg<T> Msg
                 if (Serializer != null && Data != null)
                 {
                     var bufferWriter = new NatsPooledBufferWriter<byte>(SerializationBufferSize);
-                    Serializer.Serialize(bufferWriter, Data);
+                    Serializer.Serialize(bufferWriter, Data, null);
```
adminnz: Shouldn't this be passing in the Headers property?

mtmk: fixed.

mtmk:
Hey @adminnz, I added DIMs as well for BC. the idea is for new implementations on net8+ nothing changes. you only implement the new method. the old one is also marked as obsolete. little bit of maintenance burden on our library side but i think it's easily manageable. wdyt? if you're happy please let me know, I can merge (after fixing the tests etc.) and have it released in 3.0 previews today or tomorrow.

src/NATS.Client.Core/NatsHeaders.cs
```diff
@@ -15,7 +15,7 @@ namespace NATS.Client.Core;
 [SuppressMessage("StyleCop.CSharp.LayoutRules", "SA1516:Elements should be separated by blank line", Justification = "Keep class format as is for reference")]
 [SuppressMessage("StyleCop.CSharp.LayoutRules", "SA1513:Closing brace should be followed by blank line", Justification = "Keep class format as is for reference")]
 [SuppressMessage("StyleCop.CSharp.OrderingRules", "SA1214:Readonly fields should appear before non-readonly fields", Justification = "Keep class format as is for reference")]
-public class NatsHeaders : IDictionary<string, StringValues>
+public class NatsHeaders : INatsHeaders, IDictionary<string, StringValues>
```
adminnz: Remove IDctionary<string,StringValues> as INatsHeaders interface is already inheriting IDictionary<string,StringValues>

adminnz:
That is a novel solution for the backwards compatibility problem.
From my experience working on very large code bases (especially in a pub/sub environment) its best to have a really tight feedback loop for the developer, and identify problems as early on in the project/product lifecycle.
Problems can occur at many points in the lifecycle.

1. Compile Time
2. Unit Tests _(not all projects have tests)_
3. Run Time (App Start)
4. Run Time (Specific Condition)

While the DIM allows non-breaking changes to library users with existing INatsSerialize/INatsDeserialize implementations, it does mean that a developer now has to not only run their app, they have to trigger the exact code path (stage 4) to realize they might not have implemented either of the interface methods.

It is noble to try and never introduce breaking changes, the "major" in semantic versioning is:

> MAJOR version when you make incompatible API changes

Obviously that doesn't mean you have to introduce breaking changes, but it does provide the opportunity to introduce breaking changes.

In my opinion, it would be better to have 3.0 introduce a breaking change and require the header interface variants are implemented.

src/NATS.Client.Core/NatsHeaders.cs
mtmk: fixed

mtmk:
@adminnz thanks for the feedback! I've updated the approach; the old methods no longer have DIMs so any implementor missing them gets a compile error.

However the new header overloads still keep DIMs for BC.

On the semver point agreed that 3.0 is the right place for breaking changes but the main reason we want to keep the DIM on the new overloads is transitive dependencies: third-party libraries (or e.g. company internal common libraries) implementing `INatsSerialize<T>` would also need to release updates before users can upgrade to 3.0. Keeping the old methods as obsolete with DIMs on the new overloads gives the ecosystem a smoother migration path without sacrificing compile-time safety. I realize netstandard would still be breaking and that's something we have to accept balancing complexity, maintenance and given its usage is slowly getting less and less. Please also see the discussion here #1061

rickdotnet:
Your approach looks fine to me. I also prefer obvious breaking changes over random side-effects and major versions seem appropriate for that. Seems the shims are necessary.

I had started a take on this but never got it pushed. I went with [johnweldon's approach](https://github.com/nats-io/nats.net/discussions/1061#discussioncomment-15723006) and used separate interfaces to keep compatibility. Serializers can opt in if they want headers.

Here's what I had, might add some inspiration.

```cs
/// <summary>
/// Deserializer interface for NATS messages.
/// </summary>
/// <typeparam name="T">Deserialized object type</typeparam>
public interface INatsDeserialize<out T>
{
    T? Deserialize(in ReadOnlySequence<byte> buffer);
}

/// <summary>
/// Extended deserializer interface that supports receiving headers during deserialization.
/// </summary>
public interface INatsDeserializeWithHeaders<out T> : INatsDeserialize<T>
{
    /// <summary>
    /// Deserialize value from buffer with headers.
    /// </summary>
    /// <param name="buffer">Buffer with the serialized data.</param>
    /// <param name="headers">Headers from the message.</param>
    /// <returns>Deserialized object</returns>
    T? Deserialize(in ReadOnlySequence<byte> buffer, INatsHeaders headers);
}
```

Extension methods to prevent type-check spam.

```cs
public static class NatsSerializationExtensions
{
    /// <summary>
    /// Serializes the value with header support, falling back to standard serialization if headers are not supported.
    /// </summary>
    public static void Serialize<T>(this INatsSerialize<T> serializer, IBufferWriter<byte> bufferWriter, T value, INatsHeaders? headers)
    {
        if (serializer is INatsSerializeWithHeaders<T> withHeaders && headers != null)
        {
            withHeaders.Serialize(bufferWriter, value, headers);
            return;
        }

        // Fallback to standard serialization
        serializer.Serialize(bufferWriter, value);
    }

    /// <summary>
    /// Deserializes the value with header support, falling back to standard deserialization if headers are not supported.
    /// </summary>
    public static T? Deserialize<T>(this INatsDeserialize<T> deserializer, in ReadOnlySequence<byte> buffer, INatsHeaders? headers)
    {
        if (deserializer is INatsDeserializeWithHeaders<T> withHeaders)
            return withHeaders.Deserialize(buffer, headers ?? new NatsHeaders());

        return deserializer.Deserialize(buffer);
    }
}
```

mtmk:
@rickdotnet 
> ```cs
> INatsDeserializeWithHeaders
> ```

nice. this means we also cover netstandard as another first class target without breaking backward compatibility. cc @johnweldon

adminnz:
Yes that looks like a good solution.

scottf:
Here's my review of PR #1082 — header-aware serialization for nats.net.

---

## Summary

This PR adds opt-in header-aware serialization via two new interfaces (`INatsSerializeWithHeaders<T>`, `INatsDeserializeWithHeaders<T>`), a new `INatsHeaders` interface that `NatsHeaders` now implements, and extension methods for transparent dispatch with fallback. It's a well-scoped feature with good test coverage, but there are several issues worth raising.

---

## Design & Correctness Issues

### 1. Asymmetric null-headers dispatch (bug-level concern)

In `NatsSerializationExtensions`:

```csharp
// Serialize: null headers → skips header overload, calls standard
if (headers != null && serializer is INatsSerializeWithHeaders<T> withHeaders)

// Deserialize: null headers → STILL calls header overload
if (deserializer is INatsDeserializeWithHeaders<T> withHeaders)
    return withHeaders.Deserialize(buffer, headers); // headers can be null here
```

This asymmetry is inconsistent and creates a subtle bug in the pattern shown in the documentation example. In `MyHeaderAwareSerializer`, `Serialize` with headers writes a `Content-Type` header. But if headers happen to be null at publish time, the `Content-Type` is silently never written. Then on receive, `Deserialize` (with null headers) enters the header overload, finds no `Content-Type`, and **throws `NatsException`**.

Either: both overloads should dispatch the header variant whenever the interface is implemented (passing null through), or both should skip it when null. Given the semantics, the former is more consistent and lets implementors handle null however they wish.

### 2. Unnecessary and fragile cast in the doc example

In `SerializationPage.cs`:

```csharp
// Existing code:
if (headers is NatsHeaders mutableHeaders)
{
    mutableHeaders["Content-Type"] = "application/json";
}
```

`INatsHeaders` already extends `IDictionary<string, StringValues>`, which has a mutable indexer. The cast to `NatsHeaders` is unnecessary and silently does nothing if a caller ever passes a different `INatsHeaders` implementation. This should just be:

```csharp
if (headers != null)
    headers["Content-Type"] = "application/json";
```

### 3. Documentation claims DIMs that don't exist

The serialization docs say:

> On .NET 8+, you only need to implement one overload of `Serialize` and `Deserialize` thanks to default interface methods (DIMs).

But the diff contains no `default` interface method implementations anywhere. Implementors must provide all overloads on all targets. Either the DIMs need to be added, or this documentation needs to be corrected before merge.

### 4. Serialization order change in `CommandWriter.cs`

```csharp
// Before:  write headers → serialize payload
// After:   serialize payload → write headers
```

This is intentional (header-aware serializers need to mutate headers before they're written), and it's the right call. But it's a behavioral change for any existing code that might have relied on ordering side effects. Worth a comment in the code and/or a note in the PR description about why the order changed.

---

## Dependency Concern

`NATS.Client.Abstractions.csproj` now takes on a dependency on `Microsoft.Extensions.Primitives` to expose `StringValues` in `INatsHeaders`. This is a meaningful addition to what's meant to be a minimal abstractions package — it transitivley pulls that package into every consumer of abstractions. Worth a conversation about whether `INatsHeaders` belongs in abstractions vs. core, or whether it should use a simpler contract (e.g., `IReadOnlyDictionary<string, IEnumerable<string>>`).

---

## Minor / Nits

- **`INatsDeserialize<T>`** had its `/// <returns>` doc comment removed in the diff — likely accidental.
- **`NuidTests.cs`** timeout improvements (1s → 10s with `Assert.True`) are strictly correct and good, unrelated to the main feature.
- **ABI check program** exit code improvements are also good.
- The test's `HeaderAwareSerializer.CombineWith` returning `this` is probably fine for a test helper but worth a comment since it breaks chaining semantics.

---

## Overall

The core idea is sound and the extension method fallback pattern is a clean backward-compatible approach. The main things to address before merge are: the serialize/deserialize null-headers asymmetry (issue 1), the fragile cast in the docs example (issue 2), and the incorrect DIM documentation (issue 3).

scottf (changes requested):
Added notes from Claude

adminnz:
Just a quick question & suggestion, when is the expected ETA on version 3.0 (with this feature) being released?

Suggestion: Is there maybe a natural evolution/step that should be thought of with this Serialization & Deserialization.
Specifically this is about passing the payload (byte[]) & headers (INatsHeaders) into the deserializer, what about if the user of the library is needing the subject & payload & headers.

Some deserialization scenarios:

1. consumer is consuming a single well-known subject, deserialization is simply a straight payload -> type (NatsJsonSerializer<T>)

2. consumer is consuming multiple subjects (event.*) where event.foo & event.bar are consumed deserialization is dependant on the subject & payload.

3. consumer is consuming and requires access to subject & headers & payload to correctly deserialize.

mtmk:
> Just a quick question & suggestion, when is the expected ETA on version 3.0 (with this feature) being released?

very close. once this is merged I'll release shortly after.

> 
> Suggestion: Is there maybe a natural evolution/step that should be thought of with this Serialization & Deserialization. Specifically this is about passing the payload (byte[]) & headers (INatsHeaders) into the deserializer, what about if the user of the library is needing the subject & payload & headers.

actually subject and reply-to can be added too. good idea.

mtmk:
Based on @adminnz's suggestion I propose `NatsMsgContext` to be passed to serialization methods to provide access to headers as well as subject and optional reply-to:

```csharp
public readonly struct NatsMsgContext
{
    public string Subject { get; init; }
    public string? ReplyTo { get; init; }
    public INatsHeaders? Headers { get; init; }
}
```

Serialization methods:

```csharp
public interface INatsSerializeWithContext<in T> : INatsSerialize<T>
{
    void Serialize(IBufferWriter<byte> bufferWriter, T value, in NatsMsgContext context);
}

public interface INatsDeserializeWithContext<out T> : INatsDeserialize<T>
{
    T? Deserialize(in ReadOnlySequence<byte> buffer, in NatsMsgContext context);
}
```

adminnz:
The more i think about this, the more it seams like typed serialization/deserialization maybe should not be baked into the Nats.Client.Core library, and that the library should just be dealing with the low level (over the wire protocol) , i.e. NatsMsg that has no idea of types.

Then a higher level library Nats.Client.<something> should be providing the higher level Typed serialization and deserialization functionality.

Just as an example for publishing
```
public record Foo(string Name);
public record Bar(int Id, DateTime Timestamp);

async Task ExamplePublishAsync()
{
    await using var connection = new NatsConnection();
    var client = new NatsClient(connection);
    client.SendMsgFactory = new MyNatsSendMsgFactory();
    
    var foo = new Foo("bob");
    var headers = new NatsHeaders();
    headers["metadata"] = "something;
    await  client.PublishAsync(foo, headers: headers); //i could also just call with payload, or with subject, or with all the parameters
}

//prototype / scaffold code
class NatsClient
{
    public INatsSendMsgFactory SendMsgFactory {get;set;}
    public async Task PublishAsync(object payload, string? subject = null, NatsHeaders? headers = null, CancellationToken cancellationToken = default)
    {
          var context = new SendContext (payload, subject, headers);
          INatsMsg msg = SendMsgFactory.Create(context)
          await connection.PublishAsync(msg, cancellationToken);
    }
}

class MyNatsSendMsgFactory : INatsSendMsgFactory
{
    public INatsMsg Create(SendContext context)
    {
       context.Headers["content-type"] = "application/json";
       context.Headers["message-type"] = context.Payload.GetType().Name;
       if (context.Payload is Foo foo)
       {
             context.Subject = $"foo.{foo.Name}";
       }
       else if (context.Payload is Bar bar)
       {
            context.Subject = $"bar.{bar.Id}.something";
       }
       var serializedPayload = JsonSerializer.Serialize(context.Payload);
       return new NatsMsg(context.Subject, context.Headers, payload: serializedPayload);
    }
}
```

rickdotnet (approved):
Looks good to me.

Should serializers be able to produce/add headers during serialization? Looks like this supports header-aware deserialization, not header-producing serialization. Seems appropriate to me unless there is a use-case worth tackling now.

@adminnz re: typed vs bytes - valid architectural discussion for a future layer, but this PR doesn't block that path.

Would something like this work?
```csharp
public class CustomClient
{
    private readonly INatsConnection _conn;
    public INatsSendMsgFactory Factory { get; set; }
    
    public async Task PublishAsync(object payload, NatsHeaders? headers = null)
    {
        var msg = Factory.Create(payload, headers);  // Subject from convention
        await _conn.PublishAsync(msg.Subject, msg.Payload, msg.Headers);
    }
}
```

scottf (approved):
LGTM

adminnz:
> Looks good to me.
> 
> Should serializers be able to produce/add headers during serialization? Looks like this supports header-aware deserialization, not header-producing serialization. Seems appropriate to me unless there is a use-case worth tackling now.
> 
> @adminnz re: typed vs bytes - valid architectural discussion for a future layer, but this PR doesn't block that path.
> 
> Would something like this work?
> 
> ```cs
> public class CustomClient
> {
>     private readonly INatsConnection _conn;
>     public INatsSendMsgFactory Factory { get; set; }
>     
>     public async Task PublishAsync(object payload, NatsHeaders? headers = null)
>     {
>         var msg = Factory.Create(payload, headers);  // Subject from convention
>         await _conn.PublishAsync(msg.Subject, msg.Payload, msg.Headers);
>     }
> }
> ```

Serialization (producing) should definitely be able to add to headers, otherwise sending is seriously gimped.

And yes a higher level client can the provide very easy to use apis

mtmk:
@adminnz re separating typed serialization from core: I agree a higher-level client with a factory pattern is a useful abstraction and nothing in here prevents building that on top of `INatsConnection`. The reason serialization is baked into core is zero-copy performance. Serializers write directly into a pooled `IBufferWriter<byte>` that goes straight onto the wire, no intermediate allocation. Your example materializes the payload as a `string`/`byte` array before handing it to `NatsMsg` which is an extra allocation + copy. you could change the factory to use an `IBufferWriter`, but then we're basically back to what current serializers does. Happy to explore a higher-level typed client in a separate issue if you want to open one.

Re header mutating serialization: good catch, it doesn't actually work yet. headers get locked to read-only before it reaches the serializer. the fix is to move `SetReadOnly()` after serialization. That said, the read-only flag was always a weak guarantee since headers are copied to a byte buffer synchronously regardless. Worth considering removing it entirely. @rickdotnet wdyt?

mtmk:
Went ahead and removed the read-only flag entirely in 3a54d03 rather than just moving `SetReadOnly()`, since it was protecting against mutations that can't reach the wire anyway (headers get copied to a byte buffer synchronously inside `CommandWriter`). Added a test in `SerializerTest.cs` to verify serializers can mutate `context.Headers` (e.g. set `Content-Type`) during serialization. @rickdotnet please yell if you see any issue with this.

PR #1084, opened 2026-03-05
Fix KV watcher cancellation behavior

mtmk:
This ws also caught by test Watcher_cancellation_no_warn_logs flapping.


src/NATS.Client.KeyValueStore/NatsKVStore.cs
```diff
@@ -431,51 +431,62 @@ public IAsyncEnumerable<NatsKVEntry<T>> WatchAsync<T>(string key, INatsDeseriali
...
+                        break;
+                    }
+                    catch (TaskCanceledException)
+                    {
+                        break;
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        break;
+                    }
```
scottf, line 471: Is there a way to combine these catch blocks since they all do the same thing?
Java has something like this, seems like .NET would
```
catch (ChannelClosedException | TaskCanceledException | OperationCanceledException)
```

scottf (approved):
LGTM

scottf (approved):
LGTM

PR #1085, opened 2026-03-06
Fix message loss on consumer dispose

mtmk:
JetStream consumer dispose was dropping in-flight messages. The per-sub path completed the channel before all wire messages had landed, and connection dispose killed the socket before subs got a chance to drain, leaving NumAckPending pinned until AckWait. NatsSubBase now has a DrainAsync (UNSUB -> PING/PONG -> TryComplete) that per-sub dispose calls to flush the wire cleanly. On connection dispose, drain participants run in parallel so the wait is one round-trip total, not multiplied by the number of subs.

New NatsOpts:

- `DrainSubscriptionsOnDispose` (default `false`) - opt-in for the dispose reorder.
- `ConsumerDrainOnDisposeTimeout` (default `null`) - wait for active consume loops to finish acking.
- `DrainPingTimeout` (default `5s`) - per-sub UNSUB ack budget.

Defaults preserve current behavior.


mtmk:
@claude fixed your comments, review again

scottf (approved):
LGTM

PR #1094, opened 2026-03-19
tls: document Prefer mode plaintext behavior

mtmk:
Documents TLS Prefer/Auto mode behavior and adds guidance for TLS-terminating proxy setups (nats-io/nats-server#7927).

Unlike most other NATS clients (Go, Rust, Java, Python), the .NET client attempts a TLS upgrade when the server advertises `tls_available` without `tls_required`. This can cause connection failures behind a TLS-terminating proxy. The workaround is `TlsMode.Disable`.

We are not changing the default behavior to match other clients because existing configurations may rely on the automatic TLS upgrade. Switching to plaintext silently would be a security regression for those users.

- XML doc remarks on NatsOpts.TlsOpts, NatsTlsOpts.Mode, TlsMode.Prefer, TlsMode.Auto
- TLS proxy section in docfx security page
- Real nats-server tests with `tls {}` + `allow_non_tls`
- Go client cross-check via GoHarness
- Fake-server tests for Prefer, Disable, Require, Auto modes

### DIFF
--- a/.github/workflows/test_linux.yml
+++ b/.github/workflows/test_linux.yml
@@ -25,7 +25,9 @@ jobs:
     steps:
       - name: Set up Go
         uses: actions/setup-go@v5
-        with: { cache: true }
+        with:
+          go-version: 'stable'
+          cache: true
 
       - name: Install nats-server
         run: go install github.com/nats-io/nats-server/v2@${{ matrix.config.branch }}
--- a/.github/workflows/test_linux_core.yml
+++ b/.github/workflows/test_linux_core.yml
@@ -25,7 +25,9 @@ jobs:
     steps:
       - name: Set up Go
         uses: actions/setup-go@v5
-        with: { cache: true }
+        with:
+          go-version: 'stable'
+          cache: true
 
       - name: Install nats-server
         run: go install github.com/nats-io/nats-server/v2@${{ matrix.config.branch }}
--- a/.github/workflows/test_windows.yml
+++ b/.github/workflows/test_windows.yml
@@ -40,7 +40,9 @@ jobs:
 
       - name: Set up Go
         uses: actions/setup-go@v5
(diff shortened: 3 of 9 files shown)


philpennock (approved):
Did not review tests.

If it is not too late to rename `Implicit`, it would be good to do so, as "implicit" might be taken as "ambiently available without our seeing it", via things like stunnel or equivalent cryptographic privacy (tailscale etc).  I'd name it `OnConnect` and make clear that it is a stronger form of `Require`.  I'd also mention explicitly below the table that only `Require` and (`OnConnect` or `Implicit`) provide the full protection which TLS can offer.

Looks good as is, the feedback above is non-blocking.

mtmk:
> Did not review tests.
> 
> If it is not too late to rename `Implicit`, it would be good to do so, as "implicit" might be taken as "ambiently available without our seeing it", via things like stunnel or equivalent cryptographic privacy (tailscale etc). I'd name it `OnConnect` and make clear that it is a stronger form of `Require`. I'd also mention explicitly below the table that only `Require` and (`OnConnect` or `Implicit`) provide the full protection which TLS can offer.
> 
> Looks good as is, the feedback above is non-blocking.

@philpennock thanks for the review! good point. We can deprecate `Implicit`. `OnConncet` would clash with .NET event naming convention though. Perhaps `TlsFirst` matching server? other options: `Immediate`, `Preemptive`, `Eager`. Any preference, suggestion?

Will also add the note about full TLS protection.

philpennock:
I retract my previous suggestion to avoid "Implicit". That's the RFC 8314 terminology so we should stick with it for consistency even if it is a little unfortunate.

mtmk:
test failures for 'Check_JetStream_cluster_related_fields' are not related to this PR. We can investigate after merge to main.

PR #1096, opened 2026-03-23
Fix Range attribute for MaxBytes property

partnerRuiSilva:
# Overview
- Fix range MaxBytes property in StreamConfig


mtmk (approved):
LGTM

mtmk:
thanks @partnerRuiSilva

PR #1099, opened 2026-04-01
Improve Pin ID handling

### LINKED ISSUE #1098: JetStream consume: pinned client waits for heartbeat timeout after 423 Pin ID Mismatch instead of immediately re-pulling
## Summary

  `NatsJSConsume` currently does not immediately re-issue a pull after receiving a `423 Pin ID Mismatch` status response.

  Instead, it clears the pin ID and calls `ResetPending()`, which prevents the subsequent `CheckPending()` call from scheduling another pull. In practice, recovery is then delayed until the next heartbeat timeout.

  This causes visible latency spikes for pinned pull consumers even when the connection remains healthy.

  ## Observed Behavior

  In production, we see this sequence repeatedly on a single-member pinned consumer:

  ```text
  Sending pull request for pull-loop(chk-msgs) 100, 0
  Pull request issued for pull-loop(chk-msgs) 100, 0
  Pin ID Mismatch
  ...
  Sending pull request for pull-loop(heartbeat-timeout) 100, 0
  Idle heartbeat timeout after 00:00:01.5000000ns
  Pull request issued for pull-loop(heartbeat-timeout) 100, 0

  Important detail: this is not caused by reconnects.

  On the consumer side:

  - connection state remains Open
  - there are no ConnectionDisconnected, ConnectionOpened, or ReconnectFailed events
  - messages resume only after the heartbeat-timeout pull

  We also observed the same pattern repeatedly:

  1. normal messages are received with a valid Nats-Pin-Id
  2. a refill pull (pull-loop(chk-msgs)) gets 423 Pin ID Mismatch
  3. no immediate re-pull happens
  4. recovery waits until heartbeat timeout

  ## Why this looks wrong

  Current nats.net code in NatsJSConsume.ReceiveInternalAsync() does this for 423:

  else if (headers.Code == 423)
  {
      _logger.LogDebug(NatsJSLogEvents.PinIdMismatch, "Pin ID Mismatch");
      NatsJSExtensionsInternal.HandlePinIdMismatch(_jsConsumer, _notificationChannel);
      ResetPending();
  }

  But CheckPending() runs immediately afterwards, and because pending was reset to full capacity, it does not enqueue a new pull.

  That means the client waits for heartbeat timeout instead of recovering immediately.

  ## Comparison with nats.go

  The Go client appears to handle this differently.

  In nats.go/jetstream/pull.go, the ErrPinIDMismatch path clears the pinned ID and zeroes pending counts, so the later pending check immediately schedules another pull instead of waiting for heartbeat timeout.

  That behavior matches what I would expect here as well.


  ## Actual Behavior

  After 423 Pin ID Mismatch:

  1. pin ID is cleared
  2. pending is reset
  3. no immediate pull is sent
  4. recovery happens only after heartbeat timeout

I'm still investigating why we are seeing PinMismatch from server. but the recovery could be improved

### Expected behavior

  ## Expected Behavior

  After a 423 Pin ID Mismatch response in consume mode:

  1. clear the current pin ID
  2. mark pending as exhausted (or otherwise trigger immediate pending re-check)
  3. issue the next pull immediately
  4. do not wait for heartbeat timeout to recover

### Server and client version

Nats-server 2.12.6
Ntas.net 2.73.0

### Host environment

_No response_

### Steps to reproduce

_No response_

colprog:
Always use latest pin id  to avoid Pin Mismatch Error.
Also improve error recovery in the face of said error

Fixes #1098

### DIFF
--- a/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
@@ -540,7 +540,6 @@ internal class NatsJSConsume<TMsg> : NatsSubBase
             MinPending = _priorityGroup?.MinPending ?? 0,
             MinAckPending = _priorityGroup?.MinAckPending ?? 0,
             Priority = _priorityGroup?.Priority ?? 0,
-            Id = _jsConsumer?.GetPinId(),
         },
         Origin = origin,
     });
@@ -559,6 +558,7 @@ internal class NatsJSConsume<TMsg> : NatsSubBase
             var origin = $"pull-loop({pr.Origin})";
             try
             {
+                pr.Request.Id = _jsConsumer?.GetPinId();
                 await CallMsgNextAsync(origin, pr.Request).ConfigureAwait(false);
             }
             catch (NatsConnectionFailedException)
--- a/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
@@ -386,7 +386,7 @@ internal class NatsJSConsume<TMsg> : NatsSubBase
                     {
                         _logger.LogDebug(NatsJSLogEvents.PinIdMismatch, "Pin ID Mismatch");
                         NatsJSExtensionsInternal.HandlePinIdMismatch(_jsConsumer, _notificationChannel);
-                        ResetPending();
+                        ClearPending();
                     }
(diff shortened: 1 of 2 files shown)


mtmk (approved):
LGTM thanks @colprog

mtmk:
ah, forgot! @colprog you need to sign your commits please.

colprog:
signed and force pushed.

PR #1100, opened 2026-04-02
Fix NuidTests flaky timeouts

mtmk:
Increase thread join timeouts from 1s to 10s and assert the result with a diagnostic message. The short timeouts caused flaky failures on slow CI runners.

- [x] CI green


scottf (approved):
LGTM

PR #1101, opened 2026-04-08
auth: use NATS.NKeys package for nkey signing

mtmk:
Replace the in-tree Ed25519/NKey implementation with the [NATS.NKeys package](https://www.nuget.org/packages/NATS.NKeys), dropping the vendored NaCl reference implementation and the partial public `NKeys`/`NKeyPair` API. Signing now goes through `NATS.NKeys.KeyPair` and produces byte-for-byte identical signatures.

**BREAKING** Note this removes the public `NATS.Client.Core.NKeys` and `NKeyPair` types: external callers must switch to `NATS.NKeys.KeyPair`.


scottf (approved):
LGTM

PR #1106, opened 2026-04-10
Fix release workflow

mtmk:
The release workflow triggered on push to main, so any merged PR that changed version.txt would push packages to NuGet. Switch to the `release: published` event so publishing requires a maintainer to create a GitHub release, and validate the tag matches version.txt before packing.


Jarema (approved):
LGTM!

PR #1107, opened 2026-04-13
Reset state on credential loading failure

### LINKED ISSUE #1104: Second ConnectAsync call stuck if CredsFile doesn't exist during first call
### Observed behavior

If I create an instance of NatsClients with credsFile path where the file doesn't exist, it throws an exception and the second call stucks forever no matter if the file already exists or not. Current workaround is to use AuthCredCallback. 

### Expected behavior

It shouldn't stuck and it should throw or connect to the server depending if the file exits or not. 

### Server and client version

Server: 2.11.9
Client 2.7.3, tested also with 3.0.0-prev2

### Host environment

_No response_

### Steps to reproduce

Minimal example with possible workaround: 
```
            //var natsClient = new NatsClient(new NatsOpts
            //{
            //    Url = "nats://localhost:4222",
            //    AuthOpts = new NatsAuthOpts
            //    {
            //        AuthCredCallback = async (uri, ct) =>
            //    {
            //        return NatsAuthCred.FromCredsFile("C:\\temp\\nats.creds");
            //    }
            //    }
            //});

            var natsClient = new NatsClient("nats://localhost:4222", credsFile: "C:\\temp\\nats.creds");

            try
            {
                await natsClient.ConnectAsync();
            }
            catch (Exception ex)
            {
                Console.WriteLine($"Connection failed: {ex.Message}");
            }

            try
            {
                await natsClient.ConnectAsync();
            }
            catch (Exception ex)
            {
                Console.WriteLine($"Connection failed: {ex.Message}");
            }
```

mtmk:
When ConnectAsync throws during credential loading (e.g. missing creds file), ConnectionState stays at Connecting and the waiter TCS is never completed. A second ConnectAsync call then hangs forever waiting on that TCS.

Wrap credential loading in InitialConnectAsync with a try-catch that resets ConnectionState to Closed and faults the waiter, matching the existing pattern for socket and setup failures. Also broaden the catch in LoadCredsFile from NatsException to Exception so IO errors like FileNotFoundException are wrapped properly.

fixes #1104

### DIFF
--- a/src/NATS.Client.Core/Internal/UserCredentials.cs
+++ b/src/NATS.Client.Core/Internal/UserCredentials.cs
@@ -159,7 +159,7 @@ internal class UserCredentials
             using var reader = new StreamReader(path);
             return ParseCreds(reader);
         }
-        catch (NatsException e)
+        catch (Exception e)
         {
             throw new NatsException($"Error loading creds file '{path}': {e.Message}", e);
         }
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -394,9 +394,26 @@ public partial class NatsConnection : INatsConnection
                 throw new NatsException($"URI {uri} requires TLS but TlsMode is set to Disable");
         }
 
-        if (!Opts.AuthOpts.IsAnonymous)
+        try
         {
-            _userCredentials = new UserCredentials(Opts.AuthOpts);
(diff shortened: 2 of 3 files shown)


mtmk:
hey @Prochy are you able to confirm this fix is working for you?

Prochy:
> hey @Prochy are you able to confirm this fix is working for you?

Just tested, it works.

scottf (approved):
LGTM

PR #1111, opened 2026-04-16
Fix net481 TLS test flakes

mtmk:
Two net481-only TLS test flakes caused by the same root cause: on net481 under load, the client can tear down the TCP socket (RST) before the ClientHello is observed at the server, so the usual synchronous `AuthenticationException` path isn't what the assertions see.

`Tls_fails_without_certificates` (Platform.Windows.Tests): the strict `exception.InnerException?.Message` regex compared against `null` when the cause was anywhere but the first inner slot. Walk the exception chain and accept any of `AuthenticationException`, `SocketException`, or `IOException` as evidence the TLS upgrade was rejected.

`Prefer_mode_attempts_tls_upgrade_when_server_advertises_tls_available` (Core2.Tests): the assertion relied on the server reading the ClientHello byte (0x16) first, but on the RST race the server sometimes reads 0 bytes. Fail only when the server observes a plaintext CONNECT (the actual bug this test guards against); otherwise accept either the 0x16 byte or a TLS-related exception chain from `ConnectAsync` as evidence the upgrade was attempted.

### DIFF
--- a/tests/NATS.Client.Core2.Tests/TlsPreferTest.cs
+++ b/tests/NATS.Client.Core2.Tests/TlsPreferTest.cs
@@ -1,5 +1,6 @@
 using System.Net;
 using System.Net.Sockets;
+using System.Security.Authentication;
 using System.Text;
 using Synadia.Orbit.Testing.GoHarness;
 using Synadia.Orbit.Testing.NatsServerProcessManager;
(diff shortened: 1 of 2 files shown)


Jarema (approved):
LGTM

PR #1115, opened 2026-04-20
Fix slow-consumer test first-ping RTT flap

mtmk:
The first measured ping in `JetStream_consume_slow_consumer_should_not_block_connection` occasionally clocks ~1000ms on Windows CI (against a `< 1000ms` threshold), while pings 2 through 20 come in under 0.25ms. Send one untimed ping after consume starts so the measurement loop excludes that cold-start cost. The slow-consumer-should-not-block invariant is preserved.


scottf (approved):
LGTM

PR #1116, opened 2026-04-20
Make pinned client Pin ID handling tolerant of duplicate headers

colprog:
This PR makes pinned client Pin ID handling more robust when multiple Nats-Pin-Id header values are present.

  This was discovered while testing an echo service. The echo service returns incoming headers as-is, so if a request already contains a Nats-Pin-Id the client can observe multiple Nats-Pin-Id values in the same header collection.

  The previous code was not necessarily wrong for the normal case, but it assumed Nats-Pin-Id behaved as a single-value header. With duplicate values, converting the full StringValues to string could use a combined value instead of the current pin id.

  This change updates the client to use the last Nats-Pin-Id value when multiple values are present, matching the latest value appended by the server and making pinned client handling more defensive.

  Changes

  - Use TryGetLastValue("Nats-Pin-Id", out var pinId) when updating the consumer pin id.
  - Add a regression test covering duplicate Nats-Pin-Id headers.
  - Keep existing behavior unchanged for the common single-header case.

  Why
  This is primarily a robustness improvement rather than a fix for the standard path. It protects the client from stale pin ids when server responses preserve previous headers and append a newer Nats-Pin-Id.

  Testing

  - Added coverage for multiple Nats-Pin-Id values and verified that the latest value is used.

### DIFF
--- a/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
@@ -173,7 +173,16 @@ internal class NatsJSConsume<TMsg> : NatsSubBase
 
         if (_debug)
         {
-            _logger.LogDebug(NatsJSLogEvents.PullRequest, "Sending pull request for {Origin} {Msgs}, {Bytes}", origin, request.Batch, request.MaxBytes);
+            _logger.LogDebug(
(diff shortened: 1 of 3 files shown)


mtmk (approved):
LGTM thanks @colprog

PR #1117, opened 2026-04-29
Use System.Threading.Lock on NET9_0_OR_GREATER

mtmk:
Replace `private readonly object` lock fields with `System.Threading.Lock` under `#if NET9_0_OR_GREATER`. On .NET 9+, the C# compiler emits `Lock.EnterScope()` calls instead of `Monitor.Enter`/`Exit`, avoiding the object-header overhead. The `lock (x)` syntax is unchanged.

Nine lock fields updated across core (CommandWriter, NatsConnection, NatsSubBase, SubscriptionManager, ObjectPool, ReplyTask) and JetStream (NatsJSConsume, NatsJSOrderedConsume, NatsJSConsumer). Downlevel TFMs fall through to the existing `object` path.


mtmk:
Reopening with correct base branch.

PR #1118, opened 2026-04-29
Use System.Threading.Lock on NET9_0_OR_GREATER

mtmk:
Replace \`private readonly object\` lock fields with \`System.Threading.Lock\` under \`#if NET9_0_OR_GREATER\`. On .NET 9+, the C# compiler emits \`Lock.EnterScope()\` calls instead of \`Monitor.Enter\`/\`Exit\`, avoiding the object-header overhead. The \`lock (x)\` syntax is unchanged.

Nine lock fields updated across core (CommandWriter, NatsConnection, NatsSubBase, SubscriptionManager, ObjectPool, ReplyTask) and JetStream (NatsJSConsume, NatsJSOrderedConsume, NatsJSConsumer). Downlevel TFMs fall through to the existing \`object\` path.

### DIFF
--- a/sandbox/MicroBenchmark/LockBench.cs
+++ b/sandbox/MicroBenchmark/LockBench.cs
@@ -1,0 +1,61 @@
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Jobs;
+
+namespace MicroBenchmark;
+
+// Compares object lock (all TFMs) vs System.Threading.Lock (NET9+).
+// On net8.0 both methods use object, so they serve as a same-TFM baseline.
+// On net10.0 SystemLock uses System.Threading.Lock whose EnterScope() path
+// avoids the Monitor sync-block machinery.
+[SimpleJob(RuntimeMoniker.Net80)]
+[SimpleJob(RuntimeMoniker.Net10_0)]
+[MemoryDiagnoser]
+public class LockBench
+{
+    private readonly object _objectLock = new();
+
+#if NET9_0_OR_GREATER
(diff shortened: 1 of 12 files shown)


mtmk:
Benchmarks run on this branch (BenchmarkDotNet, Release build, Windows):

**Raw lock microbench** (`LockBench`):

| Method | Runtime | Mean | Ratio |
|---|---|---|---|
| ObjectLock_Empty | .NET 8.0 | 20.31 ns | 1.00 |
| SystemLock_Empty | .NET 8.0 | 20.32 ns | 1.00 |
| ObjectLock_Empty | .NET 10.0 | 20.42 ns | 1.00 |
| SystemLock_Empty | .NET 10.0 | 17.87 ns | 0.88 |

On .NET 8 both rows use `object` (the `#if` guard is inactive), so they're identical -- expected. On .NET 10 `System.Threading.Lock` saves ~2.5 ns per acquire/release, about 12% on the raw operation.

**Serial publish throughput** (`PublishRuntimeBench`, 100k msgs, 128-byte payload):

| Runtime | Mean |
|---|---|
| .NET 8.0 | 35.35 ms |
| .NET 10.0 | 32.32 ms |

~9% faster on .NET 10 overall. The lock change contributes a fraction of that -- at ~3M msg/s the 2.5 ns saving per lock cycle adds up to well under 1% of total publish time. The bulk of the runtime delta is JIT and pipeline improvements.

MauriceVanVeen (approved):
LGTM

PR #1120, opened 2026-04-30
Add AllowBatchPublish stream config field

mtmk:
Adds the `AllowBatchPublish` field on `StreamConfig` (JSON `allow_batched`), required by streams that opt in to fast ingest batch publishing per ADR-50. Matches the field added in nats.go (#2052) and nats.js (#379). Server-side support requires nats-server v2.14.0+. The fast-ingest publisher itself will live in orbit.net alongside the existing atomic batch publisher.


MauriceVanVeen (approved):
LGTM

PR #1128, opened 2026-05-06
Add consumer field on stream source/mirror

mtmk:
Adds `Consumer` to `StreamSource` so the server uses a pre-created durable consumer to drive mirror or source replication, plus the `flow_control` ack policy that those consumers require. Available on server 2.14+.


scottf (approved):
LGTM

PR #1131, opened 2026-05-07
Migrate solution to slnx format

mtmk:
Replace `NATS.Net.sln` with `NATS.Net.slnx` (generated via `dotnet sln migrate`). Update the few references to the old filename in `README.md`, `scripts/apicompat.sh`, and `tools/Schema.Generation/Program.cs` (the latter was pointing at a non-existent `NATS.Client.sln`). Rename `NATS.Client.sln.DotSettings` to `NATS.Net.slnx.DotSettings` so Rider keeps loading the shared abbreviations and dictionary entries.


scottf (approved):
LGTM

PR #1132, opened 2026-05-08
Fix unobserved task exception during dispose

### LINKED ISSUE #1124: UnobservedTaskException from orphaned reconnect task and async void ReconnectLoop after DisposeAsync
### Observed behavior

After disposing a NatsConnection, an UnobservedTaskException is thrown by the finalizer thread.

When DisposeAsync is called, it closes the socket and unblocks WaitForClosed. This causes the background ReconnectLoop to wake up and attempt to reconnect.

During this reconnect attempt in SetupReaderWriterAsync(true) (line 594 (https://github.com/nats-io/nats.net/blob/v2.7.3/src/NATS.Client.Core/NatsConnection.cs#L594)), a reconnectTask is spawned in the
background to reestablish subscriptions. The method then waits for the PONG signal. Because the connection is being torn down, this wait times out, throwing a TimeoutException.

This timeout abandons the reconnectTask without awaiting it. When the abandoned task fails due to the closed socket, its exception goes unobserved. Furthermore, because ReconnectLoop is declared as async void
and spawned as fire-and-forget, DisposeAsync has no mechanism to cleanly observe or shut down the loop.

The finalizer thread eventually throws:
System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (can not start to connect nats server: nats://nats:4222)
 ---> NATS.Client.Core.NatsException: can not start to connect nats server: nats://nats:4222
 ---> System.TimeoutException: The operation has timed out.
  at NATS.Client.Core.NatsConnection.SetupReaderWriterAsync(Boolean reconnect)
  at NATS.Client.Core.NatsConnection.ReconnectLoop()
  at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
  at System.Threading.QueueUserWorkItemCallback.Execute()

### Expected behavior

Disposing the connection should not result in UnobservedTaskExceptions crashing the finalizer thread.

The expected fix requires two parts:

1. Prevent the orphaned task leak in SetupReaderWriterAsync:
Ensure reconnectTask is awaited or cleanly handled even if the PONG wait times out.
```
try
{
    await waitForPongOrErrorSignal.Task.WaitAsync(Opts.ConnectTimeout).ConfigureAwait(false);
}
catch (TimeoutException)
{
    _logger.LogDebug(NatsLogEvents.Connection, "Timeout waiting for initial pong");
    throw;
}
finally
{
    if (reconnectTask != null)
    {
        try { await reconnectTask.ConfigureAwait(false); } catch { /* ignore/log */ }
    }
}
```
2. Fix the ReconnectLoop signature and await it during disposal:
Change ReconnectLoop to return a Task so its lifecycle can be managed.

```
// Change signature
private async Task ReconnectLoop()  // was: private async void ReconnectLoop()

// In DisposeAsync, safely await the task
if (_reconnectLoopTask is not null)
{
    try { await _reconnectLoopTask.ConfigureAwait(false); }
    catch { /* already disposing */ }
}
```

### Server and client version

- NATS.Client.Core version: 2.7.3 (also verified in 2.6.11 and 3.0.0-preview.4)
- NATS Server version: N/A (Client-side lifecycle issue)

### Host environment

_No response_

### Steps to reproduce

```
var opts = new NatsOpts { Url = "nats://localhost:4222" };
var connection = new NatsConnection(opts);
// Connection succeeds, ReconnectLoop is spawned in the background

await connection.DisposeAsync();
// Socket closes, ReconnectLoop wakes up, tries to reconnect, times out.
// reconnectTask is abandoned and its exception goes unobserved.

GC.Collect();
// UnobservedTaskException fires here
```

mtmk:
Reported by @xwang-dev: disposing a `NatsConnection` could leak an `UnobservedTaskException` to the finalizer thread. Two related issues addressed: (1) the `reconnectTask` spawned inside `SetupReaderWriterAsync` was abandoned when the PONG wait threw, so any later fault went unobserved; (2) `ReconnectLoop` was `async void`, so its lifecycle could not be awaited from `DisposeAsync`. Now the orphan task is observed in a catch path before rethrow, the loop is `async Task ReconnectLoopAsync`, and `DisposeAsync` awaits `_reconnectLoopTask` and swallows-with-log.

Fixes #1124

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -321,6 +321,21 @@ public partial class NatsConnection : INatsConnection
             _disposedCts.Cancel();
             _initialConnectCts.Cancel();
 #endif
+
+            // Drain the reconnect loop so any exception it raises is observed here
+            // rather than leaking onto the finalizer thread.
+            var reconnectLoopTask = _reconnectLoopTask;
+            if (reconnectLoopTask is not null)
+            {
+                try
+                {
+                    await reconnectLoopTask.ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogDebug(NatsLogEvents.Connection, ex, "Reconnect loop completed with exception during dispose");
+                }
+            }
         }
     }
 
@@ -513,7 +528,7 @@ public partial class NatsConnection : INatsConnection
 #pragma warning disable VSTHRD103
             _initialConnectCts.Cancel();
 #pragma warning restore VSTHRD103
-            _reconnectLoopTask = Task.Run(ReconnectLoop);
+            _reconnectLoopTask = Task.Run(ReconnectLoopAsync);
             _eventChannel.Writer.TryWrite((NatsEvent.ConnectionOpened, new NatsEventArgs(url?.ToString() ?? string.Empty)));
         }
     }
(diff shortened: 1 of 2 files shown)


mtmk:
see the note on the issue #1124 as well. I'm not able to reproduce this error. it's most likely found through static analysis by an LLM. waiting for confirmation the user.

PR #1134, opened 2026-05-12
Fix subscription timeout overflow

### LINKED ISSUE #1125: If NatsSubOpts.Timeout is greater than 4294967294 it throws ArgumentOutOfRangeException
### Observed behavior

Not sure, if this is a bug or not but if I set NatsSubOpts with Timeout = TimeSpan.MaxValue, it throws :
 ```
'dueTime ('922337203685477') must be less than or equal to '4294967294'. (Parameter 'dueTime')
Actual value was 922337203685477.'
```

It's because of limitation in System.Threading.Timer.Change method https://learn.microsoft.com/en-us/dotnet/api/system.threading.timer.change?view=net-10.0

### Expected behavior

I would expect if I provide TimeSpan.MaxValue, it should not throw and take the highest value possible in this case 4294967294.

If it's not a bug, it should be at least documented.

### Server and client version

Server Version = 2.11.9
Client Version = v2.7.3

### Host environment

_No response_

### Steps to reproduce

```
using NATS.Net;

var natsClient = new NatsClient("nats://localhost:4222");

await natsClient.ConnectAsync();

try
{
    await natsClient.RequestAsync<object>(
        "subject",
        replySerializer: null,
        replyOpts: new NATS.Client.Core.NatsSubOpts { Timeout = TimeSpan.MaxValue });
}
catch (Exception)
{
    // Handle the exception (e.g., log it, retry, etc.)
}

```

### PR DESCRIPTION
**Pull request by mtmk**
`NatsSubOpts.Timeout = TimeSpan.MaxValue` (and the matching idle/start-up fields, plus the direct request-reply path) threw `ArgumentOutOfRangeException` because `System.Threading.Timer.Change` and `Task.WaitAsync` reject dueTimes above `uint.MaxValue - 1` ms (~49.7 days). `TimeSpan.MaxValue` and `Timeout.InfiniteTimeSpan` are now accepted as explicit "no timeout" sentinels at both sites; other out-of-range values throw `ArgumentOutOfRangeException` from the subscription/request constructor with a clear message, instead of crashing later inside the timer. Validation lives in a single internal `TimeoutValidation` helper.

Fixes #1125

### DIFF
--- a/src/NATS.Client.Core/NatsSubBase.cs
+++ b/src/NATS.Client.Core/NatsSubBase.cs
@@ -30,10 +30,6 @@ public enum NatsSubEndReason
 /// </summary>
 public abstract class NatsSubBase
 {
-    // Timer.Change and Task.WaitAsync throw for dueTime > (uint.MaxValue - 1) ms (~49.7 days);
-    // values beyond this are treated as "no timeout".
-    internal static readonly TimeSpan MaxSupportedTimeout = TimeSpan.FromMilliseconds(uint.MaxValue - 1);
-
     private static readonly byte[] NoRespondersHeaderSequence = { (byte)' ', (byte)'5', (byte)'0', (byte)'3' };
     private readonly ILogger _logger;
     private readonly object _gate = new();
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/NATS.Client.Core/NatsSubBase.cs [hunk 1]
```diff
@@ -604,6 +604,9 @@ protected void EndSubscription(NatsSubEndReason reason)
 #pragma warning restore CA2012
     }
 
+    private static TimeSpan NormalizeTimeout(TimeSpan? value, string paramName)
+        => value is { } v ? TimeoutValidation.Validate(v, paramName, TimeSpan.Zero) : TimeSpan.Zero;
+
```
scottf: You could just make Validate take TimeSpan? and do it all in one place.

scottf (approved):
LGTM

src/NATS.Client.Core/NatsSubBase.cs
mtmk: Done. `Validate` now takes `TimeSpan?` and handles the null case, so the `NormalizeTimeout` wrapper is gone and both call sites validate in one place.

PR #1136, opened 2026-05-13
Add v2.12 to test matrix, drop v2.10

mtmk:
Bump nats-server test matrix in the Linux, Linux Core, Windows, and Perf workflows to run against v2.12 and drop v2.10.


piotrpio (approved):
LGTM

Issue #1137, opened 2026-05-13
JetStream CreateOrUpdateConsumerAsync requires Name even when DurableName is set


mtmk:
`NatsJSContext.CreateOrUpdateConsumerAsync` only inspects `ConsumerConfig.Name` when constructing the create-consumer subject. If the caller sets `DurableName` but leaves `Name` empty, the request goes to the legacy ephemeral endpoint `$JS.API.CONSUMER.CREATE.<stream>` and the server rejects it with `consumer expected to be ephemeral but a durable name was set in request`.

Repro:

```csharp
await js.CreateOrUpdateConsumerAsync("ORDERS", new ConsumerConfig
{
    DurableName = "order-processor",
    AckPolicy = ConsumerConfigAckPolicy.Explicit,
});
// NatsJSApiException : consumer expected to be ephemeral but a durable name was set in request
```

Workaround is to set both `Name` and `DurableName` to the same value.

The Go client (`jetstream/consumer.go`) falls back to `Durable` and then to a generated name before building the subject, so `Durable`-only works there. Suggested fix is the same fallback in `NatsJSContext.Consumers.cs` around the subject construction:

```csharp
var name = config.Name;
if (string.IsNullOrWhiteSpace(name))
    name = config.DurableName;
if (!string.IsNullOrWhiteSpace(name))
    subject += $".{name}";
```

PR #1138, opened 2026-05-13
Add JetStream basic example

mtmk:
Adds the JetStream basic snippet to NatsIODocs, mirroring the Go, Java, and JS counterparts.

Sets both `Name` and `DurableName` on the consumer to work around #1137.


scottf (approved):
LGTM

PR #1143, opened 2026-05-14
Experiment: revert #745 to test Windows CI flap

### LINKED ISSUE #1142: Windows CI flap on TlsPreferTest after merging #745 into release/3.0
Windows non-net8.0 CI started flapping on the 3.0.0-preview.6 release PR (#1141) after #745 was merged into main and pulled into release/3.0. The same Windows runners were green on the prior preview releases (3.0.0-preview.4 #1123, 3.0.0-preview.5 #1130) and on the 2.8.0 release PR (#1135) the same day, none of which contained #745.

Failing tests (all `NATS.Client.Core.Tests`, all on net10.0 / net481, net8.0 always passes):

- `TlsPreferTest.Prefer_mode_attempts_tls_upgrade_when_server_advertises_tls_available` (8 of 10 failed jobs across 3 reruns)
- `Utf8SubjectMockServerTest.Emoji_subject_and_reply_to_are_decoded_correctly`
- `PingCancellationTest.PingAsync_throws_when_cancelled_waiting_for_pong`

TlsPreferTest signature is an assertion (not a timeout):

```
FluentAssertions.Execution.AssertionFailedException : Expected tlsRelated to be True
because Prefer mode should attempt TLS upgrade when server advertises tls_available=true,
but found False.
```

The test reads the client's first byte after sending `INFO` with `tls_available=true`. A 0x16 byte means the client attempted a TLS handshake; a `'C'` means it sent plaintext CONNECT. The assertion fires when the server saw neither AND the connect exception chain contained no `AuthenticationException` / `SocketException` / `IOException`.

The other two are the connect-timeout family already tracked as test-harness flap (`NatsUtils.ConnectRetryAsync` timing out under contention), but they have surfaced more frequently on this branch.

The only Core source change between preview.5 (green) and preview.6 (red x3) is #745. Static read of that diff suggests it should be a no-op for the TLS path (adds a new event, wraps existing `_eventChannel.Writer.TryWrite` calls in a `PushEvent` helper, adds a server-error push site that only fires when the server emits `-ERR`). The TLS test server in `TlsPreferTest` never emits `-ERR`.

Hypotheses:

1. The `PushEvent` refactor shifts startup timing on Windows non-net8.0 enough to lose the race that the test relies on.
2. Construction cost of `NatsServerErrorEventArgs` or extra branches in the event-dispatch loop add startup overhead.
3. Genuinely independent flap and the streak is coincidence (less likely given 2-for-2 green before, 0-for-3 after).

Suggested next steps:

- Cherry-pick a revert of #745 onto a throwaway branch off release/3.0, push, observe Windows CI to confirm or refute the correlation.
- If confirmed, narrow which sub-change in #745 is responsible (the PushEvent refactor itself vs the new event-dispatch branch vs the read-protocol push site).
- Likely separate work: widen `NatsUtils.ConnectRetryAsync` timeout at `tests/NATS.Client.TestUtilities/NatsUtils.cs` to mitigate the connect-timeout family.

Discovered while cutting 3.0.0-preview.6 (#1141). The preview will likely be admin-merged with this issue tracking the investigation; the failures are not new logic regressions in the release.

mtmk:
Throwaway branch to test whether reverting #745 clears the Windows non-net8.0 CI flap seen on the 3.0.0-preview.6 release PR (#1141). Not for merge.

Fixes #1142

- [ ] Confirm Windows net10.0 / net481 CI is green across 2-3 reruns
- [ ] If green, narrow which sub-change in #745 is responsible
- [ ] Close this PR without merging


mtmk:
covered by #1147

PR #1144, opened 2026-05-14
Experiment: revert #745 PushEvent callsite refactor only

mtmk:
Follow-up to #1143. Keeps everything in #745 except the inline-callsite-to-`PushEvent` refactor in `NatsConnection.cs`. New types, event surface, `-ERR` push site, and the `PushEvent` helper all stay. The 7 existing callsites are reverted to inline `_eventChannel.Writer.TryWrite(...)`.

If Windows non-net8.0 goes green here: the refactor itself (or its inlining behavior on net10.0/net481) is what shifts `TlsPreferTest` timing.
If it still flaps: the culprit is in the new event-pump dispatch case, the new event member, or some interaction.

Related to #1142. Not for merge.

- [ ] Run Windows CI 2-3 times
- [ ] Decide next step based on outcome

### DIFF
--- a/src/NATS.Client.Core/NatsConnection.cs
+++ b/src/NATS.Client.Core/NatsConnection.cs
@@ -159,7 +159,7 @@ public partial class NatsConnection : INatsConnection
         {
             if (value?.LameDuckMode == true)
             {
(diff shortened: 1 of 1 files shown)


mtmk:
covered by #1147

PR #1148, opened 2026-05-18
Quiet recoverable server-error logs

### LINKED ISSUE #847: Exceptions thrown during auth expiry despite successful reconnect
### Observed behavior

# summary

When authentication expires during an established NATS connection, the client library successfully reconnects but generates uncaught exceptions. Specifically, in NatsReadProtocolProcessor.DispatchCommandAsync(), the code sets an exception on the _waitForPongOrErrorSignal TaskCompletionSource:

_waitForPongOrErrorSignal.TrySetException(new NatsServerException(error));

This TaskCompletionSource is designed for initial connection handling but continues to be used for error signaling during the connection's lifetime. When authentication expires and the server sends an error message, this exception isn't observed by any waiting task, resulting in unobserved task exceptions that appear in logs or crash reports.

This is specifically acute in my case because we use auth callouts and set our nats user jwts to expire at 15 mins. 

# impact

- Users experience uncaught exceptions in their application logs even though the library successfully reconnects
- These exceptions create noise and may trigger error monitoring systems unnecessarily
- It's confusing for users who need to differentiate between harmless authentication expiry exceptions and actual issues
- May lead users to implement unnecessary error handling or workarounds

### Expected behavior

I can see a case to log that the user auth is expired but when the reconnect attempts are successful, I wouldn't expect to throw an exception. 

### Server and client version

Client: v2.5.10
Server: 2.10

### Host environment

_No response_

### Steps to reproduce

_No response_

mtmk:
Server-initiated graceful disconnects (User/Account Authentication Expired, Authentication Revoked, Stale Connection) are expected and recovered by the client's normal reconnect flow. Logging them at Error every 15 minutes (common with auth-callout JWTs) creates noise in monitoring without indicating an actual problem. They now log at Debug; other -ERR kinds (Permissions Violation, Maximum Connections, etc.) continue to log at Error. The `ServerError` event still fires for every kind, so callers wanting visibility can subscribe. Matches the nats.go behaviour of not surfacing these as errors and routing them through a callback instead.

The unobserved-exception part of the same issue was already fixed by #1051; the new test pins both behaviours.

The new test also exposed a pre-existing bug in -ERR parsing: when PONG and -ERR arrived in the same TCP read, `buffer.GetOffset(position.Value)` on the post-PONG slice returned the segment-internal index instead of the offset relative to the slice start, so `Slice(0, GetOffset - 1)` threw `ArgumentOutOfRangeException` and the server error was never logged or signalled. Switched to `SequencePosition`-based slicing.

Fixes #847

### DIFF
--- a/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
+++ b/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
@@ -424,13 +424,15 @@ internal sealed class NatsReadProtocolProcessor : IAsyncDisposable
                 _socketReader.AdvanceTo(buffer.Start);
                 var newBuffer = await _socketReader.ReadUntilReceiveNewLineAsync().ConfigureAwait(false);
                 var newPosition = newBuffer.PositionOf((byte)'\n');
-                var error = ParseError(newBuffer.Slice(0, newBuffer.GetOffset(newPosition!.Value) - 1));
+                var lineWithCR = newBuffer.Slice(0, newPosition!.Value);
+                var error = ParseError(lineWithCR.Slice(0, lineWithCR.Length - 1));
                 HandleServerError(error);
                 return newBuffer.Slice(newBuffer.GetPosition(1, newPosition!.Value));
             }
             else
             {
-                var error = ParseError(buffer.Slice(0, buffer.GetOffset(position.Value) - 1));
+                var lineWithCR = buffer.Slice(0, position.Value);
+                var error = ParseError(lineWithCR.Slice(0, lineWithCR.Length - 1));
                 HandleServerError(error);
                 return buffer.Slice(buffer.GetPosition(1, position.Value));
             }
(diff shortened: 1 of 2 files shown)


mtmk:
@claude can you review 4cfe182?

src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
```diff
@@ -116,6 +116,28 @@ private static string ParseError(in ReadOnlySequence<byte> errorSlice)
...
+        // for these and surfaces them via AsyncErrorCB.
+        var level = args.Kind switch
+        {
+            NatsServerErrorKind.AuthenticationExpired => LogLevel.Debug,
+            NatsServerErrorKind.AccountAuthenticationExpired => LogLevel.Debug,
+            NatsServerErrorKind.AuthenticationRevoked => LogLevel.Debug,
+            NatsServerErrorKind.StaleConnection => LogLevel.Debug,
+            _ => LogLevel.Error,
+        };
+        _logger.Log(level, NatsLogEvents.Protocol, "Server error {Error}", error);
```
scottf, line 136: Since it's not an error, maybe change the message? Also, doesn't the user already have a way to monitor connections state and do their own logging?

scottf (approved):
LGTM

PR #1149, opened 2026-05-18
Cap KV duplicate_window at max_age

### LINKED ISSUE #823: KV max_age vs duplicate_window
### Observed behavior

https://github.com/nats-io/nats-architecture-and-design/pull/336

### Expected behavior

.

### Server and client version

.

### Host environment

_No response_

### Steps to reproduce

_No response_

mtmk:
Per ADR-8 rev 9, when a KV bucket has a Key TTL the client must cap `duplicate_window` so it never exceeds the TTL: leave the server default (2 min) when `MaxAge` is unset or greater than 2 min, and set it equal to `MaxAge` when `MaxAge` is between 0 and 2 min. Applies on create, update, and create-or-update since all three paths build the stream config in one place. Matches the fix already shipped in nats.go and nats.java.

Fixes #823


scottf (approved):
LGTM

PR #1150, opened 2026-05-18
Fix durable consumer create when only DurableName is set

### LINKED ISSUE #1137: JetStream CreateOrUpdateConsumerAsync requires Name even when DurableName is set

mtmk:
`NatsJSContext.CreateOrUpdateConsumerAsync` only inspected `ConsumerConfig.Name` when building the create subject. With `DurableName` set but `Name` empty, the request went to `$JS.API.CONSUMER.CREATE.<stream>` and the server rejected it with `consumer expected to be ephemeral but a durable name was set in request`. Fall back to `DurableName` to match the Go (`upsertConsumer`) and Java (`_createConsumer`) clients.

Previously surfaced in discussion #234 and resolved with a "set both Name and DurableName" workaround rather than a code change.

Fixes #1137

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
+++ b/src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
@@ -263,6 +263,7 @@ public partial class NatsJSContext : INatsJSContext
 
         if (!string.IsNullOrWhiteSpace(name))
         {
+            ThrowIfInvalidConsumerName(name!);
             subject += $".{name}";
         }
 
--- a/src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
+++ b/src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
@@ -255,9 +255,15 @@ public partial class NatsJSContext : INatsJSContext
     {
         var subject = $"{Opts.Prefix}.CONSUMER.CREATE.{stream}";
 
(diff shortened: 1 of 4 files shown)


scottf (approved):
LGTM

PR #1151, opened 2026-05-19
Fix repeated idle heartbeat timeout notifications

### LINKED ISSUE #952: Excessive NatsJSTimeoutNotification from NatsJSConsume due to periodic heartbeat timer
### Context
In NatsJSConsume, the idle heartbeat timer is currently configured as periodic:
`_timer.Change(_hbTimeout, _hbTimeout);`
This was introduced in [PR #185](https://github.com/nats-io/nats.net/pull/185).
Other consumers (e.g. NatsJSOrderedConsume) still use the one-shot timer style:
`_timer.Change(_hbTimeout, Timeout.Infinite);`

### Problem
Because the timer is periodic, as soon as the first real idle timeout occurs, the client starts raising NatsJSTimeoutNotification repeatedly every _hbTimeout until messages arrive again.

### Possible approaches:
Hybrid approach:
Keep the timer periodic for safety, but only send NatsJSTimeoutNotification on the first tick after silence. Subsequent ticks would still re-trigger internal pull logic, but not bubble notifications to the user.

Configurable mode:
Allow users to opt into Timeout.Infinite (one-shot) or periodic behavior via NatsSubOpts. This way, high-reliability scenarios can use the safer periodic mode, while others can avoid notification spam.

### Real case
In our environment we have rare but valid messages with payload (not just heartbeats).

Messages may arrive only once in a while, which is still correct behavior for the application.

However, because of the periodic timer, the consumer produces infinite timeout notifications between these payload messages.

This makes it look like the consumer is failing, while in reality it is functioning as expected.

### Contribution

_No response_

mtmk:
`NatsJSConsume` keeps its idle heartbeat timer periodic so pulls keep getting re-issued during long silences, but it was also firing `NatsJSTimeoutNotification` on every tick. The notification carries no payload, so the repeats added no information past the first one. Track an edge-triggered flag so the notification is emitted once when silence starts and cleared on the next received message (data or heartbeat). Timer cadence is unchanged, so behaviour around pulls is unaffected.

Fixes #952

### DIFF
--- a/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSConsume.cs
@@ -46,6 +46,7 @@ internal class NatsJSConsume<TMsg> : NatsSubBase
     private long _pendingMsgs;
     private long _pendingBytes;
     private int _consecutive503Errors;
+    private int _timeoutNotified;
 
     public NatsJSConsume(
         long maxMsgs,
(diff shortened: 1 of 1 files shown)


Ivandemidov00:
Thanks @mtmk for pushing the fix. I don't have access to the original logs/packet capture anymore, but based on the issue description, it seems like (heartbeats were missing / messages including heartbeats were arriving but notifications kept firing)

PR #1159, opened 2026-05-26
Clarify DI package descriptions

mtmk:
The two DI packages serve different setups (minimal/AOT vs ad-hoc JSON serialization) but the NuGet descriptions did not surface the distinction. Update each to name the trade-off and cross-link the other package.


scottf (approved):
LGTM

PR #1163, opened 2026-05-27
Clarify DI package descriptions

mtmk:
Update the NuGet <Description> for the two DI packages to state what each one is for: NATS.Client.Hosting (minimal deps, AOT) vs NATS.Extensions.Microsoft.DependencyInjection (ad-hoc JSON by default), each pointing at the other.


scottf (approved):
LGTM

PR #1175, opened 2026-06-01
Fix server.port type and trace tag source

### LINKED ISSUE #1156: Trace activity tags use ServerInfo.Host instead of connect URI
StartSendActivity and StartReceiveActivity in src/NATS.Client.Core/Internal/Telemetry.cs
populate server.address and network.peer.address from connection.ServerInfo.Host,
which is the server's bind address (often 0.0.0.0), not the address the client
dialled.

Per OTel semconv, server.address should be from the client's perspective.
network.peer.address should be the actual remote endpoint; for NATS without
a proxy this is the same value.

Companion fix to the metric tags, which were corrected to use the connect
URI host/port.

- Replace conn.ServerInfo.Host with the current connect URI host in both
  StartSendActivity and StartReceiveActivity.
- Same for ServerPort.
- Decide whether network.peer.address should be the resolved IP (different
  from server.address when DNS is involved) or just the dialled host.

### LINKED ISSUE #1167: `server.port` activity tag stored as string
OTel semantic conventions define `server.port` as integer, but the trace
activity tags currently store it as a string via `.ToString()`. The new
OTel metric tags store it as an integer (spec-correct). Trace tags should
be updated to match.

Affects send and receive activities. Found during PR review for OTel
metrics work.

mtmk:
Emit server.port and network.peer.port as integers per the OTel messaging semantic conventions instead of strings, and source server.address/server.port from the connect URI rather than the server bind address (which is often 0.0.0.0). The host, boxed port and client id are cached per ServerInfo change so the per-message trace path avoids re-boxing and re-allocating.

Fixes #1167
Fixes #1156

### DIFF
--- a/src/NATS.Client.Core/Internal/Telemetry.cs
+++ b/src/NATS.Client.Core/Internal/Telemetry.cs
@@ -139,18 +139,19 @@ internal static class Telemetry
             if (replyTo is not null)
                 len++;
 
-            var serverPort = conn.ServerInfo.Port.ToString();
+            var serverPort = conn.BoxedServerPort; // boxed once per ServerInfo change, reused for server.port and network.peer.port
+            var serverHost = conn.ServerHost; // grabbed once per ServerInfo change
             tags = new KeyValuePair<string, object?>[len];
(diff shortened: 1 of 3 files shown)


scottf (approved):
LGTM

PR #1179, opened 2026-06-03
Fix flaky CI tests

### LINKED ISSUE #1165: Flaky test: SlowConsumerTest ping RTT exceeds 1000ms threshold on CI
`SlowConsumerTest.JetStream_consume_slow_consumer_should_not_block_connection` asserts `maxPingRttMs < 1000` to confirm the socket reader is not blocked while a JetStream consumer is slow. The test sends 20 pings; nearly all return in sub-millisecond time, but one ping intermittently spikes to just over 1000ms on CI runners, tripping the assertion.

Example from a recent run: pings 1-14 and 16-20 were ~0.2ms, ping 15 was 1027.8ms, max RTT 1027.8ms -> fail. No messages dropped, pub/sub flowed 10/10, so the connection was not actually blocked. The spike looks like runner scheduling/GC jitter rather than a client regression.

Originally seen on Windows runners; also now observed on Linux, so it is not platform-specific.

Fix options:
- Raise the threshold (e.g. to 2000ms), which weakens the "not blocked" signal.
- Drop the first/outlier ping from the max, or use a percentile instead of the raw max.
- Warm-ping before measuring so JIT/connection warmup is excluded.

Test: `tests/NATS.Client.JetStream.Tests/SlowConsumerTest.cs` (the `maxPingRttMs < 1000` assertion).

mtmk:
Environmental CI flaps, all test-only fix by increasing timeout tolerances.

Fixes #1165

### DIFF
--- a/tests/NATS.Client.Core.Tests/ConnectionRetryTest.cs
+++ b/tests/NATS.Client.Core.Tests/ConnectionRetryTest.cs
@@ -150,11 +150,16 @@ public class ConnectionRetryTest
 
         _output.WriteLine($"reconnects: {reconnects}, sent: {sent}, received: {received}");
 
-        // some messages may still be lost, as socket could have been disconnected
-        // after socket.WriteAsync returned, but before OS sent
-        // check to ensure that the loss was < 1%
+        // Some messages may still be lost, as the socket could have been disconnected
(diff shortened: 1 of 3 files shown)


Jarema (approved):
LGTM

PR #1180, opened 2026-06-04
Deprecate SkipSubjectValidation

### LINKED ISSUE #1160: Remove SkipSubjectValidation option, always validate subjects
NATS.Net `NatsOpts.SkipSubjectValidation` (default `false`) lets callers bypass subject validation. Validation rejects empty subjects and subjects containing whitespace (space, tab, CR, LF), which is what prevents CRLF protocol injection where a `\r\n` in a subject could inject arbitrary NATS commands.

Making a security check opt-out is the wrong default to even offer. The validation cost is negligible relative to the network round trip, and a caller who "fully controls all subjects" gains nothing measurable by turning it off. There is no good reason for this to be configurable.

Proposed for 3.0:
- Remove the `SkipSubjectValidation` property from `NatsOpts`.
- Always run `ValidateSubject` / `ValidateReplyTo` on the publish, subscribe, and request paths.
- Remove the option from the benchmark and update the tests that set it.

Breaking change: code that sets `SkipSubjectValidation` will no longer compile. Fits the 3.0 milestone.

mtmk:
SkipSubjectValidation is now marked obsolete but still honored: setting it true bypasses subject, reply-to, and queue-group validation as before. Disabling it is discouraged. The likely failure is mundane: a subject containing a space or tab is split into tokens on the wire, so "PUB foo bar 5" is read by the server as subject "foo" with reply-to "bar", and "SUB foo bar 1" as subject "foo" joined to queue group "bar". Either way the call is silently misrouted with no error. Validation also blocks CRLF protocol injection.

On the performance concern that motivated the original opt-out: a publish microbenchmark (100k PublishAsync per op against a local nats-server):

| Runtime | No validation | With validation | Ratio |
| --- | --- | --- | --- |
| .NET 10.0 | 38.99 ms | 38.68 ms | 0.99 |
| .NET 8.0 | 44.12 ms | 45.42 ms | 1.03 |
| .NET Framework 4.8.1 | 123.21 ms | 129.91 ms | 1.05 |

Fixes #1160


Jarema (approved):
Lgtm

PR #1181, opened 2026-06-08
Unify subscription channel overflow defaults

### LINKED ISSUE #1161: Unify sub channel overflow defaults, raise capacity to 16K
`NatsConnection` defaults `SubPendingChannelFullMode` to `DropNewest`, but `NatsClient` (the simplified API) overrides it to `Wait`. Same library, two different overflow behaviors depending on entry point.

`Wait` is also the more dangerous of the two: when a subscription's channel fills, backpressure reaches the socket read loop, the client stops draining the connection, and the server disconnects it as a slow consumer. The simplified client ships this as its default.

Proposed for 3.0:
- Raise the default `SubPendingChannelCapacity` from 1024 to 16384. A deeper buffer makes overflow rare for typical workloads, which is the protection `NatsClient` was approximating with `Wait`.
- Remove the `Wait` override in `NatsClient` so it inherits `DropNewest`. After this, the only `NatsClient`-specific default is `NatsClientDefaultSerializerRegistry` (JSON).

Behavior change: existing `NatsClient` users move from block/slow-consumer-disconnect under sustained overload to dropping newest messages. This is the safer direction but is breaking, hence 3.0. The 16K buffer is a per-subscription ceiling (not preallocated); large-payload firehoses can buffer more bytes before dropping. Drops still surface via `MessageDropped`.

mtmk:
All entry points now share one subscriber-channel overflow policy. NatsClient and the DI builder previously forced Wait while NatsConnection used DropNewest; Wait can stall the socket read loop and get the client disconnected as a slow consumer. They now inherit the NatsOpts default DropNewest, and the default channel capacity goes from 1024 to 16384 so overflow stays rare. Breaking for NatsClient and DI users under sustained overload: newest messages are dropped (surfaced via MessageDropped) instead of blocking.

Fixes #1161

### DIFF
--- a/src/NATS.Client.Core/NatsOpts.cs
+++ b/src/NATS.Client.Core/NatsOpts.cs
@@ -228,9 +228,9 @@ public sealed record NatsOpts
 
     /// <summary>
     /// This value will be used for subscriptions internal bounded message channel capacity.
-    /// The default subscriber pending message limit is 1024.
+    /// The default subscriber pending message limit is 16384.
     /// </summary>
-    public int SubPendingChannelCapacity { get; init; } = 1024;
+    public int SubPendingChannelCapacity { get; init; } = 16384;
 
     /// <summary>
     /// This value will be used for subscriptions internal bounded message channel <c>FullMode</c>.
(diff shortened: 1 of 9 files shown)


scottf (approved):
LGTM

PR #1188, opened 2026-06-16
Fix ordered push consumer subscription leak on teardown

### LINKED ISSUE #1187: memory leak of NatsJSOrderedPushConsumerSub
### Observed behavior

NatsJSOrderedPushConsumer.DisposeAsync() does not dispose its current _sub, so the Sub's OnConnectionOpened event handler never gets removed from the parent NatsConnection. Each call to NatsObjStore.GetAsync() or WatchAsync() leaves one Sub permanently rooted on the singleton connection's ConnectionOpened invocation list.

### Expected behavior

the sub should disposed.

### Server and client version

2.6.11
still present on the last version.

### Host environment

_No response_

### Steps to reproduce

when call the NatsObjStore.GetAsync() or WatchAsync().

mtmk:
NatsJSOrderedPushConsumer.DisposeAsync did not dispose its current subscription, so the sub's ConnectionOpened handler stayed attached to the connection and the subscription was rooted for the connection's lifetime. Each object store Get/Watch call leaked one subscription on long-lived connections. Dispose the current sub during teardown, after the background loops stop.

Diagnosis and original fix by @haoguanjun in #1186; this rebases the same change onto main and adds a regression test.

Fixes #1187

### DIFF
--- a/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
+++ b/src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
@@ -178,6 +178,17 @@ internal class NatsJSOrderedPushConsumer<T>
 
         _msgChannel.Writer.TryComplete();
 
+        // Dispose the current subscription. Both background loops have completed
+        // by now, so no new subscription can be created. Without this the sub's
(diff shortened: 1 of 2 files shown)


MauriceVanVeen (approved):
LGTM

PR #1190, opened 2026-06-16
Release 3.0.0-preview.10

mtmk:
Preview release with JetStream teardown fixes.

- jetstream: simplify ordered push consumer sub teardown (#1191)
- jetstream: dispose ordered push consumer sub on teardown (#1188)

### DIFF
--- a/version.txt
+++ b/version.txt
@@ -1,1 +1,1 @@
-3.0.0-preview.9
+3.0.0-preview.10


mtmk:
Both minor findings from the earlier review (missing `ConfigureAwait(false)` and the `UnsubscribeAsync`/`DisposeAsync` exception-safety note) are addressed by #1191, now folded into this branch. `DisposeAsync` calls `UnsubscribeAsync` internally, so the explicit call was dropped and the single `DisposeAsync` now uses `ConfigureAwait(false)`.

@claude can you have a quick look?

Jarema (approved):
LGTM

PR #1191, opened 2026-06-16
Simplify ordered push consumer sub teardown

mtmk:
Follow-up to the ordered push consumer dispose fix. NatsSubBase.DisposeAsync already calls UnsubscribeAsync internally, so the explicit UnsubscribeAsync before DisposeAsync was redundant. Collapse to a single DisposeAsync and add ConfigureAwait(false) to match the surrounding awaits.


MauriceVanVeen (approved):
LGTM

PR #1192, opened 2026-06-17
Move socket connection interfaces to NATS.Client.Abstractions

### LINKED ISSUE #1185: Move socket connection interfaces to NATS.Client.Abstractions
Move the socket transport contracts from NATS.Client.Core to
NATS.Client.Abstractions so custom transport implementers can depend on the
contract without referencing Core, same rationale as the serialization
interfaces already in Abstractions.

Scope:
- Move `INatsSocketConnection` and `INatsTlsUpgradeableSocketConnection`. Both
  are BCL-only (`Memory<byte>`, `Socket`), no Core type dependencies.
- Keep `INatsSocketConnectionFactory` in Core: its `ConnectAsync` takes
  `NatsOpts`, so it cannot move without dragging Core options along or changing
  the signature (breaking). Revisit only if a narrower options abstraction is
  introduced.
- `ISocketConnection` is already obsolete; leave it in Core.

Compatibility:
- Keep the `NATS.Client.Core` namespace on the moved types to avoid source
  breaks for existing implementers.
- Add `[assembly: TypeForwardedTo(...)]` in Core for each moved type to preserve
  binary compatibility.

Parent: #851

mtmk:
Move the socket transport contracts INatsSocketConnection and INatsTlsUpgradeableSocketConnection to NATS.Client.Abstractions so custom transport implementers can depend on the contract without referencing Core, matching the serialization interfaces already there. The types keep the NATS.Client.Core namespace and are forwarded via TypeForwardedTo to preserve source and binary compatibility; the obsolete ISocketConnection and INatsSocketConnectionFactory stay in Core. The ABI compatibility harness is extended to cover both moved interfaces through a transient dependency compiled against an older release.

Fixes #1185

### DIFF
--- a/src/NATS.Client.Abstractions/INatsSocketConnection.cs
+++ b/src/NATS.Client.Abstractions/INatsSocketConnection.cs
@@ -1,0 +1,43 @@
+using System.Net.Sockets;
+
+namespace NATS.Client.Core;
+
+/// <summary>
+/// Represents a socket connection to a NATS server.
+/// </summary>
+/// <remarks>
+/// This interface defines the contract for low-level socket connections to NATS servers,
+/// providing methods for sending and receiving data and disposing the socket.
+/// Disposing should attempt to perform graceful shutdown of the socket.
+/// </remarks>
+public interface INatsSocketConnection : IAsyncDisposable
+{
+    /// <summary>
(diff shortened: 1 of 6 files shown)


scottf (approved):
LGTM

Issue #1162, opened 2026-05-27
Add OTel counter for dropped messages and slow consumers


mtmk:
Dropped messages currently surface only via the `MessageDropped` and `SlowConsumerDetected` events. With `DropNewest` as the overflow default, a counter is the observability path most users will actually watch. Not covered by the metric list in #1153, so tracking separately.

Emit on the existing `OnMessageDropped` / slow-consumer paths under the `Meter("NATS.Net")` from #1153, reusing the same tag set so it correlates with the other signals:
- a dropped-messages counter (e.g. `nats.client.messages.dropped`)
- expose pending channel depth at drop time (already passed to `OnMessageDropped`) as a tag or gauge

No messaging semantic convention covers drops, so this is a deliberately NATS-specific metric under the `NATS.Net` meter. Depends on the meter from #1153 landing first.

mtmk:
Shipped in v3.0.0 via #1194. The PR merged via release/3.0 so this did not auto-close.

Issue #1157, opened 2026-05-26
Reduce trace tag cardinality from high-cardinality subjects


mtmk:
Span names are already low-cardinality via SpanDestinationName (returns
"inbox" for inbox subjects, trims to first two tokens otherwise). But
activity tag values still carry the full subject:

- messaging.destination.name
- messaging.destination_publish.name
- messaging.nats.message.subject
- messaging.nats.message.reply_to

For inbox subjects every request emits a unique value (_INBOX.<nuid>).
The same applies to any subject pattern with embedded ids:
orders.new.<id>, users.<id>.events, etc. Backends that index these
tags (Tempo, Jaeger) will see unbounded cardinality.

Companion to (closed) #861, which observed the same symptom for span
names; the underlying tag-value cardinality is still present.

Options to consider:
- Replace inbox subject with "inbox" (matching SpanDestinationName) on
  the destination/subject tags, keeping the raw value only available
  via the user enrich hook.
- For receive activities, drop messaging.destination.name when
  messaging.destination.template is already set (template carries the
  low-cardinality info).
- Same treatment for replyTo when it is an inbox.
- For arbitrary user subjects with embedded ids, no auto-detection is
  possible. Expose a hook via NatsInstrumentationOptions to let users
  map subject -> template, or rely on the existing Enrich callback to
  rewrite tags.

Target: 3.x.

mtmk:
Shipped in v3.0.0 via #1194. The PR merged via release/3.0 so this did not auto-close.

PR #1194, opened 2026-06-23
Add OTel ack/dropped metrics and collapse inbox trace tags

### LINKED ISSUE #1158: Add operation.duration histogram for JetStream ack
The messaging.client.operation.duration histogram currently covers
publish, request, and subscribe. JetStream ack is the obvious next
candidate from the original metrics scope (#1153) and was deferred
from the initial implementation.

Wire AckAsync (and the other Ack* methods) to record duration with
messaging.operation=ack, including error.type on failure following
the same pattern as publish/request/subscribe.

mtmk:
Three OpenTelemetry follow-ups on the 3.x metrics work: record the operation.duration histogram for JetStream acks, add a dropped-messages counter on the message-drop path, and collapse inbox subjects to a constant on the high-cardinality trace tags so backends don't index a unique value per request. The raw subject stays reachable through the Enrich callback.

Fixes #1158, #1162, #1157.

### DIFF
--- a/src/NATS.Client.Core/Internal/Telemetry.cs
+++ b/src/NATS.Client.Core/Internal/Telemetry.cs
@@ -151,7 +151,7 @@ internal static class Telemetry
             tags = new KeyValuePair<string, object?>[len];
             tags[0] = new KeyValuePair<string, object?>(Constants.SystemKey, Constants.SystemVal);
             tags[1] = new KeyValuePair<string, object?>(Constants.OpKey, Constants.OpPub);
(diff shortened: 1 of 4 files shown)


mtmk:
@claude please re-review the latest commit. Addresses the two earlier review notes by documenting the intentional choices: inbox-prefix breadth in LowCardinalitySubject, and collapsing all ack-protocol kinds to one OpAck histogram.

scottf (approved):
LGTM

PR #1195, opened 2026-06-23
Exclude NATS status frames from consumed metrics

### LINKED ISSUE #1168: OTel semconv: should NoResponders 503 replies count toward consumed.messages / received.bytes?
Open question raised during the OTel metrics PR review.

Current behaviour (after PR #1153 plus follow-up review fix):

- `NatsSubBase.ReceiveAsync` excludes 503 NoResponders and StopOnEmptyMsg
  sentinels from `messaging.client.consumed.messages` and
  `nats.client.received.bytes`.
- `ReplyTask.SetResult` (Direct request/reply path) now mirrors that and
  also excludes 503 replies.

Open question: is that what the OTel messaging semconv actually wants?

- `messaging.client.consumed.messages` is defined as "Number of messages
  that were delivered to the application." A 503 sentinel never reaches
  user code, so excluding it seems right.
- `nats.client.received.bytes` is our own counter (no spec to match), but
  the bytes were physically on the wire. Could be argued either way.

Spec reference:
https://opentelemetry.io/docs/specs/semconv/messaging/messaging-metrics/

Decide and document the convention so the two code paths stay consistent
and so JetStream control messages (heartbeats, flow-control) are handled
the same way.

mtmk:
consumed.messages and received.bytes now count only messages delivered to the application, per the OpenTelemetry messaging convention. NATS status and control frames (no-responder 503s, JetStream heartbeats, flow-control, and protocol notifications) are consumed internally and were previously inflating both counters on the receive path.

Fixes #1168.

### DIFF
--- a/src/NATS.Client.Core/NatsSubBase.cs
+++ b/src/NATS.Client.Core/NatsSubBase.cs
@@ -374,7 +374,12 @@ public abstract class NatsSubBase
             // the hand-off to the subscription channel. received.bytes follows the same
             // convention. Messages dropped by cancellation, channel closure, or serializer
             // exceptions are not counted.
-            if (Telemetry.ConsumedMessages.Enabled || Telemetry.ReceivedBytes.Enabled)
+            //
+            // NATS status/control frames (no-responder 503s and JetStream heartbeats,
+            // flow-control, and protocol notifications) are consumed internally and never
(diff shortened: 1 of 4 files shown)


scottf (approved):
LGTM

Issue #948, opened 2025-09-05
NATS .NET v3.0 proposal: Support .NET 10 and drop .NET 6


mtmk:
### Proposed change

A quick heads up for our plans for a new NATS .NET release coming up in a couple of months time. We are planning to drop support for .NET 6 in favor of supporting .NET 10 right after its release. The idea is to reduce the maintenance burden as much as we can while taking advantage of any new features that might be useful for performance or developer experience. With this we will be bumping the major version to v3 to signal this change in target support also aligning with .NET release cycle to make the transition hopefully easier.

Please let us know what you think.

(btw we will carry on supporting NET Standard 2.0 and 2.1)

darkwatchuk:
Sounds great. Good plan.

jacobslusser:
Is there an updated ETA for this?

mtmk:
hey @jacobslusser 

> Is there an updated ETA for this?

We have previews out already https://www.nuget.org/packages/NATS.Net/3.0.0-preview.1 stable release depends on the feedback we get. there is no set time but we're aiming at a few weeks, tops couple of months. There should be another release coming up today or tomorrow. Anytime we have a 2.x release, we'd make a 3.0-preview release covering the same changes on top of 3.0 changes.

mtmk:
v3.0.0 was released on 2026-07-10 (v3.0.1 followed on 2026-07-20): .NET 6 support is dropped, .NET 10 is added, and netstandard2.0/2.1 remain. Closing as completed.

PR #1196, opened 2026-06-23
Remove stale net6.0 references

mtmk:
net6.0 was already dropped from the target frameworks (now netstandard2.0, netstandard2.1, net8.0, net10.0). This removes leftover net6.0 mentions from the platform-compatibility docs and corrects csproj and script comments that still labelled netstandard-only dependency groups as net6.0.

Part of #948.


scottf (approved):
LGTM

Issue #851, opened 2025-05-10
Offer an abstractions package


thompson-tomo:
### Proposed change

Release a NATS.Client.Abstractions package which provide the public interfaces for things like services while also providing the classes for constructing & de-constructing messages. 

### Use case

I want to be able to pass the received nats message to the cloud events sdk which verifies the entire message and transforms it to a cloud event. By providing the abstractions cloud events SDK only gains a small dependency and hopefully no transitive dependencies.

### Contribution

Yes

mtmk:
thanks @thompson-tomo good suggestion. sorry for the delay I thought i replied but obv did that in my head :)

I think it's a good idea. we just have to be careful not to break anything and if interfaces moving across assemblies affecting some deployments we need to be careful about that too.

thompson-tomo:
No worries at all @mtmk and I agree that it needs to be carefully managed. It might be worthwhile to make the changes in stages with an initial focus onthe serializer, then services and lastly the client,

mtmk:
that sounds great. One other thing to keep in mind is #856 where we're trying to formalize extension points like serializer being a subset of some kind of message interceptors. feel free to open a pr if you have a clear idea what needs to be done we can discuss there if you prefer or we can try to work on a detailed plan here.

thompson-tomo:
Sure I will certainly keep that in mind and see what ideas I have. as such I might even do it more granular that way discussion can be had alongside the pr.

Something like
- Serialization 
- Services
- Message Handling
- connection Handling
- Jet stream
- etc

thompson-tomo:
Just to clarify the breaking changes being alluded to will be if you are reliant on a library providing `nats.client.core` this transitive dependency might disappear as it is being replaced with the abstractions dependency. The solution will be to add an explicit dependency on `nats.client.core`.

mtmk:
NATS.Client.Abstractions shipped with v3.0.0, covering the serialization interfaces and the socket connection interfaces (#1185, PR #1192), with a package boundary test to keep the dependency direction in check. Closing. Further interface moves can get their own issues.

PR #1197, opened 2026-06-23
Add abstractions package boundary test

mtmk:
Adds a test project that references only NATS.Client.Abstractions (no Core) and exercises the use case behind the abstractions package: a third party such as a CloudEvents adapter can deconstruct a received message and build an outgoing one using just the envelope (NatsMsgContext / INatsHeaders) and the serializer interfaces. Runs in Linux and Windows CI.

Part of #851.


scottf (approved):
LGTM

PR #1198, opened 2026-06-24
Gate positive-path test connections with ConnectRetryAsync

mtmk:
Integration tests that construct a NatsConnection and use it without first gating on a successful connect can fail with a TimeoutException from the 2s connect/INFO-wait when CI runs TFMs in parallel under load. The failure surfaces as an unrelated test before any assertion runs. Gate positive-path connections with the existing ConnectRetryAsync helper (retries ConnectAsync for up to 60s) and convert plain ConnectAsync setup gates to it. Negative-path tests (expected connect failures/timeouts, reconnect/cancel/TLS/auth behaviour) are left as-is.

Also fixes two unrelated test flakes surfaced by CI: a request-reply race where the request could outrun the responder's subscription, and a slow-consumer test whose episode-2 wait used an absolute event-count threshold instead of a relative one.

### DIFF
--- a/tests/NATS.Client.Core.Tests/LowLevelApiTest.cs
+++ b/tests/NATS.Client.Core.Tests/LowLevelApiTest.cs
@@ -1,5 +1,6 @@
 using System.Buffers;
 using System.Text;
+using NATS.Client.TestUtilities2;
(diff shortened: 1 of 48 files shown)


scottf (approved):
LGTM Test files only

PR #1199, opened 2026-06-25
Simplify object store base64url encoder

### LINKED ISSUE #557: Optimize Objectstore Base64URL encoding bits
### Proposed change

From [#549](https://github.com/nats-io/nats.net.v2/pull/549#issuecomment-2218781588):

While converting things to have System.Text.Json handle Base64 encoding for us, other opportunities came up.

> I think we can also obviate a decent amount of code in ObjectStore
(snip)
> Although, not sure if we should keep that in same PR or separate, just for sake of sizing... open to opinions here.
> Edit: Actually I may be off on that on some level, Base64URL != Base64... so I'd say keep it separated.

CC: 
@galvesribeiro  (Thanks for the feedback and inspiration on this issue btw! <3)

### Use case

Eliminating intermediary string allocations here (i.e. on SHA checks) would be a perf and memory improvement if we do it well.

This would benefit users of ObjectStore, especially in high load environments.

### Contribution

I'd be happy to add it to my backlog? 😅 If someone else wants to give it a go, that's awesome too!

mtmk:
Replace the object store's hand-rolled base64url encoder. The old loop allocated a fresh `char[]` on every call; the new code encodes into a stack-allocated buffer (pooled for inputs over 512 chars), so only the result string is allocated. On net9.0+ it uses the built-in `Base64Url` encoder (url-safe in a single pass); on older target frameworks it keeps a lookup-table loop over the url-safe alphabet. The `'='` padding is retained so digests and metadata subject names stay wire-compatible with the other clients. The read path compares the digest as spans rather than building an interpolated string. Verified against the Go client by round-tripping objects in both directions.

Benchmark below (Intel Xeon w5-2445, BenchmarkDotNet ShortRun; baseline = the old `new char[]` table encoder, current = this PR). Allocation halves across every size and runtime. Throughput improves substantially on net9.0+ where the built-in `Base64Url` path runs (up to ~4.5x on 4 KB input); on net8.0 it is roughly unchanged, and on net481 it is slightly slower at small inputs, so on pre-net9 frameworks the change is an allocation reduction rather than a speedup. Size 32 is the SHA-256 digest, the common case.

| Runtime | Size | Mean (base -> current) | Alloc (base -> current) |
|---|---|---|---|
| net481 | 16 | 48.3 -> 68.8 ns | 152 -> 80 B |
| net481 | 32 | 72.8 -> 92.0 ns | 233 -> 120 B |
| net481 | 256 | 450 -> 458 ns | 1436 -> 722 B |
| net481 | 4096 | 6686 -> 6500 ns | 21960 -> 10984 B |
| net8.0 | 16 | 29.7 -> 29.7 ns | 144 -> 72 B |
| net8.0 | 32 | 47.1 -> 51.5 ns | 224 -> 112 B |
| net8.0 | 256 | 291 -> 273 ns | 1424 -> 712 B |
| net8.0 | 4096 | 4443 -> 3998 ns | 21904 -> 10952 B |
| net10.0 | 16 | 27.1 -> 18.7 ns | 144 -> 72 B |
| net10.0 | 32 | 43.7 -> 26.0 ns | 224 -> 112 B |
| net10.0 | 256 | 279 -> 89 ns | 1424 -> 712 B |
| net10.0 | 4096 | 4258 -> 933 ns | 21904 -> 10952 B |

Supersedes #565; original work by @to11mtm.

Fixes #557

PS opened a proposal in the dotnet/runtime repo to cover this case so we can avoid the extra translation step https://github.com/dotnet/runtime/issues/129847


mtmk:
@claude please review again, there was a typo preventing build before.

src/NATS.Client.ObjectStore/Internal/Encoder.cs
```diff
@@ -59,83 +50,49 @@ public static string Encode(string arg)
...
+        byte[]? rented = null;
+        var utf8 = base64Length <= 512
+            ? stackalloc byte[base64Length]
+            : (rented = ArrayPool<byte>.Shared.Rent(base64Length)).AsSpan(0, base64Length);
+        try
         {
-            output[k] = Base64PadCharacter;
+            _ = System.Buffers.Text.Base64.EncodeToUtf8(inArray, utf8, out _, out var written);
+            var count = MakeUrlSafe(utf8.Slice(0, written), raw);
+            return Encoding.ASCII.GetString(utf8.Slice(0, count));
```
MihaZupan: It'd probably be more efficient to use EncodeToChars instead

You can also use the `int`-returning `EncodeToUtf8`/`EncodeToChars` helper if you're going to ignore the OperationStatus

src/NATS.Client.ObjectStore/Internal/Encoder.cs
```diff
@@ -158,6 +115,31 @@ public static byte[] DecodeBytes(string str)
...
+                break;
+            case (byte)Base64Character63:
+                utf8[i] = (byte)Base64UrlCharacter63;
+                break;
+            case (byte)Base64PadCharacter:
+                if (raw)
+                    return i;
+                break;
+            }
+        }
```
MihaZupan:
```suggestion
        utf8.Replace((byte)Base64Character62, (byte)Base64UrlCharacter62);
        utf8.Replace((byte)Base64Character63, (byte)Base64UrlCharacter63);

        if (raw && utf8.EndsWith((byte)Base64PadCharacter))
        {
            return utf8.LastIndexOfAnyExcept((byte)Base64PadCharacter) + 1;
        }
```

mtmk:
@claude please take another look. The design has changed substantially since your first review (the encoder now multi-targets a separate test project to pin each Base64UrlEncoder branch via NuGet asset selection, plus the apicompat workflow change). Please review the full current diff fresh.

src/NATS.Client.ObjectStore/Internal/Encoder.cs
mtmk: makes sense. using `EncodeToChars` now. thanks for the feedback.

scottf (approved):
LGTM

PR #1205, opened 2026-07-01
Fix null-key tag in OpenTelemetry receive fallback

mtmk:
The receive-path telemetry fallback (used when the connection is not a fully connected NatsConnection) allocated a fixed-size tag array but only populated the reply-to slot conditionally, leaving a default null-keyed entry that was passed to StartActivity when there was no reply-to. Size the array to the tags actually set, matching the send path, and add a regression test over the fallback branch.


scottf (approved):
LGTM

PR #1206, opened 2026-07-01
Match OpenTelemetry subject filters without per-message allocation

mtmk:
The subject filter installed by FilterSubjects split the subject into a string array on every traced publish/subscribe when a filter is configured. Match the subject against the pre-tokenized patterns by walking it token by token with spans, dropping the per-message allocation. Adds coverage for empty-token boundaries (trailing and consecutive dots).


mtmk:
@claude the latest commit addresses your review comments, please re-check.

scottf (approved):
LGTM

PR #1207, opened 2026-07-01
Make shared OpenTelemetry instrumentation options thread-visible

mtmk:
The shared NatsInstrumentationOptions.Default is configured once at startup, but its Filter, Enrich, and SpanDestinationNameFormatter callbacks are read on the publish/subscribe path from other threads. Back the properties with volatile fields so a configuration change applied after traffic has started is visible to those readers rather than depending on incidental memory barriers. No new test: this is a visibility guarantee and the existing configure round-trip test already covers get/set behavior.


scottf (approved):
LGTM

PR #1208, opened 2026-07-07
OpenTelemetry: opt-in W3C Baggage propagation

### LINKED ISSUE #1200: OpenTelemetry: optionally propagate W3C Baggage across message boundaries (not just trace context)
### Summary

The built-in OpenTelemetry instrumentation propagates **W3C trace context** across publish/consume, but it does **not** propagate **W3C Baggage**. It would be valuable to optionally inject baggage on publish and extract/restore it on receive, so cross-cutting context survives a NATS hop the same way it does across HTTP.

### Current behavior (v3.0 preview)

Tracing auto-creates `Activity` spans for publish/subscribe and links them via `DistributedContextPropagator.Current` — but baggage is never carried. From `src/NATS.Client.Core/Internal/Telemetry.cs`:

- **Send:** the trace-context headers are written via `DistributedContextPropagator.Current.Inject(...)`; no `baggage` header is written.
- **Receive:** `StartReceiveActivity(...)` only calls `TryParseTraceContext(headers, ...)` to set the parent `ActivityContext`. It never calls `ExtractBaggage` and never restores `Activity.Baggage`. The `NatsInstrumentationOptions.Enrich` hook exposes subject/headers/connection but not the extracted baggage.

Result: a trace stays linked across services, but any baggage on the producer side is dropped at the NATS boundary.

### Use case (generic)

Distributed systems commonly carry cross-cutting request metadata as **OpenTelemetry Baggage** — e.g. `tenant.id`, `session.id`, `enduser`/user identifiers, a `correlation_id`, or feature-flag context. Teams rely on baggage so this context can be re-stamped onto downstream spans and logs and used to **filter/correlate a single tenant's or request's activity across every service**. The OTel HTTP instrumentations propagate baggage automatically; messaging is the gap.

When a workflow crosses NATS (the common pattern: an HTTP request publishes a message, a consumer service processes it), that baggage is lost, so downstream spans/logs can no longer be attributed to the originating tenant/user/request. This forces every NATS-using app to hand-roll baggage inject/extract over `NatsHeaders`, duplicating logic the trace-context path already has.

### Proposal

Add **opt-in** baggage propagation to the instrumentation:

1. On publish, also inject the `baggage` header (e.g. via `W3CBaggagePropagator` / the OTel baggage propagator, or `Baggage`-aware injection).
2. On receive, extract the `baggage` header and restore it (set `Baggage.Current` / the new activity's baggage) so it is available to the handler and downstream propagation.

Because baggage can carry sensitive or high-cardinality values, this should be:

- **Off by default**, enabled via `NatsInstrumentationOptions` (e.g. `PropagateBaggage = true`), and ideally
- **filterable** via an allow-list of baggage keys (so callers only propagate intended, non-PII keys).

A lighter-weight alternative that would also unblock most users: **expose the parsed/extracted baggage (and the raw headers) on the `Enrich` callback context**, so applications can restore baggage themselves without re-parsing headers.

### Current workaround

Manually inject/extract baggage over `NatsHeaders` using OpenTelemetry's baggage propagator, in parallel with the SDK's automatic trace-context propagation. This works but reimplements part of what the instrumentation already does for trace context.

### Environment

- NATS.Net 3.0 (preview) OpenTelemetry instrumentation (`NATS.Net` ActivitySource).

iamadamreed:
Resolves #1200

cc @mtmk — opening as a draft for review.

## What

The built-in OpenTelemetry instrumentation propagates W3C trace context across publish/consume but drops W3C Baggage at the NATS boundary. This adds **opt-in** baggage propagation:

- `NatsInstrumentationOptions.PropagateBaggage` (default **false**) — when enabled, baggage is injected as a W3C `baggage` header on publish and, on receive, extracted (via `DistributedContextPropagator.Current.ExtractBaggage`, so the legacy `Correlation-Context` fallback also works), restored onto the receive activity, and exposed to the `Filter`/`Enrich` callbacks via a new `NatsInstrumentationContext.Baggage` property.
- `NatsInstrumentationOptions.BaggageKeyFilter` — allow-list predicate applied on both inject and extract, so only intended, non-PII keys cross the boundary.
- `NatsInstrumentationOptions.BaggageSource` — optional send-side source override for applications whose baggage lives outside `Activity.Baggage` (e.g. OpenTelemetry's `Baggage.Current` API, which this library cannot reference).
- `msg.StartActivity()` child spans (core and JetStream) copy the restored baggage across the ActivityContext-parenting gap when the option is enabled.

## Design notes

- **Off by default** because baggage can carry sensitive or high-cardinality values, and ~~servers default to a 4KB~~ JetStream 64KB header limit.
- **Encoding**: keys/values are `Uri.EscapeDataString`-encoded, comma-joined — round-trips through the BCL propagators and is readable by W3C-compliant consumers in other languages.
- **Header ownership** (resolves interplay with the ambient propagator, which on .NET 10's W3C propagator writes an unfiltered `baggage` header during `Inject`): when enabled and the source has baggage, the `baggage` header is overwritten with the filtered set; if the filter rejects every key the header is removed; if the source has no baggage the header is left untouched (preserving application-set values).
- **Zero behavior change when disabled**, including the pre-existing fact that the legacy BCL propagator writes `Correlation-Context` from activity baggage; this is called out in the docs.
- `NatsInstrumentationContext` gains an init-only property (positional ctor untouched); note record-struct equality now includes it.

## Testing

- 9 new integration tests (`tests/NATS.Net.OpenTelemetry.Tests/OpenTelemetryBaggageTest.cs`) covering off-by-default, round-trip, inject/extract filtering, W3C encoding, Enrich exposure, header removal, `BaggageSource`, and child-span copy — full project suite 15/15 green.
- 11 unit tests (`tests/NATS.Client.Core.Tests/TelemetryBaggageTest.cs`) for encoding/ownership edge cases (null values, empty keys, read-only headers, source-returns-null, app-set header preservation).
- All four TFMs build with 0 warnings; JetStream `PublishRetryTest` regression green; docs example compiles.

## Docs

New "Baggage Propagation" section in `documentation/advanced/opentelemetry.md` with a compiled snippet in `NATS.Net.DocsExamples`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/NATS.Client.Core/Internal/Telemetry.cs
+++ b/src/NATS.Client.Core/Internal/Telemetry.cs
@@ -1,5 +1,6 @@
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
+using System.Text;
 
 namespace NATS.Client.Core.Internal;
 
@@ -50,6 +51,37 @@ internal static class Telemetry
         NatsMeter.CreateCounter<long>("nats.client.messages.dropped", unit: "{message}");
 
     private static readonly object BoxedTrue = true;
+
+    private static readonly DistributedContextPropagator.PropagatorGetterCallback HeadersGetter =
+        static (object? carrier, string fieldName, out string? fieldValue, out IEnumerable<string>? fieldValues) =>
+        {
+            if (carrier is not NatsHeaders headers)
+            {
+                Debug.Assert(false, "This should never be hit.");
+                fieldValue = null;
+                fieldValues = null;
+                return;
+            }
+
+            if (headers.TryGetValue(fieldName, out var values))
+            {
+                if (values.Count == 1)
+                {
+                    fieldValue = values[0];
+                    fieldValues = null;
+                }
+                else
+                {
+                    fieldValue = null;
+                    fieldValues = values;
+                }
+            }
+            else
+            {
+                fieldValue = null;
+                fieldValues = null;
+            }
+        };
 
     /// <summary>
     /// Don't use this for metrics.
@@ -136,7 +168,8 @@ internal static class Telemetry
             Connection: connection,
             ParentContext: parentContext);
 
-        if (NatsInstrumentationOptions.Default.Filter is { } filter && !filter(instrumentationContext))
+        var options = NatsInstrumentationOptions.Default;
+        if (options.Filter is { } filter && !filter(instrumentationContext))
             return null;
 
         KeyValuePair<string, object?>[] tags;
@@ -187,7 +220,7 @@ internal static class Telemetry
             tags: tags);
 
         if (activity is not null)
-            NatsInstrumentationOptions.Default.Enrich?.Invoke(activity, instrumentationContext);
+            options.Enrich?.Invoke(activity, instrumentationContext);
 
         return activity;
     }
@@ -211,6 +244,10 @@ internal static class Telemetry
 
                 headers[fieldName] = fieldValue;
             });
+
+        var options = NatsInstrumentationOptions.Default;
+        if (options.PropagateBaggage)
+            InjectBaggage(activity, headers, options);
     }
 
     public static Activity? StartReceiveActivity(
@@ -230,6 +267,11 @@ internal static class Telemetry
         if (headers is null || !TryParseTraceContext(headers, out var context))
             context = default;
 
+        var options = NatsInstrumentationOptions.Default;
+        IReadOnlyList<KeyValuePair<string, string?>>? baggage = null;
+        if (options.PropagateBaggage && headers is not null)
+            baggage = ExtractBaggage(headers, options.BaggageKeyFilter);
+
         var instrumentationContext = new NatsInstrumentationContext(
             Subject: subject,
             Headers: headers,
@@ -238,9 +280,10 @@ internal static class Telemetry
             BodySize: bodySize,
             Size: size,
             Connection: connection,
-            ParentContext: context);
+            ParentContext: context)
+        { Baggage = baggage };
 
-        if (NatsInstrumentationOptions.Default.Filter is { } filter && !filter(instrumentationContext))
+        if (options.Filter is { } filter && !filter(instrumentationContext))
             return null;
 
         KeyValuePair<string, object?>[] tags;
@@ -309,7 +352,17 @@ internal static class Telemetry
             tags: tags);
 
         if (activity is not null)
-            NatsInstrumentationOptions.Default.Enrich?.Invoke(activity, instrumentationContext);
+        {
+            if (baggage is not null)
+            {
+                // Propagators return entries in reverse header order by design so that
+                // AddBaggage's prepend semantics restore the original order.
+                foreach (var item in baggage)
+                    activity.AddBaggage(item.Key, item.Value);
+            }
+
+            options.Enrich?.Invoke(activity, instrumentationContext);
+        }
 
         return activity;
     }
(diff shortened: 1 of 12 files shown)


mtmk:
thanks @iamadamreed. not too sure if there are conflicts but we have a few otel changes in release/3.0 so probably better to based on release/3.0 instead of main. we will merge release/3.0 into main fairly soon (in a few days hopefully)

iamadamreed:
Rebased onto release/3.0 and retargeted the PR — conflicts were minor (adopted the volatile backing-field pattern for the new options, kept LowCardinalitySubject/MeterTracker alongside the new helpers, and switched header writes to the plain indexer since the read-only latch is gone on this line).

One thing worth flagging from the rebase: the OTel test project now resolves System.Diagnostics.DiagnosticSource 10.0.0 (via OpenTelemetry 1.15.3), where the default DistributedContextPropagator is the W3C one — so an unfiltered `baggage` header already appears on publish from ambient Activity baggage regardless of this feature. `PropagateBaggage` takes ownership of that header when enabled (so `BaggageKeyFilter` applies), restores baggage on receive, and the docs note now covers both propagator generations. Also put the new baggage tests in a shared xunit collection with NatsInstrumentationExtensionsTest since both mutate NatsInstrumentationOptions.Default.

Full suite green on the new base: 51/51 in NATS.Net.OpenTelemetry.Tests, 10/10 unit tests, JetStream PublishRetryTest regression passing.

mtmk:
hey @iamadamreed, could you also sign your commits please?

iamadamreed:
Done — the commit is now signed (SSH signing key) and shows as Verified. Same content as before, re-pushed as 8cb7690.

iamadamreed:
CI is green except one infra flake: the **Test Linux Core / Linux (v2.12)** job failed during NuGet restore (`NU1301: api.nuget.org service index … not a valid JSON object`) — nuget.org hiccup, nothing built or ran on that leg. The other matrix legs and all other workflows pass, including the previously-failing Format and Test Windows. Could you re-run that one failed job when you get a chance? (No push access here to re-run it myself.)

src/NATS.Client.Core/Internal/Telemetry.cs
```diff
@@ -368,39 +432,71 @@ static string GetStackTrace(Exception? exception)
...
+
+                sb.Append(Uri.EscapeDataString(item.Key)).Append('=').Append(Uri.EscapeDataString(item.Value));
+            }
+        }
+
+        if (sb is not null)
+        {
+            headers[Constants.BaggageHeader] = sb.ToString();
+        }
+        else if (hadBaggage)
```
mtmk: When `BaggageSource` is configured but yields nothing, this branch is skipped (`hadBaggage` is false), so the `baggage` header the ambient W3C propagator wrote from `activity.Baggage` survives unfiltered. Verified on the wire: filter allowing only `tenant.id`, `BaggageSource` returning empty, send activity carrying `user.email` -> `baggage: user.email=..., tenant.id=...`. The receive side re-applies the filter, so it's invisible from the client.

The point of `BaggageSource` is to let apps whose baggage lives in OTel's `Baggage.Current` (separate from `Activity.Baggage`) use it as the source. Once that's set it should be the single source of truth - if `Baggage.Current` is empty for a message, we shouldn't leak whatever happens to be in `Activity.Baggage`. It's a narrow divergence, but the failure is a silent PII leak.

Suggest treating a configured `BaggageSource` as authoritative:

```csharp
else if (hadBaggage || options.BaggageSource is not null)
```

Leaves the no-source app-set-header passthrough unchanged.

mtmk (commented):
Ran this against a local server (core and JetStream). One inline note on `InjectBaggage`. Two more:

**Size limits.** The "4KB header limit" in the docs and description isn't accurate; 4096 is `MAX_CONTROL_LINE_SIZE` (the control line), not the header block. Actual: core headers count toward `max_payload` (1MB default; oversized publish throws `NatsPayloadTooLargeException` client-side), and JetStream caps the header block at 64KB from the u16 length encoding, returned as an ack error (`err_code 10097`). Worth stating the 64KB one since baggage rides on JetStream headers.

**Legacy propagator NOTE.** Correct but narrow: it only applies with `DiagnosticSource` < 10, and a current OpenTelemetry SDK pulls in DiagnosticSource 10 (W3C default), where `BaggageKeyFilter` applies on the wire. Fine as documented.

Otherwise works as intended on the W3C path.

mtmk (approved):
LGTM thanks @iamadamreed

PR #1210, opened 2026-07-09
Clarify AddNats vs AddNatsClient

mtmk:
Documents the two DI entry points that are easy to confuse: AddNatsClient (ad hoc JSON, registers INatsClient) and AddNats (minimal, AOT-friendly, no JSON serializer). Adds inline docs to the previously undocumented builder API, and removes the keyed-services usage example from the platform compatibility page that duplicated the dependency injection page.

Addresses questions raised in discussion #1209.


piotrpio (commented):
LGTM!

PR #1213, opened 2026-07-15
Add explicit permissions to CI workflows

mtmk:
Adds a `permissions: contents: read` block to the five workflows CodeQL flagged for missing explicit permissions (Test Windows, Test Linux, Test Linux Core, Perf, Format). None of them write to the repo, so this just scopes the default `GITHUB_TOKEN` down to read-only, matching the pattern already used in `apicompat.yml`, `release.yml`, and `signoffs.yml`.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -3,6 +3,9 @@
 on:
   pull_request: {}
 
+permissions:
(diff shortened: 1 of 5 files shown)


scottf (approved):
LGTM

PR #1214, opened 2026-07-16
Fix silent completion when JetStream list enumeration is cancelled

### LINKED ISSUE #1193: ListStreamsAsync should throw if operation is cancelled
### Proposed change

Currently, `ListStreamsAsync` and alike operations only throw an `OperationCanceledException` if cancellation is requested during an in-flight request. If cancellation is requested before or after that instead, these methods return silently. That's the case, for example, when cancellation is requested during pagination.
```csharp
 public async IAsyncEnumerable<INatsJSStream> ListStreamsAsync(
        string? subject = default,
        [EnumeratorCancellation] CancellationToken cancellationToken = default)
    {
        var offset = 0;
        while (!cancellationToken.IsCancellationRequested) // <- here!
        {
            var response = await JSRequestResponseAsync<StreamListRequest, StreamListResponse>(
                subject: $"{Opts.Prefix}.STREAM.LIST",
                request: new StreamListRequest
                {
                    Offset = offset,
                    Subject = subject,
                },
                cancellationToken);

            if (response.Streams.Count == 0)
                yield break;

            foreach (var stream in response.Streams)
                yield return new NatsJSStream(this, stream);

            offset += response.Streams.Count;
        }
    }
```

While that is a fair approach - as per [Microsoft documentation](https://learn.microsoft.com/en-gb/dotnet/standard/parallel-programming/task-cancellation), it is in fact valid to handle cancellation by just returning - it provides callers with a hard-to-interpret API. Upon successfully receiving a list of streams, they would in fact need to check if the provided cancellation token was signalled, to understand if the list is complete and therefore reliable. In my fairly limited experience with `IAsyncEnumerable`s, it is also untypical for enumeration to complete seemingly successfully upon cancellation. So all in all, in my view, it would be nicer to always throw an `OperationCancelledException` instead.

### Use case

N/A

### Contribution

A possible implementation could look like this:
```csharp
 public async IAsyncEnumerable<INatsJSStream> ListStreamsAsync(
        string? subject = default,
        [EnumeratorCancellation] CancellationToken cancellationToken = default)
    {
        var offset = 0;
        while (true)
        {
            cancellationToken.ThrowIfCancellationRequested();
            var response = await JSRequestResponseAsync<StreamListRequest, StreamListResponse>(
                subject: $"{Opts.Prefix}.STREAM.LIST",
                request: new StreamListRequest
                {
                    Offset = offset,
                    Subject = subject,
                },
                cancellationToken);

            if (response.Streams.Count == 0)
                yield break;

            foreach (var stream in response.Streams)
                yield return new NatsJSStream(this, stream);

            offset += response.Streams.Count;
        }
    }
```

mtmk:
JetStream paginated list operations (streams, stream names, consumers, consumer names) used the cancellation token only as a loop guard, so cancelling before the first request or between pages ended the enumeration silently instead of throwing. Callers could not distinguish a complete list from one truncated by cancellation. These now throw OperationCanceledException, matching the conventional IAsyncEnumerable contract and the watcher-based reads which already throw on cancel by default. The KV bucket and status listings inherit this via their delegation to the stream-name listing. Long-running consume/fetch loops are unchanged, since cancellation there is the normal stop signal.

Fixes #1193

### DIFF
--- a/src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
+++ b/src/NATS.Client.JetStream/NatsJSContext.Consumers.cs
@@ -73,7 +73,11 @@ public partial class NatsJSContext : INatsJSContext
         var offset = 0;
         while (true)
         {
-            cancellationToken.ThrowIfCancellationRequested();
+            if (Opts.ThrowOnListCancellation)
+                cancellationToken.ThrowIfCancellationRequested();
+            else if (cancellationToken.IsCancellationRequested)
+                yield break;
+
             var response = await JSRequestResponseAsync<ConsumerListRequest, ConsumerListResponse>(
                 subject: $"{Opts.Prefix}.CONSUMER.LIST.{stream}",
                 new ConsumerListRequest { Offset = offset },
@@ -100,7 +104,11 @@ public partial class NatsJSContext : INatsJSContext
         var offset = 0;
         while (true)
         {
-            cancellationToken.ThrowIfCancellationRequested();
+            if (Opts.ThrowOnListCancellation)
+                cancellationToken.ThrowIfCancellationRequested();
+            else if (cancellationToken.IsCancellationRequested)
+                yield break;
+
             var response = await JSRequestResponseAsync<ConsumerNamesRequest, ConsumerNamesResponse>(
                 subject: $"{Opts.Prefix}.CONSUMER.NAMES.{stream}",
                 new ConsumerNamesRequest { Offset = offset },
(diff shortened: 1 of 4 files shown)


mtmk:
@claude please re-review the latest commit. The approach changed since your last review: throwing on list cancellation is now opt-in via `NatsJSOpts.ThrowOnListCancellation`, and the default keeps the 3.0.0 silent-completion behaviour so upgrading the library changes no observable behaviour. Focus on backward compatibility and whether the opt-in flag is reachable through the usual context-creation paths.

scottf (approved):
LGTM

PR #1215, opened 2026-07-17
Release 3.0.1

mtmk:
Patch release on top of 3.0.0. Adds opt-in W3C Baggage propagation for OpenTelemetry and fixes a JetStream list enumeration that could complete silently when cancelled.

- Fix silent completion when JetStream list enumeration is cancelled (#1214)
- OpenTelemetry: opt-in W3C Baggage propagation (#1208)
- ci: set explicit read-only permissions on test/perf/format workflows (#1213)


Jarema (approved):
LGTM!

PR #1218, opened 2026-07-24
Add subject-dispatch serialization example

### LINKED ISSUE #732: Integrate headers into the NatsSerializer API
### Proposed change

Adapt the `INatsDeserialize` and `INatsSerialize` interfaces in a way that makes it possible to access the headers during deserialization and set them during serialization. E.g. like this
```csharp

public interface INatsSerialize<in T>
{
    void Serialize(IBufferWriter<byte> bufferWriter, T value);
    void Serialize(IBufferWriter<byte> bufferWriter, T value, out NatsHeaders headers);
}

public interface INatsDeserialize<out T>
{
    T? Deserialize(in ReadOnlySequence<byte> buffer);
    T? Deserialize(in ReadOnlySequence<byte> buffer, NatsHeaders headers);
}
```
One could add a default implementation for the new method to call the old method without forwarding the headers to ensure backwards compatibility. I'm not sure if there is a way to do this and still have the option to only implement the new method and not both.

Alternatively, one could also provide the whole message object to be able to access everything necessary (maybe the subject is also interesting?). However, that open the object to be changed in non-obvious ways.

### Use case

With this change the headers can be used to change the behavior of the deserialization (e.g. which data format was used, which encoding was used and so on). Of course, this would need the headers to also be correctly set during serialization. An example where this would be helpful is the proposal for usage of CloudEvents with NATS in the binary mode: https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/nats-protocol-binding.md#31-binary-content-mode

With the current state, it is not possible for the Headers to be accessed during (de-)serialization. If one wants to implement cloudevents protocol bindings, this needs to happen in a separate step to the deserialization even though it is deserialization itself. To use them as conveniently as normal serialization, one needs to add wrappers for consumers and extension methods for subscribe, publish and so forth.

### Contribution

I'd probably be able to implement it, but I'd want some tips on how this can best be done backwardscompatibly.

mtmk:
Expands the message context section of the serialization docs with a subject-dispatch deserializer example, the use case raised in the linked issue, and notes that the library does not allocate a headers instance during publish, so context-aware serializers can only add headers when the caller passes an instance in.

Fixes #732

### DIFF
--- a/src/NATS.Client.Abstractions/INatsSerialize.cs
+++ b/src/NATS.Client.Abstractions/INatsSerialize.cs
@@ -20,6 +20,7 @@ public interface INatsSerializer<T> : INatsSerialize<T>, INatsDeserialize<T>
 /// Serializer interface for NATS messages.
 /// </summary>
 /// <typeparam name="T">Serialized object type</typeparam>
+/// <seealso cref="INatsSerializeWithContext{T}"/>
 public interface INatsSerialize<in T>
 {
     /// <summary>
@@ -34,6 +35,7 @@ public interface INatsSerialize<in T>
 /// Deserializer interface for NATS messages.
 /// </summary>
 /// <typeparam name="T">Deserialized object type</typeparam>
+/// <seealso cref="INatsDeserializeWithContext{T}"/>
 public interface INatsDeserialize<out T>
 {
     /// <summary>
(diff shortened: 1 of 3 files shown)


scottf (approved):
LGTM

PR #1220, opened 2026-07-27
Audit transitive dependencies

mtmk:
NuGet audit runs in `direct` mode by default, so vulnerable packages reached through a dependency are never reported. Enabling full-graph auditing surfaced two vulnerable packages, both in test, sandbox and tooling projects rather than anything shipped. The first commit clears them, the second turns the audit on. Dependabot has the same blind spot, since GitHub's NuGet dependency graph for this repo resolves direct references only without lock files.

Audit only runs at restore, so a branch nobody builds is never checked. The scheduled job covers that. It starts on `main` only; `release/3.0` needs the same dependency fixes backported before it can be added to the matrix.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,50 @@
+name: Dependency Audit
+
+on:
+  schedule:
+    # Mondays 06:00 UTC
+    - cron: '0 6 * * 1'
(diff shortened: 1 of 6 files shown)


scottf (approved):
LGTM

PR #1231, opened 2026-08-06
Feat: add public push consumer API

### LINKED ISSUE #686: Implement Push Consumers
### Discussed in https://github.com/nats-io/nats.net/discussions/685

<sup>Originally posted by **yang-xiaodong** December  2, 2024</sup>
I see that push-based consumers are not currently supported, and I was wondering if pull-based consumers support DeliverGroup for effects similar to Kafka Consumer Groups (A event sent to multiple groups, with load balancing of consumers within groups).</div>

We need to implement push consumers. API wise i suggest a new method similar to ordered consumers:
```csharp
ValueTask<INatsJSPushConsumer> CreatePushConsumerAsync(...);
```

Ivandemidov00:
Adds INatsJSPushConsumer : INatsJSConsumer, NatsJSPushConsumer, NatsJSOrderedPushConsumer, NatsJSPushConsumerOpts, and four factory methods on INatsJSContext. Pull operations throw NatsJSProtocolException on push consumers (priority groups are pull-only per ADR-42).

26 integration tests included.

If accepted, NatsObjStore migration to the public API will follow in a separate PR. KV watcher is not a migration candidate (KV-specific logic).

Resolves #686

@mtmk


mtmk:
thanks @Ivandemidov00 this one has been in the pipeline for a long time really appreciate the contribution. I just fixed the sample to make the ci green. will take a proper look this week.

src/NATS.Client.JetStream/NatsJSContext.PushConsumers.cs
```diff
@@ -0,0 +1,122 @@
...
+            Name = opts.Name,
+            DurableName = opts.DurableName,
+            Description = opts.Description,
+            FilterSubject = opts.FilterSubject,
+            DeliverPolicy = opts.DeliverPolicy,
+            AckPolicy = opts.AckPolicy,
+            IdleHeartbeat = opts.IdleHeartbeat,
+            FlowControl = opts.FlowControl,
+            HeadersOnly = opts.HeadersOnly,
+            DeliverSubject = NewBaseInbox(),
```
mtmk, line 78: We need to expose this as `DeliverSubject` in the options. it's callers responsibility to set it. we should also validate here that it is set as well.

src/NATS.Client.JetStream/NatsJSOrderedPushConsumer.cs
```diff
@@ -0,0 +1,177 @@
...
+                yield break;
+
+            var internalOpts = new NatsJSOrderedPushConsumerOpts
+            {
+                DeliverPolicy = _opts.DeliverPolicy,
+                OptStartSeq = _opts.OptStartSeq,
+                OptStartTime = _opts.OptStartTime,
+                ReplayPolicy = _opts.ReplayPolicy,
+                HeadersOnly = _opts.HeadersOnly,
+                FilterSubjects = _opts.FilterSubjects,
```
mtmk, line 74: we should not set this if `_opts.FilterSubjects` is empty. also some of the other options should be guarded as well. could you check the pull consumer version and use the same pattern?

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// </remarks>
+    public TimeSpan? InactiveThreshold { get; init; }
+
+    /// <summary>
+    /// Amount of idle time the server should wait before sending a heartbeat.
+    /// </summary>
+    /// <remarks>
+    /// Defaults to 5 seconds.
+    /// </remarks>
+    public TimeSpan IdleHeartbeat { get; init; } = TimeSpan.FromSeconds(5);
```
mtmk, line 115: Can we make this nullable and not set it here to match Go client's defaults?

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// <summary>
+    /// Sends a control message to the consumer's deliver subject to regulate the rate of message delivery.
+    /// </summary>
+    /// <remarks>
+    /// When <c>true</c> the server will send flow control messages with a command status
+    /// header and a reply subject when the consumer's delivery buffer is getting full. The
+    /// client must respond to these messages to continue receiving messages. Defaults to
+    /// <c>true</c>.
+    /// </remarks>
+    public bool FlowControl { get; init; } = true;
```
mtmk, line 126: Same here. Go client's default is no flow control as default.

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// <summary>
+    /// Options affecting the underlying subscription created for the delivery subject.
+    /// </summary>
+    public NatsSubOpts? SubOpts { get; init; }
+
+    /// <summary>
+    /// A handler function invoked for notifications related to consumption, allowing custom handling of
+    /// operational updates or state changes during the consumer's lifecycle.
+    /// </summary>
+    public Func<INatsJSNotification, CancellationToken, Task>? NotificationHandler { get; init; }
```
mtmk, line 151: This doesn't seem to be used.

mtmk (commented):
Thanks @Ivandemidov00 looking good, a few comments blow.

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
Ivandemidov00: Wired through in 2e78941. The handler is passed to NatsJSPushConsume

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
Ivandemidov00: Change to default false in 2e78941

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
Ivandemidov00: IdleHeartbeat is now TimeSpan? nullable, no default. The factory only sends it to the server config when it's explicitly provided

src/NATS.Client.JetStream/Internal/NatsJSOrderedPushConsumer.cs
```diff
@@ -34,7 +33,10 @@ internal record NatsJSOrderedPushConsumerOpts
     /// <summary>
     /// Idle heartbeat interval
     /// </summary>
-    public TimeSpan IdleHeartbeat { get; init; } = TimeSpan.FromSeconds(5);
+    /// <remarks>
+    /// When not set, the server does not send heartbeats.
+    /// </remarks>
+    public TimeSpan? IdleHeartbeat { get; init; } = TimeSpan.FromSeconds(5);
```
mtmk, line 39: ~~default should be null~~ edit: ah sorry didn't realized this is internal. ignore this plese.

src/NATS.Client.JetStream/NatsJSContext.PushConsumers.cs
```diff
@@ -0,0 +1,133 @@
...
+        return new ValueTask<INatsJSPushConsumer>(new NatsJSOrderedPushConsumer(stream, this, opts, cancellationToken));
+    }
+
+    private async ValueTask<INatsJSPushConsumer> CreatePushConsumerInternalAsync(
+        string stream,
+        NatsJSPushConsumerOpts opts,
+        ConsumerCreateAction action,
+        CancellationToken cancellationToken)
+    {
+        var deliverSubject = opts.DeliverSubject ?? NewBaseInbox();
```
mtmk: not providing subject should be an error.

Ivandemidov00: Done CreatePushConsumerInternalAsync now throws NatsJSException

src/NATS.Client.JetStream/NatsJSPushConsumer.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        var deliverSubject = Info.Config.DeliverSubject;
+        if (deliverSubject is not { Length: > 0 })
+        {
+            throw new NatsJSException($"Consumer '{_stream}:{_consumer}' doesn't have a deliver subject so it can't be consumed as a push consumer.");
+        }
+
+        serializer ??= _context.Connection.Opts.SerializerRegistry.GetDeserializer<T>();
+
+        var idleHeartbeat = Info.Config.IdleHeartbeat == TimeSpan.Zero ? TimeSpan.FromSeconds(5) : Info.Config.IdleHeartbeat;
```
mtmk, line 93: **[bug]** `ConsumeAsync` substitutes a 5s client idle heartbeat whenever `Info.Config.IdleHeartbeat` is `TimeSpan.Zero`. That undoes the factory change that omits `IdleHeartbeat` unless the caller set it (Go default: no server heartbeats). After the first data message arms the timer, 10s of silence fires `NatsJSTimeoutNotification` even though the server is not sending heartbeats. Starting the timer at subscribe time (Issue 2) without removing this fallback would make the false timeout fire 10s after `ConsumeAsync` even with no messages.

**Suggestion:** Pass `Info.Config.IdleHeartbeat` through as-is. Only run the heartbeat timer when that value is greater than zero. Do not invent a consume-time default that the consumer was not created with.

src/NATS.Client.JetStream/NatsJSPushConsumer.cs
```diff
@@ -0,0 +1,227 @@
...
+            context: _context,
+            subject: deliverSubject,
+            queueGroup: Info.Config.DeliverGroup,
+            idleHeartbeat: idleHeartbeat,
+            notificationHandler: _notificationHandler ?? opts?.NotificationHandler,
+            serializer: serializer,
+            opts: subOpts,
+            cancellationToken: cancellationToken);
+
+        await _context.Connection.AddSubAsync(sub: sub, cancellationToken).ConfigureAwait(false);
```
mtmk, line 107: **[bug]** After `AddSubAsync`, pull consume calls `ResetHeartbeatTimer()` immediately. Push consume never does; the timer is armed only from `ReceiveInternalAsync` (first message or heartbeat). If `IdleHeartbeat` is configured and nothing arrives (deleted consumer, wrong deliver subject, dropped first heartbeat), consume hangs until the caller's cancellation token fires. The timer is also one-shot (`Change(due, Infinite)` at `NatsJSPushConsume.cs:111`) and the callback returns without re-arming when `ConnectionState == Reconnecting` (`NatsJSPushConsume.cs:65`), so a timeout during reconnect disables detection until the next message.

**Suggestion:** After a successful `AddSubAsync`, call `sub.ResetHeartbeatTimer()` when idle heartbeat is configured. Re-arm in the callback (or use a periodic timer like `NatsJSConsume`) and do not drop the timer on the reconnect path.

src/NATS.Client.JetStream/Internal/NatsJSPushConsume.cs
```diff
@@ -0,0 +1,245 @@
...
+            }
+            else
+            {
+                _logger.LogWarning(NatsJSLogEvents.ProtocolMessage, "Protocol error: No header found in control message");
+            }
+        }
+        else
+        {
+            var msg = new NatsJSMsg<T>(
+                NatsMsg<T>.Build(
```
mtmk, line 204: **[bug]** Data messages use obsolete `NatsMsg<T>.Build` and never assign `ReceiveActivity`. `NatsJSConsume` / `NatsJSFetch` / the internal ordered push sub use `BuildInternal(..., subscriptionSubject: Subject, queueGroup: QueueGroup)` and set `ReceiveActivity` before `WriteAsync` so `NatsSubBase.ReceiveAsync` can end the span if handoff fails. Here a failed `WriteAsync` leaves the receive activity running. The flow-control branch at line 165 also calls `Build` to `ReplyAsync`; that activity is never handed to `ActivityEndingMsgReader` and is not ended. Reply with `PublishAsync(replyTo)` like the `Nats-Consumer-Stalled` path instead of building a user message.

**Suggestion:** Mirror `NatsJSConsume.ReceiveInternalAsync`: `BuildInternal` with subscription subject and queue group, `ReceiveActivity = msg.Headers?.Activity` before `WriteAsync`. For flow-control requests, publish to `replyTo` without constructing a `NatsJSMsg`.

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
```diff
@@ -0,0 +1,158 @@
...
+    public string? Name { get; init; }
+
+    /// <summary>
+    /// A unique name for a durable consumer.
+    /// </summary>
+    public string? DurableName { get; init; }
+
+    /// <summary>
+    /// The delivery subject for the push consumer. The server uses this subject
+    /// to deliver messages to the client. When not set, a new inbox is generated.
```
mtmk, line 38: **[suggestion]** Docs still say a missing deliver subject generates an inbox (`NatsJSPushConsumerOpts` class remarks at line 11, this property, and `INatsJSContext.CreatePushConsumerAsync` at line 33: "on a generated subject"). The factory now throws `NatsJSException` if `DeliverSubject` is empty. IntelliSense will tell callers the opposite of what the code does.

**Suggestion:** Document that `DeliverSubject` is required (callers typically pass `connection.NewInbox()`), and drop the generated-subject language from the opts record and `INatsJSContext`.

src/NATS.Client.JetStream/NatsJSOrderedPushConsumer.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            NatsJSOrderedPushConsumer<T>? pushConsumer = null;
+            ChannelReader<NatsJSMsg<T>> reader = default!;
+            var errored = false;
+
+            try
+            {
+                pushConsumer = new NatsJSOrderedPushConsumer<T>(
+                    _context, _stream, string.Empty, serializer, internalOpts, new NatsSubOpts(), cancellationToken);
+                pushConsumer.Init();
```
mtmk, line 86: **[suggestion]** Public ordered consume is a thin loop over the KV/Object Store engine and does not match `NatsJSOrderedConsumer`. `Init()` only queues create on a background loop that logs and swallows errors, so a failed create leaves `WaitToReadAsync` hanging until cancel. `Info` is never assigned after the empty constructor despite the comment that it is filled during consume. The outer retry always rebuilds `NatsJSOrderedPushConsumerOpts` from the original `OptStartSeq` (line 70); if the inner channel ever completes, messages would be re-delivered from the start. `NatsJSConsumeOpts` (notification handler, etc.) is ignored.

**Suggestion:** Await and propagate consumer creation like `NatsJSOrderedConsumer.RecreateConsumer`, publish `Info` from the created consumer, and either drop the outer retry (the inner engine already recreates on sequence mismatch / heartbeat) or track last stream sequence across recreates.

src/NATS.Client.JetStream/NatsJSContext.PushConsumers.cs
```diff
@@ -0,0 +1,131 @@
...
+        NatsJSPushConsumerOpts opts,
+        ConsumerCreateAction action,
+        CancellationToken cancellationToken)
+    {
+        if (string.IsNullOrWhiteSpace(opts.DeliverSubject))
+        {
+            throw new NatsJSException("A push consumer requires a non-empty deliver subject.");
+        }
+
+        var config = new ConsumerConfig
```
mtmk, line 72: **[suggestion]** `NatsJSPushConsumerOpts` / this mapping omit several `ConsumerConfig` fields that pull consumers set directly, notably `ReplayPolicy` (present on `NatsJSOrderedConsumerOpts`), `Backoff`, `Metadata`, `SampleFreq`, `RateLimitBps`, `NumReplicas`, and `MemStorage`. Push users cannot choose original-speed replay or NAK backoff without dropping to raw `ConsumerConfig` APIs that still return `INatsJSConsumer`.

**Suggestion:** Map the remaining push-relevant `ConsumerConfig` properties (at least `ReplayPolicy`, `Backoff`, `Metadata`, `RateLimitBps`) with the same "set only when specified" guards used for `AckWait` / `MaxDeliver`.

src/NATS.Client.JetStream/NatsJSContext.PushConsumers.cs
```diff
@@ -0,0 +1,131 @@
...
+            subject: $"{Opts.Prefix}.CONSUMER.INFO.{stream}.{consumer}",
+            request: null,
+            cancellationToken);
+
+        if (string.IsNullOrWhiteSpace(response.Config.DeliverSubject))
+        {
+            throw new NatsJSException($"Consumer '{stream}:{consumer}' doesn't have a deliver subject so it can't be consuming as a push consumer.");
+        }
+
+        return new NatsJSPushConsumer(this, response);
```
mtmk, line 47: **[suggestion]** `GetPushConsumerAsync` constructs `NatsJSPushConsumer` without `SubOpts` or `NotificationHandler`. Create uses a 1000-message channel (line 127) and can attach a notification handler on the opts. Bind-then-consume therefore gets connection default channel options, and notifications only if passed later via `NatsJSConsumeOpts.NotificationHandler`. `SubOpts` cannot be supplied at consume time.

**Suggestion:** Either add consume-time `NatsSubOpts` (or honor `NatsJSConsumeOpts` more fully) or document that channel capacity / notification handler must be set at create, and that `GetPushConsumerAsync` will not see them.

src/NATS.Client.JetStream/INatsJSContext.cs
```diff
@@ -30,6 +29,79 @@ ValueTask<INatsJSConsumer> CreateOrderedConsumerAsync(
...
+    /// </remarks>
+    /// <param name="stream">Name of the stream to create the consumer under.</param>
+    /// <param name="opts">Push consumer options.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the API call.</param>
+    /// <returns>The NATS JetStream push consumer object which can be used retrieving data from the stream with <see cref="INatsJSConsumer.ConsumeAsync{T}"/>.</returns>
+    /// <exception cref="NatsJSException">There was an issue retrieving the response.</exception>
+    /// <exception cref="NatsJSApiException">Server responded with an error.</exception>
+    /// <exception cref="ArgumentException">The <paramref name="stream"/> name is invalid.</exception>
+    /// <exception cref="ArgumentNullException">The <paramref name="stream"/> name is <c>null</c>.</exception>
+    ValueTask<INatsJSPushConsumer> CreatePushConsumerAsync(
```
mtmk, line 52: **[suggestion]** Push factories are added only on `INatsJSContext`. `INatsJSStream` already has `CreateOrderedConsumerAsync` / `CreateOrUpdateConsumerAsync` / `GetConsumerAsync` wrappers. Callers that hold a stream object still have to drop to the context and re-supply the stream name.

**Suggestion:** Add matching `CreatePushConsumerAsync`, `CreateOrUpdatePushConsumerAsync`, `GetPushConsumerAsync`, and `CreateOrderedPushConsumerAsync` on `INatsJSStream` / `NatsJSStream`, forwarding to the context like the pull ordered factory.

src/NATS.Client.JetStream/Internal/NatsJSPushConsume.cs
```diff
@@ -0,0 +1,245 @@
...
+                    {
+                    }
+                    else if (headers.HasTerminalJSError())
+                    {
+                        _userMsgs.Writer.TryComplete(new NatsJSProtocolException(headers.Code, headers.Message, headers.MessageText));
+                        EndSubscription(NatsSubEndReason.JetStreamError);
+                    }
+                    else if (headers.Code != 100 || headers.Message != NatsHeaders.Messages.Text)
+                    {
+                        _logger.LogWarning(NatsJSLogEvents.ProtocolMessage, "Unhandled control message: {Code} {Description}", headers.Code, headers.MessageText);
```
mtmk, line 188: **[suggestion]** Unhandled control messages are only logged. Pull consume also notifies `NatsJSProtocolNotification` (and treats leadership-change 409 as a notification). Clustered push consumers can see the same 409 on the deliver subject; with a handler attached, push currently stays silent except for a warning log.

**Suggestion:** Notify `NatsJSProtocolNotification` on the unhandled branch, matching `NatsJSConsume`.

src/NATS.Client.JetStream/NatsJSPushConsumer.cs
```diff
@@ -0,0 +1,227 @@
...
+                        {
+                            read = false;
+                            jsMsg = default;
+                        }
+
+                        if (!read)
+                            break;
+
+                        // if yield is blocked by the application code, we don't want
+                        // heartbeat timer kicking in and issuing a spurious flow control.
```
mtmk, line 149: **[nit]** The comment says stopping the heartbeat timer avoids "issuing a spurious flow control." The timer only notifies `NatsJSTimeoutNotification` (and may complete the channel). It does not send flow control; that is a server-to-client request this client replies to.

**Suggestion:** Describe the actual reason, e.g. avoid a spurious idle-timeout notification while the caller is blocked in the `yield return`.

src/NATS.Client.JetStream/NatsJSException.cs
```diff
@@ -52,6 +52,12 @@ public NatsJSProtocolException(int headerCode, NatsHeaders.Messages headerMessag
     public NatsHeaders.Messages HeaderMessage { get; }
 
     public string HeaderMessageText { get; }
+
+    /// <summary>
+    /// Returns an exception describing a push based consumer when a pull based operation is attempted.
+    /// </summary>
+    public static NatsJSProtocolException ConsumerIsPushBased() =>
+        new(100, NatsHeaders.Messages.ConsumerIsPushBased, "Consumer is push based");
```
mtmk, line 60: **[nit]** `ConsumerIsPushBased()` uses header code `100`. The server emits this condition as `409` (`HasTerminalJSError` treats `ConsumerIsPushBased` as a 409). Callers switching on `HeaderCode` will not match the real protocol.

**Suggestion:** Use `409` to match the server and `NatsHeaders.Messages.ConsumerIsPushBased`.

mtmk (commented):
I've done this review with AI.

mtmk (commented):
Thanks for the update, most of the earlier comments are addressed. A few remaining items inline.

None of the behaviour changes in this batch come with tests: `PushConsumerTest.cs`, `PushFlowControlTest.cs` and `OrderedPushConsumerTest.cs` are unchanged. At least these would be worth covering: heartbeat timer fires when nothing arrives after subscribe with `IdleHeartbeat` set; no timeout notification when `IdleHeartbeat` is not set; unhandled control message reaches the notification handler; ordered push consumer create failure surfaces to the caller instead of hanging.

src/NATS.Client.JetStream/Internal/NatsJSPushConsume.cs
```diff
@@ -0,0 +1,261 @@
...
+            }
+            else
+            {
+                _logger.LogWarning(NatsJSLogEvents.ProtocolMessage, "Protocol error: No header found in control message");
+            }
+        }
+        else
+        {
+            var msg = new NatsJSMsg<T>(
+                NatsMsg<T>.BuildInternal(
```
mtmk, line 201: This switched to `BuildInternal` but drops the `subscriptionSubject`/`queueGroup` arguments and still does not assign `ReceiveActivity` before the `WriteAsync`. Without that, `NatsSubBase.ReceiveAsync` cannot end the receive span if the channel write fails, so the leak from the earlier comment is still there. The internal ordered push sub in this same commit does it the right way, so it is just a matter of mirroring that:

```csharp
var msg = new NatsJSMsg<T>(
    NatsMsg<T>.BuildInternal(subject, replyTo, headersBuffer, payloadBuffer, Connection, Connection.HeaderParser, _serializer,
        replyParentContext: default, subscriptionSubject: Subject, queueGroup: QueueGroup),
    _context);
ReceiveActivity = msg.Headers?.Activity;
```

src/NATS.Client.JetStream/Internal/NatsJSPushConsume.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        _timer = new Timer(
+            static state =>
+            {
+                var self = (NatsJSPushConsume<T>)state!;
+
+                if (self._draining)
+                    return;
+
+                if (self.Connection.ConnectionState == NatsConnectionState.Reconnecting)
```
mtmk, line 66: The disconnect/open hooks cover the common reconnect case, thanks. The callback itself is still one-shot though: if it fires while `ConnectionState == Reconnecting` it returns without re-arming, and `ConnectionOpened` only re-arms when `_wasDisconnected` was set by the event. If the timer wins the race before the disconnected event lands, heartbeat detection stays off until the next message. Simplest fix is to re-arm in the callback on that early return (`self.ResetHeartbeatTimer()`), or use a periodic timer like `NatsJSConsume`.

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
```diff
@@ -0,0 +1,182 @@
...
```
mtmk, line 39: Still says "When not set, a new inbox is generated". The factory throws if this is empty, and the record-level remarks above now say so. Suggest: "Required. Subject the server delivers messages to. Callers typically pass `connection.NewInbox()`."

src/NATS.Client.JetStream/INatsJSContext.cs
```diff
@@ -30,6 +29,79 @@ ValueTask<INatsJSConsumer> CreateOrderedConsumerAsync(
         NatsJSOrderedConsumerOpts? opts = default,
         CancellationToken cancellationToken = default);
 
+    /// <summary>
+    /// Creates a new push consumer that delivers messages to a client subscription on a generated subject.
```
mtmk, line 33: Same as the opts comment: "on a generated subject" is no longer true. Something like "delivers messages to the subject given in `NatsJSPushConsumerOpts.DeliverSubject`" and an `<exception cref="NatsJSException">` for the missing subject case.

src/NATS.Client.JetStream/NatsJSContext.PushConsumers.cs
```diff
@@ -0,0 +1,153 @@
...
+        NatsJSPushConsumerOpts? opts = default,
+        CancellationToken cancellationToken = default)
+    {
+        ThrowIfInvalidStreamName(stream);
+        opts ??= NatsJSPushConsumerOpts.Default;
+        return CreatePushConsumerInternalAsync(stream, opts, ConsumerCreateAction.CreateOrUpdate, cancellationToken);
+    }
+
+    /// <inheritdoc />
+    /// <remarks>
```
mtmk, line 31: This points callers at `NatsJSConsumeOpts.SubOpts`, but `NatsJSConsumeOpts` has no `SubOpts` property and `ConsumeAsync` only reads the `_subOpts` field set at construction. Only `NotificationHandler` is picked up at consume time. Either add a `SubOpts` to what `ConsumeAsync` accepts, or change the remark to say channel capacity can only be set at create and is not carried over by `GetPushConsumerAsync`.

src/NATS.Client.JetStream/NatsJSOrderedPushConsumer.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            Name = string.Empty,
+            StreamName = string.Empty,
+        };
+    }
+
+    /// <summary>
+    /// Consumer info object created during consume operations.
+    /// </summary>
+    public ConsumerInfo Info { get; private set; }
```
mtmk, line 48: `Info` is still never assigned after the constructor. The internal engine now exposes `Info` from the created consumer, so this can be set from `pushConsumer.Info` once creation completes (or on each recreate), matching what the property doc says.

src/NATS.Client.JetStream/NatsJSOrderedPushConsumer.cs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// <inheritdoc />
+    public async IAsyncEnumerable<INatsJSMsg<T>> ConsumeAsync<T>(
+        INatsDeserialize<T>? serializer = default,
+        NatsJSConsumeOpts? opts = default,
```
mtmk, line 53: `NatsJSConsumeOpts opts` is unused. At minimum `NotificationHandler` should be forwarded; if the rest cannot apply to a push consumer, document that on the method.

src/NATS.Client.JetStream/NatsJSOrderedPushConsumer.cs
```diff
@@ -0,0 +1,187 @@
...
+        var retry = 0;
+        while (!cancellationToken.IsCancellationRequested)
+        {
+            if (_context.Connection is NatsConnection { IsDisposed: true })
+                yield break;
+
+            var internalOpts = new NatsJSOrderedPushConsumerOpts
+            {
+                DeliverPolicy = _opts.DeliverPolicy,
+                OptStartSeq = _opts.OptStartSeq,
```
mtmk, line 70: Creation failures now propagate from the engine, good. The outer loop still rebuilds `internalOpts` from the original `OptStartSeq` on every iteration, so if the inner channel completes and we loop, delivery restarts from the beginning. Either track the last seen stream sequence and start from `seq + 1`, or drop the outer retry since the engine already recreates on sequence mismatch and heartbeat timeout.

src/NATS.Client.JetStream/Internal/NatsJSPushConsume.cs
Ivandemidov00: Passed subscriptionSubject: Subject and queueGroup: QueueGroup to NatsMsg<T>.BuildInternal and assigned ReceiveActivity = msg.Headers?.Activity before WriteAsync, mirroring NatsJSOrderedPushConsumerSub<T>.ReceiveInternalAsync.

src/NATS.Client.JetStream/Internal/NatsJSPushConsume.cs
Ivandemidov00: Changed ResetHeartbeatTimer from one shot to periodic: _timer.Change(_hbTimeout, _hbTimeout), matching NatsJSConsume. The early return on Reconnecting is left in place; the periodic timer self-heals after reconnect, so heartbeat detection is not disabled by a race.

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
Ivandemidov00: Updated the DeliverSubject doc to state it is required, that the factory throws NatsJSException when empty, and that callers typically pass connection.NewInbox() (linked as INatsJSContext.NewBaseInbox)

src/NATS.Client.JetStream/INatsJSContext.cs
Ivandemidov00: Rewrote the CreatePushConsumerAsync summary to reference NatsJSPushConsumerOpts.DeliverSubject and added <exception cref="NatsJSException"> for the missing-subject case on both CreatePushConsumerAsync and CreateOrUpdatePushConsumerAsync.

src/NATS.Client.JetStream/NatsJSContext.PushConsumers.cs
Ivandemidov00: Removed the stale reference to NatsJSConsumeOpts.SubOpts. Remarks now state that channel capacity is set at create time via NatsJSPushConsumerOpts.SubOpts, and that NotificationHandler can be supplied at consume time via NatsJSConsumeOpts.NotificationHandler

src/NATS.Client.JetStream/NatsJSOrderedPushConsumer.cs
Ivandemidov00: Added Action<ConsumerInfo>? OnConsumerCreated to the internal NatsJSOrderedPushConsumerOpts. The engine invokes it after Info = consumerInfo.Info on every creation/recreation, and the public class passes info => Info = info

src/NATS.Client.JetStream/NatsJSOrderedPushConsumer.cs
Ivandemidov00: Forwarded opts?.NotificationHandler into the internal opts. The engine creates a NatsJSNotificationChannel and sends NatsJSTimeoutNotification.Default from the heartbeat timer callback.

src/NATS.Client.JetStream/NatsJSOrderedPushConsumer.cs
Ivandemidov00: Tracked ulong seq = metadata.Sequence.Stream inside the TryRead loop. When seq > 0 and the outer loop iterates, internalOpts is rebuilt via with { OptStartSeq = seq + 1, DeliverPolicy = ByStartSequence, OptStartTime = default }, matching NatsJSOrderedConsumer.RecreateConsumer and avoiding duplicate delivery from the start of the stream.

tests/NATS.Client.JetStream.Tests/PushFlowControlTest.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        var js = new NatsJSContext(nats);
+        await js.CreateStreamAsync($"{prefix}s1", [$"{prefix}s1.*"], cts.Token);
+
+        var consumer = await js.CreatePushConsumerAsync(
+            $"{prefix}s1",
+            new NatsJSPushConsumerOpts
+            {
+                Name = $"{prefix}c1",
+                DeliverSubject = nats.NewInbox(),
```
mtmk, line 25: I think this should fix the tests: we need to use `js.NewBaseInbox()` here otherwise it goes to inbox multiplexer. API difference compared to Go.

mtmk:
once this test fix #1262 is merged, merging main should give us a full green ci. (and the inbox test fix above)

tests/NATS.Client.JetStream.Tests/PushFlowControlTest.cs
mtmk: To expand on the Go remark, i think it might be ambiguous:

`js.NewBaseInbox()` is the .NET equivalent of Go's `nc.NewInbox()`, so this isn't a case of the two clients differing in behavior; it's the fact that method names are different.

Go's `nc.NewInbox()` returns `_INBOX.<nuid>`, and the request-reply muxer builds its wildcard from that.

.NET's `NewInbox()` returns `_INBOX.<connection-nuid>.<nuid>`, so it's under the muxers `_INBOX.<connection-nuid>.*` already, because that nesting is how replies get demultiplexed.

`NewBaseInbox()` is the one that gives you a sibling `_INBOX.<nuid>`. we should've named them the same really, but now it'd be breaking change.

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
```diff
@@ -0,0 +1,183 @@
...
+    /// <summary>
+    /// A unique name for a durable consumer.
+    /// </summary>
+    public string? DurableName { get; init; }
+
+    /// <summary>
+    /// The subject the server delivers messages to. Required: consumer creation throws
+    /// <see cref="NatsJSException"/> when it's not set. Callers typically pass an inbox
+    /// generated with <see cref="INatsJSContext.NewBaseInbox"/>.
+    /// </summary>
```
mtmk, line 41: perhaps add a warning about not using NewInbox():
```
    /// <remarks>
    /// Do not pass <c>NatsConnection.NewInbox()</c>. It returns a subject underneath the
    /// connection's request-reply inbox prefix, which the reply multiplexer is already
    /// subscribed to, so no messages reach the consumer and nothing is logged. This differs
    /// from the Go client, where <c>nc.NewInbox()</c> is a valid deliver subject; the .NET
    /// equivalent of that call is <see cref="INatsJSContext.NewBaseInbox"/>.
    /// </remarks>
```

mtmk:
#1262 is just merged

src/NATS.Client.JetStream/NatsJSPushConsumerOpts.cs
Ivandemidov00: Added remarks

tests/NATS.Client.JetStream.Tests/PushFlowControlTest.cs
Ivandemidov00: Changed nats.NewInbox() to js.NewBaseInbox()

mtmk (approved):
LGTM thanks @Ivandemidov00 for this long awaited feature implmentation!

PR #1234, opened 2026-08-07
Release 3.1.1-preview.1

mtmk:
Preview on the 3.1 patch line so the receive activity leak fix can be tested before the stable 3.1.1 release.

- Fix receive activity leak when tracing is enabled (#1229)


scottf (approved):
LGTM

Issue #941, opened 2025-08-28
Object store rents 128k ends up in LOH


mtmk:
### Proposed change

When saving (and potentially loading) objects from object store we allocate 128K chunks of memory from array pool for publishing. This would create arrays larger than 85K ending up in LOH. We need to be able to write multiple slices of memory to socket when publishing or read on subscriptions. A memory sequence type can be used for this at least for binary serialization.

### Use case

Avoid LOH

MiloszKrajewski:
I kind of gave you idea which is not that simple.
According to [this](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream) it is actually desired behaviour.

> RecyclableMemoryStream improves GC performance by ensuring that the larger buffers used for the streams are put into the gen 2 heap and stay there forever. This should cause full collections to happen less frequently. If you pick buffer sizes above 85,000 bytes, then you will ensure these are placed on the large object heap, which is touched even less frequently by the garbage collector.

YMMV and "rule no 1 of performance is 'measure'"

PR #1235, opened 2026-08-12
Fix inverted array check in object store put path

fresh55:
Noticed this while reading the put path for #941.

The netstandard2.0 branch in `PutAsync` has the `TryGetArray` check inverted:

```csharp
if (MemoryMarshal.TryGetArray((ReadOnlyMemory<byte>)memory, out var segment) == false)
{
    // segment is default here so segment.Array is null
    read = await hashedStream.ReadAsync(segment.Array!, segment.Offset, segment.Count, cancellationToken);
}
else
{
    // taken when memory IS array backed, which is always
    var bytes = ArrayPool<byte>.Shared.Rent(memory.Length);
    ...
}
```

`NatsMemoryOwner<byte>.Memory` is always array backed so `TryGetArray` always succeeds. That means netstandard2.0 rents an extra chunk sized buffer and copies the whole chunk on every iteration. The other branch is unreachable and would NRE if it ever ran. Came in with #513.

This just flips the condition so the array backed case reads straight into the chunk buffer and keeps the pooled buffer as the fallback.

No test added since the branch only compiles under `#if NETSTANDARD2_0` and the object store tests target net8.0Builds clean on all four TFMs.

### DIFF
--- a/.github/workflows/test_windows.yml
+++ b/.github/workflows/test_windows.yml
@@ -92,7 +92,8 @@ jobs:
           cd tests/NATS.Client.ObjectStore.Encoder.Tests
           dotnet test -c Release --no-build -p:TestTfmsInParallel=false
 
-      # net481 exercises the netstandard2.0 build of the client (e.g. the PutAsync read loop)
+      # net481 resolves the netstandard2.0 build of the client, the only TFM that compiles the
+      # PutAsync read loop; NetStandardPutPathTest is what actually asserts against it
       - name: Test Object Store
         run: |
           tasklist | grep -i nats-server && taskkill -F -IM nats-server.exe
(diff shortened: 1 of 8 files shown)


mtmk:
thanks @fresh55 I just added object store tests to run on windows net481 so we have a little better verification.

mtmk (approved):
LGTM

mtmk:
hey @fresh55 could you add your ssh key as signing in your github settings https://github.com/settings/keys? I am not able to merge otherwise with error: 'Merging is blocked: Commits must have verified signatures'. (it's our repo setting requiring signed commits because of CNCF rules) thanks!

fresh55:
@mtmk added it

PR #1240, opened 2026-08-15
Fix KV reads on buckets with AllowDirect disabled

### LINKED ISSUE #1238: KV reads are broken on buckets with AllowDirect = false
### Observed behavior

reading a KV bucket whose stream has AllowDirect = false gives wrong results.

GetEntryAsync(key, revision) always throws NatsKVException "Unexpected subject". the check at line 389 returns the error when the subject matches, line 316 does the same check the other way round.

deleted keys come back as live entries. that path never looks at the headers so KV-Operation is ignored, and you get Operation = Put with a null value instead of NatsKVKeyDeletedException. plain GetEntryAsync(key) is affected too, no revision needed.

after a delete, CreateAsync on that key never succeeds again. the marker looks like a live entry to TryCreateAsync.

different from nats-io/nats.net#1060, that one was about creating buckets with the flag off. this is about reading ones that already have it.

### Expected behavior

same as a bucket with direct get on. or a proper error if these buckets aren't supported.

### Steps to reproduce

made the stream by hand with AllowDirect = false and subjects $KV.oldbucket.>, then GetStoreAsync("oldbucket"). same client and server, next to a normal bucket:

```
                                normal      AllowDirect false
GetEntryAsync("k")              ok v1       ok v1
GetEntryAsync("k", revision 1)  ok v1       Unexpected subject
GetEntryAsync after delete      deleted     entry op=Put value=null
CreateAsync on deleted key      ok          Can't create entry
```

fresh55:
On buckets whose stream has AllowDirect disabled, the non-direct read path in TryGetEntryAsync had its subject check inverted, so GetEntryAsync with a revision always threw "Unexpected subject". It also never decoded the response headers, so KV-Operation DEL and PURGE were ignored and deleted keys came back as live entries with Operation = Put, which affects plain GetEntryAsync too and left CreateAsync failing permanently on a deleted key.

Operation parsing now lives in one helper shared by both paths, matching nats.go which parses the message once for both. 

Fixes #1238


mtmk:
merged main with fixed tests

mtmk (approved):
LGTM thanks @fresh55

PR #1241, opened 2026-08-16
Add auto select in GetAsync/GetDirectAsync

### LINKED ISSUE #1059: Proposal: JetStream GetAutoAsync() and NatsStreamMsg<T>
### Proposed change

It looks like we are not very helpful when it comes to retrieving messages from streams. A new type to encapsulate a message and metadata on top of two different ways of getting messages from a stream should help applications to access data more naturally.

On to of that we don't have a mechanism to get the message conveniently trying direct get first if allowed then falling back to stream get. This could be implemented as GetAutoAsync() on top of GetAsync() and GetDirectAsync().

### Use case

Application needs access to stream messages without creating a consumer and wants to have a way to access message data in a strongly typed fashion.

### Contribution

Suggestions welcome

Ivandemidov00:
Resolves #1059


mtmk:
@Ivandemidov00 thanks for the PR. I'm just thinking actually this isn't a universal function among all clients and perhaps it belongs in [orbit .net ](https://github.com/synadia-io/orbit.net)rather than here.

Ivandemidov00:
I create PR https://github.com/synadia-io/orbit.net/pull/76

PR #1242, opened 2026-08-19
Update OpenTelemetry documentation

mtmk:
Corrects a statement about `Activity.Current` that does not match the code, and documents the receive span behaviour added in #1236.


scottf (approved):
LGTM

PR #1243, opened 2026-08-21
Fix stall when TCP splits the trailing CRLF of an empty payload message

### LINKED ISSUE #1237: Empty payload message is not delivered if TCP splits the trailing CRLF
### Observed behavior

if TCP splits an empty payload message right between the \r and the \n at the end of the frame, the client never delivers it. the read loop waits for a byte that is not coming, and nothing else on that connection gets delivered either.

it unblocks when the server sends anything else. on an idle connection that is the PONG for our own keepalive, and PingInterval defaults to 2 minutes, so a request can time out even though the reply arrived.

NatsReadProtocolProcessor line 216 calls ReadAtLeastAsync(2), but ReadAtLeastAsync only counts bytes read during that call, so a \r already in the buffer does not count. lines 168 and 235 both subtract what is already buffered. line 216 should be ReadAtLeastAsync(2 - (int)payloadSlice.Length).

### Expected behavior

message delivered as soon as its last byte arrives.

### Steps to reproduce

server sends INFO, answers PING with PONG, then writes:

    write 1:  MSG foo 2 0\r\n\r
    write 2:  \n

subscribe to foo. the message arrives 3 seconds later, and only because the server sends an unrelated PING then. with the line 216 change it arrives in 6 ms.

mtmk:
An empty payload MSG split between the trailing CR and LF was not delivered until other traffic arrived on the connection.

Fixes #1237

### DIFF
--- a/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
+++ b/src/NATS.Client.Core/Internal/NatsReadProtocolProcessor.cs
@@ -213,7 +213,7 @@ internal sealed class NatsReadProtocolProcessor : IAsyncDisposable
                             if (payloadSlice.Length < 2)
                             {
                                 _socketReader.AdvanceTo(payloadBegin);
-                                buffer = await _socketReader.ReadAtLeastAsync(2).ConfigureAwait(false); // \r\n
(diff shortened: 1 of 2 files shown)


AdamPayzant (approved):
LGTM

PR #1245, opened 2026-08-25
Release 3.1.1

mtmk:
Stable 3.1.1 on the 3.1 patch line, promoting 3.1.1-preview.1 and the fixes that landed after it.

- Fix receive activity leak when tracing is enabled (#1229)
- Fix inverted array check in object store put path (#1235)
- Fix OpenTelemetry receive span lifecycle, parenting and tags (#1236)
- docs: correct Enrich ambient context and cover endpoint spans (#1242)
- Fix stall when TCP splits the trailing CRLF of an empty payload message (#1243)


scottf (approved):
LGTM

PR #1246, opened 2026-08-25
Fix TlsPreferTest hang that aborts the Core2 test run

mtmk:
The fake server closed the socket as soon as the handshake completed. When that beat ConnectAsync's return the client was mid-reconnect, and ConnectAsync waited on a connection the single-accept listener could never give it, hanging the assembly until the 600s session timeout. Cause of the intermittent Core2 timeouts since June.

Second commit reports long-running tests, since an aborted assembly names nothing.


MauriceVanVeen (approved):
LGTM

caspervonb (approved):

PR #1248, opened 2026-08-28
Add offline asset reporting to stream and consumer lists

mtmk:
Server 2.14 reports assets that need a higher API level than the server supports in an offline map on the stream and consumer list responses, keyed by name with the reason as the value, and repeats the name in missing. Both were dropped, and the consumer list response had no missing either.


scottf (approved):
LGTM

PR #1249, opened 2026-08-28
Add sourcing to consumer config

mtmk:
Server 2.14 sets sourcing on the consumers it creates for interest and work queue stream sourcing, and relaxes work queue ack policy validation for them. Read-only exposure in practice, kept settable for symmetry with the existing direct flag.


scottf (approved):
LGTM

PR #1252, opened 2026-09-01
Add option to retain history of recently deleted keys in PurgeDeletesAsync

### LINKED ISSUE #1247: Allow keeping history of recently deleted keys in PurgeDeletesAsync
### Proposed change

Currently it's a hard-coded behaviour in `NatsKVStore.PurgeDeletesAsync` that [if a key was deleted recently, it will be pruned to just its tombstone](https://github.com/nats-io/nats.net/blob/49f28003ca5d6b8a6d648c86e121fdfd1d5b35f5/src/NATS.Client.KeyValueStore/NatsKVStore.cs#L609). I propose we add a boolean options field to this operation changing it so that instead (opt-in), `PurgeDeletesAsync` purges every delete sufficiently old, but doesn't touch keys deleted within the tolerance.

### Use case

It is useful in KV stores where history is valuable, eg. for observability. Currently the main way to GC a KV store is this method but sometimes you only want to GC old revisions, whether current or not. Without this it's a finicky reimplementation of most of the internals of the method to get the desired behaviour.

### Contribution

I can contribute if the proposal is accepted

adamjones2:
fixes #1247

### DIFF
--- a/src/NATS.Client.KeyValueStore/NatsKVOpts.cs
+++ b/src/NATS.Client.KeyValueStore/NatsKVOpts.cs
@@ -87,7 +87,18 @@ public record NatsKVDeleteOpts
 
 public record NatsKVPurgeOpts
 {
-    public static readonly NatsKVPurgeOpts Default = new() { DeleteMarkersThreshold = TimeSpan.FromMinutes(30) };
+    public static readonly NatsKVPurgeOpts Default = new()
+    {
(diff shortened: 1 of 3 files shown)


mtmk:
Looks good. I'd be nice to have a test for it.

mtmk (approved):
LGTM thanks @adamjones2

PR #1253, opened 2026-09-01
Add option to pass own JsonWriterOptions to NatsJsonSerializer

### LINKED ISSUE #1219: Add JsonWriterOptions to NatsJsonSerializer<T> constructor
### Proposed change

Could we allow providing our own JsonWriterOptions? Would be nice to enable using our own `JavaScriptEncoder` while reusing the logic of the serialiser class. Eg. the unicode escaping in messages at the moment is quite tough on the eyes when inspecting the raw values of the keys via NATS CLI etc.

### Use case

Described above

### Contribution

I can contribute if the idea is approved

adamjones2:
fixes #1219

### DIFF
--- a/src/NATS.Client.Serializers.Json/NatsJsonSerializer.cs
+++ b/src/NATS.Client.Serializers.Json/NatsJsonSerializer.cs
@@ -1,6 +1,5 @@
 using System.Buffers;
 using System.Text.Json;
-using System.Text.Json.Serialization;
 using NATS.Client.Core;
 
 namespace NATS.Client.Serializers.Json;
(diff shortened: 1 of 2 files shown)


src/NATS.Client.Serializers.Json/NatsJsonSerializer.cs
```diff
@@ -55,9 +67,13 @@ public NatsJsonSerializer()
     public void Serialize(IBufferWriter<byte> bufferWriter, T? value)
     {
         Utf8JsonWriter writer;
-        if (_jsonWriter == null)
+        if (_writerOpts != null)
+        {
+            writer = new Utf8JsonWriter(bufferWriter, _writerOpts.Value);
```
mtmk: allocated every call we should mention in docs. we should dispose as well.

src/NATS.Client.Serializers.Json/NatsJsonSerializer.cs
```diff
@@ -43,6 +44,17 @@ public NatsJsonSerializer()
     /// <param name="opts">Serialization options</param>
     public NatsJsonSerializer(JsonSerializerOptions opts) => _opts = opts;
```
mtmk: should we also forward the `Encoder` option? i.e.:
```csharp
public NatsJsonSerializer(JsonSerializerOptions opts)
    : this(opts, new JsonWriterOptions { Encoder = opts.Encoder, Indented = false, SkipValidation = true })
```

mtmk (commented):
Looks good. couple of comments.

src/NATS.Client.Serializers.Json/NatsJsonSerializer.cs
adamjones2: can do! it's a change in behaviour for existing clients though so thought maybe we'd want to keep as is. Should be safe though.

adamjones2: I updated it to do what you suggested in the other thread and removed the path where the writer's reused. It did simplify it down a fair bit.

mtmk:
hey @adamjones2 I wanted to make sure we didn't change the GC profile for the existing applications. wdyt about this approach? #1259

adamjones2:
> hey @adamjones2 I wanted to make sure we didn't change the GC profile for the existing applications. wdyt about this approach? #1259

@mtmk looks good to me!

mtmk (approved):
LGTM thanks @adamjones2

PR #1255, opened 2026-09-02
Enable stopping of service endpoints independently

### LINKED ISSUE #1254: Allow removing endpoints from Services
### Proposed change

- Make `INatsSvcServer.AddEndpointAsync` and `Group.AddEndpointAsync` return `ValueTask<INatsSvcEndpoint>` - returning the endpoint they added
- Add `ValueTask StopAsync(CancellationToken cancellationToken = default)` to `INatsSvcEndpoint`, which drains and disposes the endpoint (analogous to `INatsSvcServer.StopAsync`)
- Make the implementation of the endpoint returned from the `AddEndpointAsync` methods also deregister themselves from `_endpoints` (using a decorator so `NatsSvcEndpoint<>` stays as is)

Clients can then stop endpoints added with `AddEndpointAsync` without needing to stop the whole service.

### Use case

A quite useful feature of Nats request-reply is that servers can dynamically opt in and out of responding to requests. Sometimes this should be done on a more granular basis (as a pod's responsibilities shift over its lifetime or as different in-process services become healthy or unhealthy).

Currently we can only safely stop the entire service rather than individual endpoints on it. It's also possible to do this by constructing the `NatsSvcEndpoint<>` ourselves and not using `INatsSvcContext` but this opts out of INFO/STATS etc on the endpoints.

### Contribution

I can contribute

adamjones2:
fixes #1254

### DIFF
--- a/src/NATS.Client.Services/NatsSvcServer.cs
+++ b/src/NATS.Client.Services/NatsSvcServer.cs
@@ -102,11 +102,15 @@ public class NatsSvcServer : INatsSvcServer
     /// <param name="serializer">Serializer to use when deserializing incoming messages (defaults to connection's serializer).</param>
     /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to stop the endpoint.</param>
     /// <typeparam name="T">Serialization type for messages received.</typeparam>
-    /// <returns>A <seealso cref="ValueTask"/> representing the asynchronous operation.</returns>
+    /// <returns>The endpoint that was added.</returns>
     /// <remarks>
     /// One of name or subject must be specified.
+    /// <para>
+    /// Use <see cref="INatsSvcEndpoint.StopAsync(CancellationToken)"/> on the returned endpoint to retire
+    /// it gracefully without affecting the other endpoints on this service
+    /// </para>
     /// </remarks>
-    public ValueTask AddEndpointAsync<T>(Func<NatsSvcMsg<T>, ValueTask> handler, string? name = default, string? subject = default, string? queueGroup = default, IDictionary<string, string>? metadata = default, INatsDeserialize<T>? serializer = default, CancellationToken cancellationToken = default)
+    public ValueTask<INatsSvcEndpoint> AddEndpointAsync<T>(Func<NatsSvcMsg<T>, ValueTask> handler, string? name = default, string? subject = default, string? queueGroup = default, IDictionary<string, string>? metadata = default, INatsDeserialize<T>? serializer = default, CancellationToken cancellationToken = default)
     {
         queueGroup ??= _config.UseQueueGroup ? _config.QueueGroup : null;
         return AddEndpointInternalAsync<T>(handler, name, subject, queueGroup, metadata, serializer, cancellationToken);
@@ -216,7 +220,7 @@ public class NatsSvcServer : INatsSvcServer
         }
     }
 
-    private async ValueTask AddEndpointInternalAsync<T>(Func<NatsSvcMsg<T>, ValueTask> handler, string? name, string? subject, string? queueGroup, IDictionary<string, string>? metadata, INatsDeserialize<T>? serializer, CancellationToken cancellationToken)
+    private async ValueTask<INatsSvcEndpoint> AddEndpointInternalAsync<T>(Func<NatsSvcMsg<T>, ValueTask> handler, string? name, string? subject, string? queueGroup, IDictionary<string, string>? metadata, INatsDeserialize<T>? serializer, CancellationToken cancellationToken)
     {
         serializer ??= _nats.Opts.SerializerRegistry.GetDeserializer<T>();
 
@@ -234,6 +238,8 @@ public class NatsSvcServer : INatsSvcServer
         }
 
         await ep.StartAsync(cancellationToken).ConfigureAwait(false);
+
+        return new RegisteredNatsSvcEndpoint(ep, () => ((ICollection<KeyValuePair<string, INatsSvcEndpoint>>)_endPoints).Remove(new(ep.Name, ep)));
     }
 
     private async Task MsgLoop()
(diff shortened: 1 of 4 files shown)


mtmk:
hi @adamjones2 thanks for the PR. Could you sign your commits please? thanks.

adamjones2:
> hi @adamjones2 thanks for the PR. Could you sign your commits please? thanks.

sure, done!

mtmk:
need to think about this one since it's a breaking change.

mtmk:
thinking about this a bit more, the thing I want to avoid is changing existing `AddEndpointAsync` signature. it would be mostly source compatible but it is a binary break and it not only breaks anyone implementing `INatsSvcServer` (obvious) but also users of the code (not recompiled) would see a `MissingMethodException`.

alternative is to leave `AddEndpointAsync` as it is and add a `RemoveEndpointAsync` (by name) to the service. Endpoints are already tracked by name, so it can look one up, drain it, dispose it and drop it from the map. That also drops the need for the `RegisteredNatsSvcEndpoint` wrapper and for `StopAsync` on `INatsSvcEndpoint`.

that still adds a member to `INatsSvcServer`, which is a compile break for anyone implementing the interface but I'd say it's relatively lot lower usage chance for transitive dependencies, we probably break hand written test mocks, also a default interface member is not an option since we target netstandard2.0.

`DrainAsync` override that awaits the handler task looks good to me independently of which option we go with, be good to keep that.

would you be up for reworking it?

adamjones2:
> thinking about this a bit more, the thing I want to avoid is changing existing `AddEndpointAsync` signature. it would be mostly source compatible but it is a binary break and it not only breaks anyone implementing `INatsSvcServer` (obvious) but also users of the code (not recompiled) would see a `MissingMethodException`.
> 
> alternative is to leave `AddEndpointAsync` as it is and add a `RemoveEndpointAsync` (by name) to the service. Endpoints are already tracked by name, so it can look one up, drain it, dispose it and drop it from the map. That also drops the need for the `RegisteredNatsSvcEndpoint` wrapper and for `StopAsync` on `INatsSvcEndpoint`.
> 
> that still adds a member to `INatsSvcServer`, which is a compile break for anyone implementing the interface but I'd say it's relatively lot lower usage chance for transitive dependencies, we probably break hand written test mocks, also a default interface member is not an option since we target netstandard2.0.
> 
> `DrainAsync` override that awaits the handler task looks good to me independently of which option we go with, be good to keep that.
> 
> would you be up for reworking it?

sure no problem. I thought this might come up hah, considered the `RemoveEndpoint` route but wasn't sure how hard we wanted to preserve back compat here and erred to async disposable as I thought the returned endpoint could be handy for other things. But the other route works too, have switched to that now.

mtmk:
thanks @adamjones2 will check soon. btw dropped a comment about BC in https://slack.nats.io #dotnet channel https://natsio.slack.com/archives/C1V81FKU6/p1788945212253759

src/NATS.Client.Services/NatsSvcServer.cs
```diff
@@ -112,6 +112,32 @@ public ValueTask AddEndpointAsync<T>(Func<NatsSvcMsg<T>, ValueTask> handler, str
...
+    /// Endpoints added through a <see cref="Group"/> are removed by their endpoint name, without
+    /// the group prefix.
+    /// </para>
+    /// </remarks>
+    /// <exception cref="NatsSvcException">There is no endpoint with that name.</exception>
+    public async ValueTask RemoveEndpointAsync(string name, CancellationToken cancellationToken = default)
+    {
+        if (!_endPoints.TryRemove(name, out var ep))
+        {
+            throw new NatsSvcException($"Endpoint '{name}' does not exist");
```
mtmk, line 134: should we add a `TryRemove` method to avoid exception handling?

mtmk:
thanks @adamjones2 looks good. Added a minor comment above as well. also it looks like we should drain properly in `StopAsync` too (your drain implementation made me think), possibly another PR. opened an issue for it #1260.

a few comments from an AI review:

> **Endpoint names derived from subjects**: The remarks say group endpoints are removed by name without the group prefix, which only holds when name was passed. With just a subject, the key is the full subject with dots replaced by dashes, group prefix included - so an endpoint added as subject: "s1" on group g1 is removed as g1-s1. That rule was an internal detail before and is now part of the public API, so it needs documenting on RemoveEndpointAsync, and a test for removing a group endpoint both ways.
> 
> **The cancellation token**: The doc says it bounds the drain, but NatsSubBase.DrainAsync only uses it to shorten the PING fence and never throws on cancel, and the await _handlerTask after it ignores the token entirely. A handler that blocks forever hangs RemoveEndpointAsync regardless of the token. Either bound that wait or reword to match the base class.
> 
> **The endpoint is removed from the dictionary before the drain**: so if anything throws in between it is neither listed nor disposed - both paths swallow today, but a finally around the dispose is cheap insurance. And asserting the exact exception message in the test couples it to the wording.

src/NATS.Client.Services/NatsSvcServer.cs
adamjones2: usually I would prefer this but here it would have an asymmetry with `AddEndpointAsync` which has this same "throw if already there" behaviour. My inclination would be to have the Remove API mirror the Add API but I can change it if you prefer. If we did change it I guess we'd probably want to add just a `TryRemove` to the interface and a `Remove` as an extension on it.

mtmk: good point. created an issue for it #1261

adamjones2:
> thanks @adamjones2 looks good. Added a minor comment above as well. also it looks like we should drain properly in `StopAsync` too (your drain implementation made me think), possibly another PR. opened an issue for it #1260.
> 
> a few comments from an AI review:
> 
> > **Endpoint names derived from subjects**: The remarks say group endpoints are removed by name without the group prefix, which only holds when name was passed. With just a subject, the key is the full subject with dots replaced by dashes, group prefix included - so an endpoint added as subject: "s1" on group g1 is removed as g1-s1. That rule was an internal detail before and is now part of the public API, so it needs documenting on RemoveEndpointAsync, and a test for removing a group endpoint both ways.
> > **The cancellation token**: The doc says it bounds the drain, but NatsSubBase.DrainAsync only uses it to shorten the PING fence and never throws on cancel, and the await _handlerTask after it ignores the token entirely. A handler that blocks forever hangs RemoveEndpointAsync regardless of the token. Either bound that wait or reword to match the base class.
> > **The endpoint is removed from the dictionary before the drain**: so if anything throws in between it is neither listed nor disposed - both paths swallow today, but a finally around the dispose is cheap insurance. And asserting the exact exception message in the test couples it to the wording.

@mtmk on the AI comments:

- I added a doc and test for the endpoint names derived from subjects
- I don't think there's much we can do about the hanging because `DisposeAsync` also awaits the task and doesn't take a cancellation token, but I've clarified the doc
- I added a try-finally for throws between the dictionary removal and drain

mtmk (approved):
LGTM thanks @adamjones2 

note, we're leaving unrelated failed tests which are already fixed in main.

PR #1256, opened 2026-09-03
Fix consumer list test against nats-server main

mtmk:
nats-server 2.15 gives every stream a default limit of 1000 consumers unless the stream sets one itself, so the consumer list test fails against the server main branch while creating the 1200 consumers it needs to page past the names API limit. The stream now carries an explicit limit. Note that -1 does not work: the server substitutes its default for any value not greater than zero, so unlimited is no longer expressible per stream.

Currently red on every open PR.


Jarema (approved):
LGTM

PR #1258, opened 2026-09-08
feat: add opt-in case-sensitive header handling

### LINKED ISSUE #1257: NatsHeaders is case-insensitive, but ADR-4 says client header operations should be case-sensitive
### Observed behavior

`NatsHeaders` builds its backing store with `StringComparer.OrdinalIgnoreCase`, so
`X-Trace` and `x-trace` are treated as the same header. ADR-4 says the opposite:

ref: https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-4.md

> All operations are _case-sensitive_. Application implementations should provide an
> option(s) to enable clients to work in a case-insensitive [manner]

and separately that NATS headers are "case preserving — the publisher's case will be
preserved". So case-insensitive is meant to be an opt-in option, and here it's the only
behaviour available.

**Repro** (no server needed):

```csharp
var headers = new NatsHeaders();
headers["X-Trace"] = "alpha";
headers["x-trace"] = "bravo";

Console.WriteLine(headers.Count);          // 1   — expected 2
Console.WriteLine(headers["X-Trace"]);     // bravo — alpha is gone
```

Two symptoms fall out of this:

- **Publishing loses data.** The indexer overwrites, so the wire gets a single
  `X-Trace: bravo` and `alpha` is dropped silently.
- **Receiving loses the publisher's casing.** A message published from e.g. the Go client
  with both `X-Trace: alpha` and `x-trace: bravo` arrives as one slot keyed `X-Trace` with
  values `[alpha, bravo]` — indistinguishable from a message that sent `X-Trace` twice.

### Expected behavior

The headers are by default handled as case sensitive. It's currently not possible at all afaict!

The receive side is the one that can't be worked around: `NatsHeaders` doesn't retain the
raw header block, and there's no way to supply a parser or a comparer through `NatsOpts`.

On the publish side there is a workaround, which also suggests the fix is small — the
public constructor accepts a custom store as an arg:

```csharp
var headers = new NatsHeaders(new Dictionary<string, StringValues>(StringComparer.Ordinal));
```

With that, both keys go out on the wire with casing intact, so it's really just the default comparer plus the parser's `EnsureStore` that cause the behaviour.

Would you take a PR that makes the store ordinal by default, or adds a `NatsOpts` switch
if changing the default is too breaking?

### Server and client version

NATS.Client.Core 2.8.1
nats-server 2.14.3

### Host environment

runtime: .NET 10.0.9
host: Ubuntu 24.0.4 LTS
via the  dotnet sdk:10.0, SDK 10.0.301 oci container.

### Steps to reproduce

**Repro** (no server needed):

```csharp
var headers = new NatsHeaders();
headers["X-Trace"] = "alpha";
headers["x-trace"] = "bravo";

Console.WriteLine(headers.Count);          // 1   — expected 2
Console.WriteLine(headers["X-Trace"]);     // bravo — alpha is gone
```

jeacott1:
Fixes #1257.

NatsHeaders hard-codes StringComparer.OrdinalIgnoreCase, which is the opposite of what ADR-4 asks for.
Case-sensitive should  be the default and case-insensitive the option.
On the publish side it's worse than just matching - two keys that differ only by case collide in the indexer and one value is silently dropped before it ever hits the wire.

As suggested in the issue, I've left the default alone and added the opt-in instead:

- `new NatsHeaders(caseSensitive: true)`  - per instance, works for both publishing and reading headers.
- `NatsOpts.CaseSensitiveHeaders` - drives the headers the client builds for received messages.

`NatsHeaders(Dictionary<string, StringValues>)` is untouched, so anyone already passing their own store keeps their own comparer. There's a TODO on the comparer flagging it to reverse at the next major release.

There could be something like `connection.CreateHeaders()` so `CaseSensitiveHeaders` reaches the publish side too, but I haven't added it here. It would need to be on `INatsConnection` to be useful, which breaks implementors, and a default interface member won't work on netstandard2.0.

happy to add it if you prefer though.
Cheers

### DIFF
--- a/src/NATS.Client.Core/NatsMsg.cs
+++ b/src/NATS.Client.Core/NatsMsg.cs
@@ -389,7 +389,7 @@ public readonly record struct NatsMsg<T> : INatsMsg<T>
 
         if (headersBuffer != null)
         {
-            headers = new NatsHeaders();
+            headers = new NatsHeaders(headerParser.CaseSensitiveHeaders);
 
             try
             {
--- a/tests/NATS.Client.CoreUnit.Tests/NatsHeaderTest.cs
+++ b/tests/NATS.Client.CoreUnit.Tests/NatsHeaderTest.cs
@@ -228,4 +228,54 @@ public class NatsHeaderTest
 
         Assert.Equal(expected.Length, bytesLength);
     }
+
+    [Fact]
+    public void ParserCaseInsensitiveByDefaultTests()
+    {
+        var parser = new NatsHeaderParser(Encoding.UTF8);
+        var text = "NATS/1.0\r\nX-Trace: alpha\r\nx-trace: bravo\r\n\r\n";
+        var input = new SequenceReader<byte>(new ReadOnlySequence<byte>(Encoding.UTF8.GetBytes(text)));
+        var headers = new NatsHeaders();
+        parser.ParseHeaders(input, headers);
+
+        Assert.False(parser.CaseSensitiveHeaders);
+        Assert.Single(headers);
+        Assert.Equal(new[] { "alpha", "bravo" }, headers["X-Trace"].ToArray());
+    }
+
+    [Fact]
+    public void ParserCaseSensitiveTests()
+    {
+        var parser = new NatsHeaderParser(Encoding.UTF8, caseSensitiveHeaders: true);
+        var text = "NATS/1.0\r\nX-Trace: alpha\r\nx-trace: bravo\r\n\r\n";
+        var input = new SequenceReader<byte>(new ReadOnlySequence<byte>(Encoding.UTF8.GetBytes(text)));
+        var headers = new NatsHeaders(caseSensitive: true);
+        parser.ParseHeaders(input, headers);
+
+        Assert.True(parser.CaseSensitiveHeaders);
+        Assert.Equal(2, headers.Count);
+        Assert.Equal("alpha", headers["X-Trace"]);
+        Assert.Equal("bravo", headers["x-trace"]);
+    }
+
+    [Fact]
+    public async Task WriterCaseSensitiveTests()
+    {
+        var headers = new NatsHeaders(caseSensitive: true)
+        {
+            ["X-Trace"] = "alpha",
+            ["x-trace"] = "bravo",
+        };
+        var pipe = new Pipe(new PipeOptions(pauseWriterThreshold: 0));
+        var writer = new HeaderWriter(Encoding.UTF8);
+        var written = writer.Write(pipe.Writer, headers);
+
+        var text = "NATS/1.0\r\nX-Trace: alpha\r\nx-trace: bravo\r\n\r\n";
+        var expected = new ReadOnlySequence<byte>(Encoding.UTF8.GetBytes(text));
+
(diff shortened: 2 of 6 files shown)


jeacott1:
@mtmk my first pass forgot the signed commit. thats been fixed now.

src/NATS.Client.Core/NatsMsg.cs
```diff
@@ -389,7 +389,7 @@ internal static NatsMsg<T> BuildInternal(
 
         if (headersBuffer != null)
         {
-            headers = new NatsHeaders();
```
mtmk, line 392: there are couple more places we new up headers: telemetry paths and deserialize error path.

tests/NATS.Client.CoreUnit.Tests/NatsHeaderTest.cs
```diff
@@ -228,4 +228,54 @@ public void GetBytesLengthTest()
...
+        var written = writer.Write(pipe.Writer, headers);
+
+        var text = "NATS/1.0\r\nX-Trace: alpha\r\nx-trace: bravo\r\n\r\n";
+        var expected = new ReadOnlySequence<byte>(Encoding.UTF8.GetBytes(text));
+
+        Assert.Equal(expected.Length, written);
+        await pipe.Writer.FlushAsync();
+        var result = await pipe.Reader.ReadAtLeastAsync((int)written);
+        Assert.True(expected.ToSpan().SequenceEqual(result.Buffer.ToSpan()));
+    }
```
mtmk, line 281: it'd be nice to have a couple of tests using `NatsOpts { CaseSensitiveHeaders = true }`

mtmk (commented):
thanks @jeacott1 looks good in general to me. added couple of comments.

jeacott1:
@mtmk right, fixed on both counts. 

note - I've fixed the telemetry path but that means the connection option now
does affect publishing, in the one case where the connection builds the
headers itself (tracing on, caller passed none). So the "this is
receive-side only" isn't quite right any more — I've reworded the NatsOpts
doc to say it covers headers the client creates on your behalf.

Left alone: the ??= new NatsHeaders() sites in NatsJSContext.TryPublishAsync,
NatsSvcMsg.ReplyErrorAsync and NatsKVStore. Those stamp fixed-case
well-known keys onto a caller-owned headers object and the Core option
isn't cleanly reachable from JetStream/Services — happy to extend if you
want them included.

mtmk:
thanks @jeacott1 all look good.

tests are failing but i think it's a simple compile error.

Yes, agreed. we can leave those alone since they're only used for internal well known headers, not looked up only serialized, case difference would not affect the outcome at all.

jeacott1:
@mtmk sorry, thats on me, I didnt build and run the full suite. 
fixed.

mtmk:
> @mtmk sorry, thats on me, I didnt build and run the full suite. fixed.

@jeacott1 no worries, thanks for fixing them. ci has a test failure that is fixed in main as well, so i'm going to ignore that in this PR. we could merge main, but it would be wasted ci cycles.

mtmk (approved):
LGTM thanks @jeacott1

PR #1259, opened 2026-09-15
Keep the JSON writer cached per instance

mtmk:
Stacked on #1253. Serialization is on the publish hot path, so existing applications must not see a changed GC profile. Gating writer reuse on the owning instance keeps the encoder fix and the allocation profile.

Allocated per serialize, short run, net8.0:

| | base | #1253 | this |
| --- | ---: | ---: | ---: |
| reflection JSON | 0 B | 120 B | 0 B |
| client default chain | 0 B | 120 B | 0 B |
| other serializers | 0 B | 0 B | 0 B |


mtmk:
Both commits are now on #1253, so this is redundant.

PR #1262, opened 2026-09-17
Fix snapshot test chunk collection

mtmk:
The snapshot backup test treated a message with no reply subject as the end of the snapshot, but the server sends the final partial chunk without one and only the end marker has an empty payload. Any snapshot whose size was not a chunk multiple lost its tail, and the remaining assertions could not see it: chunk count, byte count and the S2 magic all hold for any non-empty prefix. Server 2.15 now compresses a small snapshot below the 1KiB chunk size the test asks for, which made the discarded chunk the only one and turned the silent truncation into a failure on the nats-server main CI leg. The test now collects until the payload is empty and restores the snapshot back over the deleted stream, comparing message count and contents, so a dropped chunk fails.


scottf (approved):
LGTM

